PR #31, opened 2026-07-01
ci: add native-Windows test lane + POSIX/Win32 test gating

dracic:
## What

Add a native **`windows-latest` CI job** and the **POSIX/Win32 test-gating** that makes the suite green on it. Source is already POSIX-decoupled; this is the test-layer + CI half (the follow-up flagged in the win32-process-liveness PR).

## Why

The suite carried POSIX-only test constructs that failed — or *killed the pytest runner mid-suite* — on native Windows, and there was no Windows CI lane to catch regressions. A verified native-Windows baseline was **24 failures** (WSL, being Linux, had masked the Windows-only ones). One real product bug hid among them: worktree git-exclude patterns were written with `\` separators, which git never anchors.

## How

- **CI:** new `test-windows` (windows-latest, py3.13) job running the deterministic suite; `persist-credentials: false` on checkouts.
- **One src fix:** `_worktree_local_exclude` normalizes `\`→`/` before writing `.git/info/exclude` (git patterns are POSIX-slash on every platform; a no-op on POSIX).
- **Test gating (POSIX skipped on win32, never on Linux):** cross-platform verify commands (`test -f` / `if exist`) and dead-PID spawns (`[sys.executable, "-c", ""]` vs `["true"]`); `real_signal` uses `signal.raise_signal` (catchable in-process on both OSes — keeps Windows coverage instead of `TerminateProcess`-killing the runner); `HAVE_TMUX` excludes win32; `USERPROFILE` for expanduser; `echo hi& exit 7` for cmd; `encoding="utf-8"` reads; a `.cmd`-shim launcher for fake CLIs; `skipif(win32)` on the 3 POSIX-only Unity symlink-priming tests.

## Testing

`test_portability_guard.py` (the existing regression guard) stays green. Full suite verified on **native Windows** (24 baseline failures → 0 in-scope) and **WSL/Linux** (1236 passed — no regression; ports are POSIX no-ops). Ruff clean. Remaining local-only failures (`test_module_skills_sync`) are workspace `.claude`/`.agents` drift that **skips in CI** and is out of scope.

## Follow-up

Native-Windows Unity Library symlink-priming (the 3 skipped tests) — tracked separately.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated Windows CI job running the deterministic test suite on Python 3.11–3.14.
* **Bug Fixes**
  * Improved Windows worktree local-exclude handling for more reliable path matching.
  * Strengthened UTF-8 behavior in Windows test execution to reduce encoding-related failures.
* **Tests**
  * Updated tests for cross-platform command generation, fake script launchers, marker-file checks, and safer subprocess signaling.
  * Reduced CI flakiness by replacing fixed delays with condition-based waits.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,11 @@ jobs:
     # Real-Windows runner for the deterministic suite (live psmux e2e stays a manual gate).
     name: test (windows, py${{ matrix.python-version }})
     runs-on: windows-latest
+    # The suite reads/writes UTF-8 files; Windows' cp1252 default would break plain
+    # text I/O. UTF-8 mode (PEP 686's Python 3.15 default) makes the runner match the
+    # files under test; the conftest guard enforces the same for local win32 runs.
+    env:
+      PYTHONUTF8: '1'
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,8 @@ jobs:
         python-version: ['3.11', '3.12', '3.13', '3.14']
     steps:
       - uses: actions/checkout@v7
+        with:
+          persist-credentials: false
 
       - name: Install uv
         uses: astral-sh/setup-uv@v8.2.0
@@ -41,6 +43,31 @@ jobs:
           sudo apt-get update -o Acquire::Retries=3 || true
           sudo apt-get install -y tmux
           tmux -V
+
+      - name: Install the project
+        run: uv sync --locked --all-extras
+
+      - name: Run tests
+        run: uv run pytest -q
+
+  test-windows:
+    # Real-Windows runner for the deterministic suite (live psmux e2e stays a manual gate).
+    name: test (windows, py${{ matrix.python-version }})
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.11', '3.12', '3.13', '3.14']
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          persist-credentials: false
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v8.2.0
+        with:
+          enable-cache: true
+          python-version: ${{ matrix.python-version }}
 
       - name: Install the project
         run: uv sync --locked --all-extras
--- a/src/automator/install.py
+++ b/src/automator/install.py
@@ -180,9 +180,8 @@ def _worktree_local_exclude(worktree: Path, patterns: Sequence[str]) -> None:
     git's standard local-only exclude (never committed or pushed); it does not
     affect already-tracked files. Best-effort — skipped if git can't be queried.
     """
-    # git's exclude patterns are POSIX-slash on every platform; a Windows-derived
-    # relative path carries os.sep ("\\") and would anchor nothing. Normalize.
-    patterns = [p.replace("\\", "/") for p in patterns]
+    # Callers pass POSIX-slash patterns (glob rels via as_posix; config strings as
+    # authored); git's exclude is POSIX-slash on every platform, so nothing to fix here.
     try:
         common = subprocess.run(
             ["git", "-C", str(worktree), "rev-parse", "--git-common-dir"],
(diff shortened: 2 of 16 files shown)


pbean:
will need the conflicts cleaned up after the two prior PRs were merged in.

dracic:
Please squash, this is ugly...

dracic:
This was a test-synchronization race - Textual's push_screen assigns app.screen before the modal's children finish mounting, so on a slow CI Windows runner the pilot.click("#ok") fired before #ok existed — fixed by having the until wait for the button to actually be mounted, not just for the screen type to change.  Do we wan't to leave all these python version checks?

pbean:
> This was a test-synchronization race - Textual's push_screen assigns app.screen before the modal's children finish mounting, so on a slow CI Windows runner the pilot.click("#ok") fired before #ok existed — fixed by having the until wait for the button to actually be mounted, not just for the screen type to change. Do we wan't to leave all these python version checks?

We really should keep them it is the version range we support. If flaky tests on additional PR's still keep being a thing with the ci runner, for windows only we can then consider removing the previous versions and add some disclaimer that while supported on the linux/macos/wsl side the windows side is not officially tested for the older versions.

dracic:
Then I will push more safeguards for until(screen is Modal) → immediate click here.

`await pilot.click("#ok") -> await pilot.click(await ready(pilot, "#ok"))`

pbean (changes requested):
**Validation ✅ — no Linux/macOS regression, approach is sound. Two small asks before I approve.**

Went through this focused on whether it risks the Linux/macOS core. It doesn't — confirmed three ways:
- All 4 existing Linux jobs (py3.11–3.14) stay green alongside the new Windows lane.
- The `os.kill(getpid())` → `signal.raise_signal` swap is equivalent on POSIX (single-threaded `engine.run()`, same main-thread handler) and stops `TerminateProcess` from killing the Windows runner.
- Every gated test's non-win32 branch is byte-identical or differs only by no-op quoting / `exit 0`↔`true` (equivalent under `/bin/sh`).

And the `_worktree_local_exclude` fix is a real one — `str(relative_to(repo_root))` renders `os.sep`, so `seed_globs` matches became `.claude\skills\<name>` on Windows and git never anchored them. Good catch. The `ready()`/`until()` TUI gates also cut flakiness on slow Linux runners — nice.

Two things I'd like tightened before I approve/merge:

1. **Scope the exclude normalization to its actual source.** The blanket `patterns = [p.replace("\\", "/") for p in patterns]` at the top of `_worktree_local_exclude` is broader than needed — only the glob-derived path picks up `os.sep`. Please normalize at the source instead — `seeded.append(rel.as_posix())` at `install.py:314` — and drop the loop. `skill_tree`/`config_path`/`seed_files` are already POSIX-slash config strings, so this keeps the fix precise and never rewrites a pattern that could (pathologically) carry a legitimate backslash on POSIX.

2. **De-dupe the platform helpers.** `_file_exists_cmd`, `_touch_run`/`_exists_run`/`_seeded_then_touch`, `_OK`/`_RUN`, etc. repeat the same `sys.platform == "win32"` branching across test files (CodeRabbit flagged this too). Please hoist the shared ones into `conftest.py` (next to `write_script_launcher`) so there's a single platform-detection spot.

Everything else looks great — the per-gate comments and the UTF-8 guard especially. Once those two land I'll approve; please squash on merge.

dracic:
Will address this tomorrow

dracic:
I ran a rerun, and I think I caught if not all, then at least the most critical tests that are candidates for flakiness.

pbean (approved):
**Approved ✅ — both asks verified landed.**

Re-validated the head (`ba12f18`) against my earlier review:

1. **Exclude normalization scoped to source** — the blanket `\\`→`/` loop in `_worktree_local_exclude` is gone (net diff vs main is comment-only there); the fix is now exactly `seeded.append(rel.as_posix())` at the glob source in `install.py`. Precise, and a provable no-op on POSIX.
2. **Platform helpers de-duped** — `_OK`/`_RUN`/`_file_exists_cmd`/`_touch_run`/`_exists_run`/`_seeded_then_touch` hoisted into `conftest.py` next to `write_script_launcher`; the engine test modules import them, so `sys.platform` branching lives in one spot. (Also closes CodeRabbit's duplication nitpick.)

Re-confirmed Linux/macOS safety on the final diff: every gated test's POSIX branch is byte-identical or shell-equivalent, and the extra `git status --short` regression assertion in `test_install.py` is a nice guard for the exclude fix. All 10 checks green on `ba12f18` (4 Linux + 4 Windows lanes, lint, version-sync).

Squash-merging as requested. Thanks for the fast, thorough turnaround — the Unity symlink-priming follow-up stays tracked separately.

PR #39, opened 2026-07-02
fix(cli): surface unknown liveness in resolve/delete/archive (win32)

### LINKED ISSUE #38: CLI resolve/delete/archive: surface unknown liveness (win32) — TUI/CLI parity follow-up to #36
## Problem

Follow-up to #36 (tri-state `unknown` liveness, fixing #33). **Native-Windows only**; POSIX (Linux/macOS) is unaffected.

#36 wired the `unknown` state (live-but-unreadable pid, e.g. win32 `ERROR_ACCESS_DENIED`) into the **TUI** guards:

- launch / resume / resolve treat `unknown`+pid as possibly-live (block or warn),
- delete / archive warn-not-block ("engine may still be live (unverifiable pid)").

The **CLI** still gates solely on the strict boolean `runs.engine_alive`, which reads `unknown` as `False`:

- `cmd_resolve` (`src/automator/cli.py:603`) — proceeds and launches a resolve driver against a possibly-live engine → **double-drive risk**. The TUI now refuses this exact case.
- `cmd_delete` (`src/automator/cli.py:826`) / `cmd_archive` (`src/automator/cli.py:851`) — proceed with **no warning** while the engine may still be writing to the run dir. The TUI now warns.

## Proposed fix

Mirror the TUI semantics, preserving the `runs.engine_alive` invariant ("`unknown` must not block stop/delete" — an unverifiable/recycled pid must never block cleanup forever):

- Add a tri-state read at the **runs layer** (e.g. `runs.engine_liveness(run_dir) -> 'alive'|'dead'|'unknown'`, thin wrapper over `ProcessHost.liveness_of`) so the CLI doesn't import TUI modules; `tui.data.liveness` can delegate its pid-backed branch to it.
- `resolve`: refuse on `unknown`+pid ("engine may still be live — stop it first"), like the TUI. Not a cleanup path, so the invariant doesn't apply.
- `delete` / `archive`: keep proceeding on `unknown`, but print a stderr warning ("engine may still be live (unverifiable pid)") before acting.

## Affected files

`src/automator/runs.py` (new tri-state helper), `src/automator/cli.py` (resolve/delete/archive), `src/automator/tui/data.py` (optional delegation), `tests/test_cli*.py`, `tests/test_runs.py`.

## Related

- #33 (parent: win32 unreadable-identity false-dead)
- #36 (TUI half of the fix)

dracic:
## What

Routes the CLI's `resolve` / `delete` / `archive` commands through a tri-state liveness read so a live-but-unreadable engine pid (`unknown`) is surfaced instead of silently read as dead — bringing the CLI to parity with the TUI half of the fix.

## Why

On native Windows a running engine whose identity is unreadable (`psutil.create_time()` → `ERROR_ACCESS_DENIED`) reads as tri-state `unknown`. The CLI gated only on the strict boolean `runs.engine_alive`, which collapses `unknown` to `False`, so `resolve` could launch a driver against a possibly-live engine (double-drive) and `delete`/`archive` proceeded with no warning while the engine may still be writing the run dir. POSIX is unaffected.

Fixes #38

## How

- Added `runs.engine_liveness` (plus a shared `probe_liveness` body) wrapping `ProcessHost.liveness_of`, so the CLI reads tri-state liveness without importing TUI modules; `unknown` is preserved and an unexpected probe failure degrades to `unknown`, never a false `dead`.
- `resolve` now refuses on `alive` **or** `unknown` (both possibly-live); `delete`/`archive` warn `engine may still be live (unverifiable pid)` on `unknown` but still proceed — an unverifiable/recycled pid must never block cleanup forever. A single liveness sample drives both the warning and the strict block, so a mid-check identity flip can't fire one without the other.
- Routed `tui.data.liveness` through the shared `probe_liveness`, keeping one copy of the pid-liveness logic and reading the pid file once.

## Testing

`pytest tests/test_runs.py tests/test_cli.py tests/test_tui_data.py tests/test_process_host.py` — added cases for the tri-state read (no-pid→dead, alive, reused→dead, unreadable→unknown, probe-raises→unknown), `resolve` refusing on `unknown`, `delete`/`archive` warn-but-proceed on `unknown`, and the TUI delegation regression. Full suite green (1303 passed; the only failures are a pre-existing skill-sync drift, unrelated to this change).

## Summary by CodeRabbit

* **New Features**
  * Added a `--force` option to `resolve`, allowing operators to continue when engine liveness is unverifiable (`unknown`) while still blocking provably-live engines.
* **Bug Fixes**
  * Engine liveness is now tri-state (live, dead, unverifiable), improving how resume/resolve/delete/archive/cleanup decisions are made.
  * `delete`, `archive`, `clean`, and cleanup/TUI flows now warn for unverifiable PIDs and proceed safely when appropriate.
  * `resume` now blocks provably-live engines, but warns and recovers when liveness is unverifiable.
* **Tests**
  * Updated and expanded CLI/TUI/run liveness coverage for the new tri-state behavior and `--force` handling.

### DIFF
--- a/src/automator/cli.py
+++ b/src/automator/cli.py
@@ -600,8 +600,11 @@ def cmd_resolve(args: argparse.Namespace) -> int:
             file=sys.stderr,
         )
         return 1
-    if runs.engine_alive(run_dir):
-        print(f"run {args.run_id} is still live — stop it first", file=sys.stderr)
+    # Not a cleanup path, so the "unknown must not block" invariant does not apply:
+    # an unverifiable-but-live pid must not be re-driven. Refuse on alive OR unknown.
+    if runs.engine_liveness(run_dir) != "dead":
+        # "may" not "is": the gate fires on unknown too, where liveness is unverifiable.
+        print(f"run {args.run_id} may still be live — stop it first", file=sys.stderr)
         return 1
     story_key = args.story or state.paused_story_key
     task = state.tasks.get(story_key) if story_key else None
@@ -815,6 +818,27 @@ def cmd_stop(args: argparse.Namespace) -> int:
     return 0
 
 
+def _stop_or_block_live_engine(run_dir: Path, run_id: str, force: bool) -> int | None:
+    """Shared delete/archive guard. One liveness sample drives both the warning and the
+    block, so a mid-check identity flip can't fire one without the other. An unverifiable
+    pid (``unknown``) warns but never blocks cleanup; only a provably-live engine blocks,
+    or is stopped first under ``force``. Returns an exit code to propagate, or None to
+    proceed with cleanup."""
+    live = runs.engine_liveness(run_dir)
+    if live == "unknown":
+        print(f"run {run_id}: engine may still be live (unverifiable pid)", file=sys.stderr)
+    if live == "alive":
+        if not force:
+            print(f"run {run_id} is still live — stop it first (or pass --force)", file=sys.stderr)
+            return 1
+        try:
+            runs.stop_run(run_dir)
+        except (runs.StopRunError, ProcessHostError) as e:
+            print(str(e), file=sys.stderr)
+            return 1
+    return None
+
+
 def cmd_delete(args: argparse.Namespace) -> int:
     project = _project(args)
     try:
@@ -823,18 +847,9 @@ def cmd_delete(args: argparse.Namespace) -> int:
         print(str(e), file=sys.stderr)
         return 1
     args.run_id = run_dir.name
-    if runs.engine_alive(run_dir):
-        if not args.force:
-            print(
-                f"run {args.run_id} is still live — stop it first (or pass --force)",
-                file=sys.stderr,
-            )
-            return 1
-        try:
-            runs.stop_run(run_dir)
-        except (runs.StopRunError, ProcessHostError) as e:
-            print(str(e), file=sys.stderr)
-            return 1
+    rc = _stop_or_block_live_engine(run_dir, args.run_id, args.force)
+    if rc is not None:
+        return rc
     runs.delete_run(run_dir)
     print(f"run {args.run_id} deleted")
     return 0
@@ -848,18 +863,9 @@ def cmd_archive(args: argparse.Namespace) -> int:
         print(str(e), file=sys.stderr)
         return 1
     args.run_id = run_dir.name
-    if runs.engine_alive(run_dir):
-        if not args.force:
-            print(
-                f"run {args.run_id} is still live — stop it first (or pass --force)",
-                file=sys.stderr,
-            )
-            return 1
-        try:
-            runs.stop_run(run_dir)
-        except (runs.StopRunError, ProcessHostError) as e:
-            print(str(e), file=sys.stderr)
-            return 1
+    rc = _stop_or_block_live_engine(run_dir, args.run_id, args.force)
+    if rc is not None:
+        return rc
     dest = runs.archive_run(project, run_dir)
     print(f"run {args.run_id} archived to {dest}")
     return 0
--- a/src/automator/cli.py
+++ b/src/automator/cli.py
@@ -928,7 +928,7 @@ def cmd_cleanup(args: argparse.Namespace) -> int:
         if live:
             print(f"leaving {len(live)} live session(s) untouched")
         return 0
-    killed = runs.prune_sessions(project)
+    killed, _ = runs.prune_sessions(project)
     windows = launch.prune_ctl_windows(project)
     print(f"removed {len(killed)} session(s), {len(windows)} ctl window(s)")
     if live:
--- a/src/automator/cli.py
+++ b/src/automator/cli.py
@@ -911,7 +911,11 @@ def cmd_cleanup(args: argparse.Namespace) -> int:
     from .tui import launch  # pure stdlib; no textual import
 
     project = _project(args)
-    prunable, live = runs.prunable_sessions(project)
+    prunable, live, unknown = runs.prunable_sessions(project)
+    for run_id in sorted(unknown):
+        # warn-only: unknown never blocks cleanup (same wording as delete/archive).
+        # Printed for --dry-run too — it describes state, not action.
+        print(f"run {run_id}: engine may still be live (unverifiable pid)", file=sys.stderr)
     if args.dry_run:
         windows = launch.prunable_ctl_windows(project)
         if not prunable and not windows:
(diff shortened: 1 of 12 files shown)


dracic:
Thanks for the reviews — addressed in the latest push:

- **`probe_liveness` blind `except` (CodeRabbit):** fixed. `get_process_host()` is now called outside the `try`, so a `ProcessHostError` misconfiguration (e.g. an invalid `BMAD_AUTO_PROCESS_HOST`) propagates loudly instead of being masked as `unknown` on every check. Only the actual `liveness_of()` probe still fails open to `unknown`. Added a regression test (`test_engine_liveness` now asserts the misconfig propagates).

- **`resolve` message overstates `unknown` (augment):** fixed. The gate fires on `alive` *or* `unknown`, so the message is now "run … **may** still be live — stop it first" (accurate for both; the action is unchanged).

- **`delete`/`archive` "(unverifiable pid)" wording (augment):** left as-is. That block message ("… is still live …") only prints on a **confirmed** `alive` engine; the `unknown` path prints the separate, already-hedged warning "engine may still be live (unverifiable pid)". So the cause attribution is correct where each message fires.

CI is green across all lanes (incl. Windows 3.11–3.14).

pbean:
@dracic thanks — this is a solid implementation of #38, and the review confirmed it does exactly what the issue asked with no POSIX behavior change (the full suite is green on Linux). I ran a deep review pass over it and, rather than round-trip a list of change requests, pushed three commits to your branch (maintainer edits are enabled) addressing what the review surfaced. Summary:

**502c5e8 — tri-state resolve/resume gates.** Two issues with the `!= "dead"` resolve gate as it stood:
1. Its remedy couldn't work: for a squatted pid with an unreadable identity, `stop` takes the mark-stopped fallback (`alive_and_ours` reads the squatter as not-ours) and `engine.pid` is never deleted — so "stop it first" looped back to the same refusal forever, with no escape. `resolve` now distinguishes the states: provably-`alive` always blocks (never bypassable), `unknown` blocks unless the operator vouches with a new `--force`, and the refusal message gives honest recovery guidance.
2. `resume` had no liveness gate at all, making the resolve gate trivially bypassable — and `resume` rewrites `engine.pid` and kills the agent session, i.e. the exact double-drive the gate exists to prevent. It now blocks on `alive` and warns-but-proceeds on `unknown` (it must stay usable there: rewriting the pid file is the recovery path for the unverifiable-pid state).

**58d17b2 — misconfig degrade + cleanup warnings.** `probe_liveness` letting `ProcessHostError` propagate is right for CLI decision paths, but pre-PR the TUI caught it (`get_process_host()` was inside the old try in `data.liveness`): post-PR a bad `BMAD_AUTO_PROCESS_HOST` crashed the whole TUI on the first dashboard poll tick (Textual worker `exit_on_error`) and broke `list`. `data.liveness` now catches `ProcessHostError` → `'unknown'` at the display layer only. Also extended your delete/archive warning to the other cleanup surfaces: `prunable_sessions` returns the unknown subset (classification unchanged — unknown still never blocks cleanup) so `cleanup` (CLI + TUI) and `clean` flag possibly-live engines before acting.

**d9f5dbe — dedup.** `alive_and_ours` is now literally `liveness_of(pid, identity) == "alive"` (provably equivalent on every branch), and the test `_FakeHost` subclasses `ProcessHost` so both derivations are inherited instead of hand-copied — the binary and tri-state probes can no longer drift apart.

Full suite: 1314 passed, 1 skipped; trunk clean. Happy to walk through any of it — and if you disagree with a direction (the `--force` semantics in particular), say so and we can adjust before merge.

pbean:
Follow-up: the first CI round failed on all four native-Windows lanes — `test_alive_and_ours_matches_only_same_identity` stubbed `identity` but not `is_alive`, and the derived `alive_and_ours` now probes `is_alive` on the unreadable-identity branch, so the real `PosixProcessHost.is_alive` fixture hit `os.kill(4242, 0)` → `WinError 87` on win32 runners (production unaffected — win32 selects `WindowsProcessHost`/psutil). Fixed in 6a08feb by stubbing `is_alive` like the adjacent `liveness_of` test and asserting not-ours for both sub-branches (gone, and live-but-unreadable). CI is now fully green across all lanes: https://github.com/bmad-code-org/bmad-auto/actions/runs/28622823911

pbean:
augment review

pbean:
Addressed the two new review comments in caa3eb9 + 8a49cea:

**caa3eb9 — cleanup unknown-pid warning tied to the pruned partition (augment, `tui/app.py:566`).** The TOCTOU was real: the notification's `unknown` set came from a `prunable_sessions()` sample taken before `prune_sessions()` re-partitioned internally, so a liveness flip between the two samples could make "pruned session(s) had an unverifiable engine pid" name sessions that were never pruned (or miss ones that were). Rather than soften the wording, `prune_sessions` now returns `(killed, unknown)` from its own single partition sample and the TUI worker drops its pre-sample — same "one sample drives both the warning and the action" principle as `_stop_or_block_live_engine` (502c5e8). `cmd_cleanup` keeps its pre-prune warning: it's needed for `--dry-run` and its wording describes state rather than action, so it survives the race either way. Regression test: `test_prune_sessions_returns_unknown_from_same_sample`.

**8a49cea — force-stop error-path tests (CodeRabbit, `cli.py:857-877`).** Added `test_delete_force_stop_error_blocks` (`StopRunError` via `delete`) and `test_archive_force_stop_error_blocks` (`ProcessHostError` via `archive`) — both caught exception types and both commands' wiring through the shared helper, asserting exit 1, the error on stderr, and the run dir untouched.

Full suite green locally (1317 passed) and trunk clean.

pbean:
@coderabbit review
augment review

pbean:
Addressed the augment follow-up in eeb1f87:

**`cmd_cleanup` double-sample (augment, `cli.py:931`).** Fair catch — my earlier state-not-action rationale covered the unknown *warnings* but not the `live` count: "left N live session(s) untouched" is an action claim printed from a pre-prune sample, so a live→dead flip between the two samples could report a pruned session as left untouched (and a dead→unknown flip could kill a session with no warning). Rather than warn from the returned `unknown` on just the non-dry-run path, `prune_sessions` now returns the full `(killed, live, unknown)` partition and `cmd_cleanup` routes **both** paths through a single call (`dry_run=` suppresses the kill) — every message now derives from the same sample that drove the action, and the CLI no longer touches `prunable_sessions` directly. Net simplification; this also retires the "needed for --dry-run" argument, since the dry-run path gets the same partition back.

Full suite green locally (1317 passed) and trunk clean.

pbean:
augment review

Issue #40, opened 2026-07-02
Multiplexer `_run` seam: let subclasses set output encoding and per-call env without reimplementing the spawn primitive


dracic:
## Background

The terminal-multiplexer backends share a base class, `BaseTmuxBackend` (`src/automator/adapters/tmux_base.py`), that funnels every `tmux` invocation through a single subprocess primitive:

```python
def _run(self, argv: list[str], *, check: bool = True) -> subprocess.CompletedProcess[str]:
    proc = subprocess.run(
        ["tmux", *argv], capture_output=True, text=True, timeout=TMUX_TIMEOUT_S
    )
    if check and proc.returncode != 0:
        raise TmuxError(f"tmux {' '.join(argv[:2])} failed: {proc.stderr.strip()}")
    return proc
```

The POSIX `TmuxMultiplexer` uses this as-is. The `_run` docstring already advertises the extension point:

> Subclasses (e.g. a native-Windows psmux) override this to tweak the binary, decoding (`encoding="utf-8"`), or timeout — and nothing else.

## Problem

A native-Windows backend (a `psmux` `tmux.exe` drop-in) needs two things the seam does **not** actually expose, despite the docstring:

1. **Output decoding.** Windows defaults to cp1252, which mangles the UTF-8 that psmux emits for `-F` format output and multibyte field separators — so every read must force `encoding="utf-8"`. The base `_run` hard-codes `text=True` with no `encoding`, so there is no hook to set it.
2. **Per-call `env`.** Creating a session from *inside* an existing multiplexer pane requires spawning `new-session` with a scrubbed environment (e.g. dropping a nesting-guard variable) **without** mutating the parent process's env. The base `_run` passes no `env`, so there is no hook for this either.

Because neither is exposed, a subclass that needs *either* is forced to **re-implement the whole `subprocess.run(...)` primitive** — duplicating the `capture_output`/`text`/`timeout` setup, the `check` handling, and the `TmuxError` raise — and to widen the method signature ad-hoc. That is exactly the kind of copy-paste-with-drift the single-seam design was meant to prevent: any later change to timeout/error handling in the base silently fails to reach the override.

## Proposed resolution

Make the seam carry the two knobs it already promises, with defaults that are byte-for-byte the current behavior:

1. **Output encoding as a class attribute.**
   ```python
   class BaseTmuxBackend(TerminalMultiplexer):
       _ENCODING: str | None = None  # system default (today's behavior); a Windows leaf sets "utf-8"
[...]

PR #47, opened 2026-07-03
refactor(adapters): extract the shell-dialect seam from new_window/new_parked_window

### LINKED ISSUE #46: Tmux-family base: factor the shell-dialect wrapper out of new_window/new_parked_window into overridable hooks
**Describe your idea**

`BaseTmuxBackend.new_window` / `new_parked_window` (`src/automator/adapters/tmux_base.py`) hard-code the POSIX shell wrapper: `sh -c`, `shlex` quoting, `$?` exit capture, `echo`, `read -r`, and an sh return-trailer. Factor those shell-*dialect* fragments into small overridable hooks with POSIX defaults, so the tmux argv construction, the parked-window protocol, and the recipe shape (`inner; exit-capture; banner; park; return-trailer`) live in the base exactly once.

**Why is this needed?**

A native-Windows backend (the `psmux` `tmux.exe` drop-in, follow-up to #40) has no `/bin/sh` — its windows run under `pwsh`. With the wrapper hard-coded, a Windows subclass is forced to duplicate **both entire methods** just to swap the shell dialect, even though the tmux verbs and argv layout are identical (`new-window -t =S: -n NAME -c CWD -P -F '#{window_id}' …`, and the trailer's `show-options`/`switch-client`/`detach-client` hit `tmux.exe` unchanged). A prototype Windows backend did exactly that, and the two copies immediately became the biggest drift risk in the adapter layer: any later fix to the parked protocol in the base silently fails to reach the override.

**How should it work?**

Keep both contract methods in the base and route only the dialect through hooks (POSIX defaults shown; a Windows leaf overrides the strings, never a method body):

```python
_EXIT_CAPTURE = "ec=$?"          # pwsh later: "$ec = $LASTEXITCODE"
_ECHO = "echo"                   # pwsh later: "Write-Host"
_PARK = "read -r"                # pwsh later: "Read-Host"

def _join_argv(self, argv): ...      # shlex.join    | pwsh: & 'a' 'b' with '' escaping
def _parked_trailer(self, opt): ...  # sh if/elif    | pwsh: if/elseif (verbs identical)
def _source_prefix(self): ...        # ""            | pwsh: env-hygiene prelude
def _shell_wrap(self, source): ...   # ["sh","-c",s] | pwsh: -NoProfile -EncodedCommand
def _window_launch(self, env, cmd):  # ["-e","K=V"…, cmd] | pwsh: wrapped cmd + prelude
```

`new_parked_window` composes `_source_prefix() + f"{inner}; {exit}; {echo} \"[bmad-auto exited $ec — press enter]\"; {park}; {trailer}"` and spawns `*self._shell_wrap(source)`. The recipe line is genuinely dialect-neutral (`;` sequencing and `"$ec"` interpolation behave the same in sh and pwsh); only the trailer stays a whole-string hook because sh/pwsh control-flow syntax can't be tokenized.

Guarantees:
- POSIX output stays **byte-identical** — locked by tests asserting the literal `subprocess.run` argv (including the full `sh -c` source) before/after.
- `tmux_backend.py` (the POSIX leaf) overrides nothing new.
- The pwsh side is deliberately **not** implemented here; this only opens the seam for the Windows backend to land against.

**PR**

I'm working on this — implementation with byte-identity + fake-dialect-leaf tests is ready on my fork; PR to follow.

**Additional context**

Companion to #40 (which widened the `_run` spawn seam for encoding/env the same way: default-preserving hooks instead of method duplication). Together they get a tmux-family Windows leaf to ~zero copied plumbing.

dracic:
## What

`BaseTmuxBackend.new_window`/`new_parked_window` no longer hard-code the POSIX shell wrapper: the tmux argv construction, parked-window protocol, and recipe shape stay in the base once, while the shell-*dialect* fragments become small overridable hooks with POSIX-sh defaults.

## Why

A native-Windows tmux-family leaf (follow-up to #40) has no `/bin/sh`; with the wrapper hard-coded it had to duplicate both entire methods just to swap the shell dialect — the biggest drift risk in the adapter layer. 

Fixes #46

## How

- Added dialect hooks on `BaseTmuxBackend`: `_EXIT_CAPTURE`/`_ECHO`/`_PARK` fragments, `_join_argv`, `_parked_trailer`, `_source_prefix`, `_shell_wrap`, `_window_launch` — each maps 1:1 to a divergence a non-POSIX leaf actually needs; the recipe line itself is dialect-neutral (`;` sequencing and `"$ec"` interpolation are common across the family).
- Rewrote `new_window`/`new_parked_window` to compose from the hooks; the POSIX leaf (`tmux_backend.py`) overrides nothing and its output is byte-identical.
- Documented the hook contract (an `_EXIT_CAPTURE` override must bind `ec`; `_source_prefix` must self-terminate with `"; "`; which hooks a non-POSIX leaf must override). No non-POSIX implementation lands here — this only opens the seam.

## Testing

New tests lock the exact pre-refactor `subprocess.run` argv (including the full `sh -c` source literal) for both methods, and a fake-dialect test leaf proves overriding only the hooks keeps the base scaffolding intact. `tests/test_multiplexer.py` 11/11; full suite green on Windows against the known baseline (`pytest -k "not real_signal"`, 1322 passed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved POSIX tmux window and parked-window launch command assembly for more consistent behavior.
  * Refined shell-dialect handling so dialect-specific overrides are applied cleanly, including environment flag generation.

* **Tests**
  * Added stricter coverage to confirm byte-identical command argv construction and correct routing through shell-dialect hooks.

* **Documentation**
  * Updated adapter authoring and porting guidance to clarify the supported tmux shell-dialect customization surface.

### DIFF
--- a/docs/adapter-authoring-guide.md
+++ b/docs/adapter-authoring-guide.md
@@ -55,7 +55,9 @@ To add a backend, build a `TerminalMultiplexer` (`adapters/multiplexer.py`) and
 returns the first backend whose `matches` is true, with tmux as the default
 fallback. There are two build paths: extend `BaseTmuxBackend` (`adapters/tmux_base.py`)
 for a tmux-family backend — overriding only its single spawn primitive `_run()`
-plus the few divergent methods (e.g. the parked-window trailer) — or implement
+plus the shell-dialect hooks (`_shell_wrap`, `_join_argv`, `_parked_trailer`,
+`_source_prefix`, `_window_launch` and the `_EXIT_CAPTURE`/`_ECHO`/`_PARK`
+fragments) — or implement
 `TerminalMultiplexer` fresh for a host with no tmux-shaped CLI. The non-transport
 seams of a full OS port are in
 [Porting bmad-loop to a new OS](porting-to-a-new-os.md). The contract groups into:
@@ -66,9 +68,9 @@ seams of a full OS port are in
   (read a user option across all sessions), `set_session_option`.
 - **Windows** — `new_window` (run a command in a fresh window), `new_parked_window`
   (run a command, then _park_ on a keypress so the exit status stays inspectable,
-  then return any attached client to its origin — this is where the POSIX `sh -c`
-  trailer is quarantined; a non-POSIX backend reimplements the same behavior in
-  its own terms), `list_window_ids`, `list_windows` (selected fields per window),
+  then return any attached client to its origin — the POSIX `sh -c` recipe is
+  composed from the base's overridable shell-dialect hooks, so a non-POSIX
+  backend swaps the dialect fragments, not the method body), `list_window_ids`, `list_windows` (selected fields per window),
   `window_alive`, `kill_window`, `select_window`, `set_window_option`,
   `unset_window_option`, `show_window_option`, `pipe_pane` (tee a pane to a log),
   `send_text`.
--- a/docs/porting-to-a-new-os.md
+++ b/docs/porting-to-a-new-os.md
@@ -63,10 +63,13 @@ fallback, so POSIX behavior is unchanged. (The result is cached — see
   through one primitive, `_run(argv, *, check=..., env=...)`. A native-Windows
   "psmux" that speaks a tmux-like CLI sets the `_ENCODING` class attribute for
   output decoding (e.g. `"utf-8"`) and passes a per-call `env=` where needed —
-  overriding `_run()` itself only to tweak the binary or timeout — plus the few
-  genuinely divergent methods (e.g. the
-  parked-window `sh -c` trailer in `new_parked_window`) — **without editing**
-  `tmux_base.py` or its POSIX leaf `tmux_backend.py` (`TmuxMultiplexer`).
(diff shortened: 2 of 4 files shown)


pbean:
Reviewed — the seam is sound: POSIX argv/source verified byte-identical, the leaf overrides nothing, all call sites unaffected. Pushed `ccb042a` directly (maintainer edit) to fold in the three doc-level review nits rather than round-tripping comments:

- `tmux_base.py` module docstring: scoped the *"without editing … any contract method body"* claim to `new_window`/`new_parked_window` — `pipe_pane` still hands tmux a POSIX `cat >>` redirection, so it remains the one method-body override a non-POSIX leaf needs.
- `docs/adapter-authoring-guide.md` + `docs/porting-to-a-new-os.md`: both still described the pre-seam model ("override the few divergent methods, e.g. the parked-window trailer"); now point at the dialect hooks.

No code-behavior changes; `tests/test_multiplexer.py` 11/11 and the full suite green locally.

Issue #49, opened 2026-07-03
Cap the lifecycle of refs/attempt-preserve-dirty/* worktree-snapshot refs (follow-up to #32)


dracic:
## Background

#29 introduced two families of local recovery refs written before an auto-rollback's `git reset --hard`:

1. `refs/heads/attempt-preserve/<run_id>-<sha>` branches parking **committed** work (`preserve_commits`), and
2. `refs/attempt-preserve-dirty/<run_id>-<baseline>-<attempt>` refs parking the **uncommitted** working-tree snapshot (`snapshot_worktree` via `_preserve_attempt_worktree`).

#32 added bounded retention for family 1 (`ScmPolicy.preserve_keep`, pruned at run start) and explicitly fenced family 2 out of its scope.

## Problem

Nothing ever removes the `refs/attempt-preserve-dirty/*` refs. Every dirty auto-rollback attempt creates one (keyed by attempt number, so retries each add a new ref), and they accumulate unbounded exactly like the branches did.

**Scope / severity: low.** Same as #32: these refs are local-only (never pushed) and, living outside `refs/heads/`, they don't even clutter `git branch` output — so the nuisance is smaller. But they still add ref-walking overhead forever and hold snapshot commits alive against `git gc` indefinitely.

## Proposed resolution

Extend the run-start bounded retention to this family:

- Reuse the existing `scm.preserve_keep` budget (simplest — one knob governs both recovery-ref families), or add a sibling field if independent tuning is wanted.
- These are not branches, so delete via `git update-ref -d` (or batched `git update-ref --stdin`), not `git branch -D`.
- `for-each-ref --sort=-committerdate refs/attempt-preserve-dirty/` works unchanged: each ref points at the snapshot commit created at preservation time, so committer date *is* the park time here.
- Keep the #32 hardening: per-ref best-effort deletes (one stuck ref must not wedge the tail), full-refname handling, prune failures journalled and never blocking run start.

## Acceptance criteria

- [ ] `refs/attempt-preserve-dirty/*` refs are pruned to a bounded count at run start, keeping the most recently created.
- [ ] Retention is configurable; `0`/disabled keeps every ref (consistent with `preserve_keep` semantics).
- [ ] Pruning only ever touches `refs/attempt-preserve-dirty/*` — never `refs/heads/*` (the #32 prune already owns `refs/heads/attempt-preserve/*`).
[...]

PR #50, opened 2026-07-03
feat(scm): bounded retention for attempt-preserve recovery refs

### LINKED ISSUE #32: Cap the lifecycle of attempt-preserve/* recovery refs (follow-up to #29)
## Background

#29 adds `attempt-preserve/<run_id>-<sha>` recovery branches that park an in-place attempt's committed work at HEAD before an auto-rollback runs `git reset --hard <baseline>`, so the commits survive `git gc` and stay recoverable by name. That's a correct, valuable safety fix.

## Problem

Nothing ever removes those refs. Every auto-rollback that had committed work leaves a permanent `attempt-preserve/*` branch, so in a long-lived project with `scm.rollback_on_failure` on they grow unbounded — cluttering `git branch` output and eventually adding overhead to ref-walking operations.

**Scope / severity:** low. There is no `git push` anywhere in `src/automator` (in-place mode never pushes; worktree mode merges locally), so these refs are **local-only** — repo clutter, not a remote leak. But the unbounded growth is still worth capping.

## Proposed resolution

Mirror the retention patterns already in the codebase (`runs.prune_sessions` / `keep_n`, and the `delete_branch` / `keep_failed` config style in `ScmPolicy`):

1. **Bounded-retention prune.** Keep the N most-recent `attempt-preserve/*` refs and delete the tail. A natural home is run start, alongside `runs.prune_sessions`:
   ```
   git for-each-ref --sort=-committerdate \
     --format='%(refname:short)' refs/heads/attempt-preserve/
   # keep the first N; `git branch -D` the rest
   ```
   Sort by `committerdate` so the most recently preserved work is the last thing pruned. Default N to a safely non-zero value so an aggressive default can't defeat the safety net.
2. **Expose it as an `ScmPolicy` field** to match the existing config surface, e.g. `preserve_keep: int = 20`. Operators who want zero clutter can lower it; those who want maximum safety can raise it. `0` could mean "never prune".
3. *(Optional)* a `bmad-auto` maintenance subcommand to list/prune `attempt-preserve/*` refs on demand.

## Acceptance criteria

- [ ] `attempt-preserve/*` refs are pruned to a bounded count on a normal orchestrator lifecycle event (e.g. run start), keeping the most recently preserved refs.
- [ ] Retention is configurable via `ScmPolicy` with a safe non-zero default; `0`/disabled keeps every ref.
- [ ] Pruning only ever touches `refs/heads/attempt-preserve/*` — never a user or unit branch.
- [ ] Tests cover: keep-N boundary, ordering by `committerdate`, disabled/`0` = no-op, and that non-`attempt-preserve` branches are untouched.

## Notes

- Non-blocking follow-up to #29 — that PR is good to merge as-is; this just caps the lifecycle of the refs it introduces.
- Related minor observation from the #29 review: `_preserve_attempt_commits` wraps `preserve_commits` in `try/except verify.GitError` but not the leading `commits_above` call. Not a bug (baseline is trusted; the base `reset --hard` already fails on a bad baseline), but could be unified for a uniform failure surface if touched here.

dracic:
## Summary

Closes #32

Auto-rollback parks committed attempt work under `attempt-preserve/*` branches (#29), but nothing ever removed them — with `scm.rollback_on_failure` on they grew unbounded. This adds bounded retention at run start, mirroring the existing retention knobs:

- **`verify.prune_preserve_refs(repo, keep)`** — lists `refs/heads/attempt-preserve/` via `for-each-ref --sort=refname --sort=-committerdate` (newest by committer date kept; ties break deterministically by ascending refname) and force-deletes the tail. Full refnames (not `:short`) so a tag sharing the name can't derail the delete.
- **`ScmPolicy.preserve_keep`** (default 20, `0` = never prune) — strictly typed at policy load (`PolicyError` on bool/float/string/negative, so a TOML typo can't silently shrink a safety budget), plus settings-UI field and policy-template line.
- **Engine hook at run start** — best-effort: any failure is journalled (`attempt-preserve-prune-failed`) and never blocks or crashes the run.

Hardening from adversarial review:

- Per-ref best-effort deletes: one stuck ref (e.g. checked out for inspection) never wedges the refs behind it. All deletions are attempted first; `PrunePreserveError(GitError)` then carries `.deleted`/`.failed` so the engine journals the destructive half structurally (`attempt-preserve-pruned`) alongside the failure event — a partial prune is never auditable only via an error string.
- `refs/attempt-preserve-dirty/*` snapshot refs are untouched by design; their symmetric retention gap is tracked in #49.

## Tests

11 new/extended tests: keep-N boundary, committer-date ordering proven with dates deliberately disagreeing with creation order, refname tie-break, keep=0 no-op, non-preserve branches untouched, stuck-ref partial failure (structured attrs + both journal events), engine wiring at run start, prune failure never blocking the run, and strict config typing (`true`/`1.9`/`"5"`/`-1` all rejected).

Verified on Windows (`201 passed`, ruff clean) and Linux/WSL (`202 passed`, includes the signal test Windows skips).

### DIFF
--- a/src/bmad_loop/data/settings/core.toml
+++ b/src/bmad_loop/data/settings/core.toml
@@ -252,6 +252,13 @@ default_ref = "ScmPolicy.rollback_on_failure"
 label = "auto-rollback failed attempts"
 description = "⚠ in-place mode (isolation=none): when ON, a failed attempt's tracked changes are auto-reverted and the untracked files this run created are deleted (its uncommitted work is lost). When OFF (default), the orchestrator never touches your tree — it pauses with manual recovery steps. Governs unattended/stopped attempts only: a resolved escalation's re-drive always auto-recovers regardless (reverts the failed source, keeps the corrected spec). Prefer isolation=worktree to keep failures off your main checkout."
 [[section.field]]
+key = "preserve_keep"
+kind = "int"
+minimum = 0
+default_ref = "ScmPolicy.preserve_keep"
+label = "preserve-ref retention"
+description = "attempt-preserve/* recovery branches kept at each run start · the newest by committer date survive and the tail is deleted (0 = never prune)"
+[[section.field]]
 key = "seed_adapter_defaults"
 kind = "switch"
 default_ref = "ScmPolicy.seed_adapter_defaults"
--- a/src/bmad_loop/engine.py
+++ b/src/bmad_loop/engine.py
@@ -215,6 +215,7 @@ class Engine:
                 # repo), so it must sit inside the pause handler, not before it.
                 self._emit_run_boundary("pre_run")
                 self._ensure_target_branch()
+                self._prune_preserve_refs()
                 self._loop()
                 self.state.finished = True
                 self._gc_run_worktrees()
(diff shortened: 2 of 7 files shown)


pbean (approved):
Approving after local validation on the PR branch.

**Verification (reproduces green CI):**
- Full suite on Linux (py3.13): **1344 passed, 7 skipped, 0 failures**.
- `trunk check --upstream main`: **no issues** across all modified files.
- New tests run directly and pass — 6 `prune_preserve_refs` cases in `test_verify.py` (delete-beyond-keep, at/under-budget no-op, `keep=0` never-prunes, ignores-other-branches, tie-break-by-refname, continues-past-undeletable-ref), 3 engine run-start wiring cases in `test_engine.py` (prune-on-run-start, prune-failure-journals-and-proceeds, partial-prune-journals), and `test_scm_preserve_keep_settings` in `test_policy.py`.

**Integration seams cross-checked against core — all clean:**
- `verify._git` caller-checks-rc contract and `delete_branch(..., force=True)` (`git branch -D`) used correctly.
- Refs live at `refs/heads/attempt-preserve/*` (created via `git branch -f`); `removeprefix("refs/heads/")` yields valid `branch -D` names, and the prune only ever touches that namespace.
- Engine hook placed after `_ensure_target_branch()`; the broad `except` correctly absorbs the `TimeoutExpired` that `_git` can leak, so a prune failure journals and the run proceeds.
- `self.workspace.root` is the same repo where `_preserve_attempt_commits` parks the refs.
- `preserve_keep` typing/validation mirrors the existing `max_parallel`/`failed_diff_max_mb` `PolicyError` convention; policy template + settings-UI field kept in sync (`minimum = 0` is None-checked, no falsy-zero bug).

**Windows-safe:** list-argv subprocess (no shell), git ref syntax (not filesystem paths), `removeprefix` (requires-python >=3.11). Windows CI (py3.11 + py3.14) is green.

**Issue #32 acceptance criteria — all met:** bounded prune at run start; configurable with a safe default; only `refs/heads/attempt-preserve/*` is touchable; all required test cases present (plus tie-break and partial-failure extras).

**Notes:**
- Pruning is **on by default** (`preserve_keep = 20`) — previously these refs were kept forever. This is the behavior #32 asked for (safe non-zero default; `0` = never prune), just worth a CHANGELOG mention at release time.
- The symmetric `refs/attempt-preserve-dirty/*` snapshot retention gap is deliberately out of scope here and is tracked in #49.

PR #54, opened 2026-07-03
feat(scm): bounded retention for attempt-preserve-dirty worktree-snapshot refs

### LINKED ISSUE #49: Cap the lifecycle of refs/attempt-preserve-dirty/* worktree-snapshot refs (follow-up to #32)

dracic:
> [!NOTE]
> **Stacked on #50** — this branch is based on `feat/preserve-ref-retention` and reuses its `_prune_refs`/`preserve_keep` machinery. Review only the top commit; the diff collapses to it once #50 merges. Marked draft until then.

## What

Extends the run-start bounded retention from #50 to the second recovery-ref family: `refs/attempt-preserve-dirty/*` worktree-snapshot refs (written before each dirty auto-rollback) now share the same `scm.preserve_keep` budget — newest by committer date survive, tail deleted, `0` = never prune. No new config knob.

Closes #49

## How

- `verify.py`: the #32 retention loop is extracted into one shared `_prune_refs` helper; the branch family delegates to it unchanged (public behavior byte-identical, existing tests untouched). New `prune_preserve_dirty_refs` lists `refs/attempt-preserve-dirty/` and deletes via `git update-ref -d` — these are not branches, so no `branch -D`. Per-ref deletes are best-effort across the whole subprocess surface (a timeout/OSError on one ref no longer wedges the tail), with `PrunePreserveError.deleted/.failed` keeping partial prunes structurally auditable.
- `engine.py`: the run-start hook loops both families; each journals its own `*-pruned` / `*-prune-failed` events (stuck refs now journalled as a structured `failed` field), and a failure in one family never skips the other.
- `preserve_keep` docs (policy comment, TOML template, settings UI) updated — they previously claimed dirty snapshots were never touched.

## Tests

8 new tests: keep-N boundary, committer-date ordering, refname tie-break, keep=0 proves zero git invocations, branches/tags untouchable, stuck ref via a real `.lock` file (rest of tail still deleted, structured error attrs), engine wiring at run start, and family independence (branch-family failure does not skip the dirty prune).

Verified on Windows: `pytest tests/test_verify.py tests/test_engine.py tests/test_policy.py` — 209 passed; ruff clean.

## Summary by CodeRabbit

* **New Features**
  * Expanded retention handling to cover both recovery branches and dirty worktree snapshots at run start.
  * Added clearer status reporting when pruning succeeds or fails, including separate outcomes for each ref family.

* **Bug Fixes**
  * Pruning now continues for the remaining ref family even if one family fails.
  * Retention behavior is clearer and more consistent, including the “0 means never prune” setting.

### DIFF
--- a/src/bmad_loop/data/settings/core.toml
+++ b/src/bmad_loop/data/settings/core.toml
@@ -257,7 +257,7 @@ kind = "int"
 minimum = 0
 default_ref = "ScmPolicy.preserve_keep"
 label = "preserve-ref retention"
-description = "attempt-preserve/* recovery branches kept at each run start · the newest by committer date survive and the tail is deleted (0 = never prune)"
+description = "attempt-preserve/* recovery branches and attempt-preserve-dirty/* worktree snapshots kept at each run start (per family) · the newest by committer date survive and the tail is deleted (0 = never prune)"
 [[section.field]]
 key = "seed_adapter_defaults"
 kind = "switch"
--- a/src/bmad_loop/engine.py
+++ b/src/bmad_loop/engine.py
@@ -848,30 +848,41 @@ class Engine:
         )
 
     def _prune_preserve_refs(self) -> None:
-        """Bounded retention for the attempt-preserve/* recovery branches at run
-        start: keep the newest scm.preserve_keep by committer date, delete the
-        tail (mirrors the runs/cleanup retention knobs — without it the refs
-        grow unbounded on a long-lived project). Best-effort: a git failure is
-        journalled and never blocks or pauses the run — the refs are a safety
-        net, not run state. preserve_keep = 0 disables pruning entirely."""
+        """Bounded retention for both recovery-ref families at run start — the
+        attempt-preserve/* branches and the refs/attempt-preserve-dirty/*
+        worktree snapshots: keep the newest scm.preserve_keep of each by
+        committer date, delete the tail (mirrors the runs/cleanup retention
+        knobs — without it the refs grow unbounded on a long-lived project).
+        Best-effort: a git failure is journalled per family and never blocks or
+        pauses the run — the refs are a safety net, not run state — and a
+        failure in one family never skips the other. preserve_keep = 0 disables
+        pruning entirely."""
         keep = self.policy.scm.preserve_keep
         if keep <= 0:
             return
-        try:
-            deleted = verify.prune_preserve_refs(self.workspace.root, keep)
-        except Exception as exc:  # noqa: BLE001 - housekeeping must never crash the
-            # run: a git timeout/OSError here would otherwise escape to the crash
-            # handler, so anything beyond the expected GitError is journalled too
-            # A partial prune (PrunePreserveError) already deleted refs before one
-            # stuck — that destructive half must stay structurally auditable, not
-            # buried in the error string.
-            partial = getattr(exc, "deleted", [])
-            if partial:
-                self.journal.append("attempt-preserve-pruned", count=len(partial), refs=partial)
-            self.journal.append("attempt-preserve-prune-failed", error=str(exc))
-            return
-        if deleted:
-            self.journal.append("attempt-preserve-pruned", count=len(deleted), refs=deleted)
+        for family, prune in (
+            ("attempt-preserve", verify.prune_preserve_refs),
+            ("attempt-preserve-dirty", verify.prune_preserve_dirty_refs),
+        ):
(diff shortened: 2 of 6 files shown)


pbean (approved):
Validated end-to-end against post-#50 `main` — approving.

**Correctness**: prefix isolation holds by construction (the writer's slug is sanitized to `[alnum_-]`, so `refs/attempt-preserve-dirty/` listing can never pick up nested or foreign refs); committer-date-as-park-time is genuinely valid since `snapshot_worktree` builds the snapshot commit via `commit-tree` at preservation time; `update-ref -d` on full refnames is the right deletion primitive for bare refs. The broadened per-ref `except Exception` is a real fix — `_git` runs with `timeout=GIT_TIMEOUT_S`, and `TimeoutExpired` is not a `GitError`, so under #50 one timed-out delete would have wedged every tail ref behind it.

**Core safety**: swept the codebase for consumers — the dirty refs are write-only breadcrumbs (no resume/re-drive/resolve/sweep/TUI path reads them back; resumed attempts always write new attempt-keyed names), and the journal is schema-free with every consumer (diagnostics histogram/scrubber, TUI `journal_line`) tolerant of the new kinds and the new `failed=` field. The still-paused-older-run retention exposure is the same tradeoff #32/#50 already accepted for the branch family, with `preserve_keep = 0` as the documented opt-out.

**Portability**: all git calls stay on the native, timeout-bounded, shell-free `subprocess` path — no POSIX-only surface; Windows CI green plus your local Windows run.

Two non-blocking notes:

1. PR-body nit: "public behavior byte-identical" for the branch family is slightly overstated — the per-ref `except GitError` → `except Exception` broadening does change branch-family behavior (a timeout/OSError on one ref no longer aborts the tail). It's the right change and the body discloses it a sentence later; just flagging the phrasing for future archaeology.
2. The `.lock`-file stuck-ref tests (this PR's and #50's) assume git's *files* ref backend — under reftable (`refs.format=reftable`, opt-in today) a stale loose-ref lock won't block `update-ref -d`. Not worth changing now; maybe worth a one-line comment in the tests so the failure is self-explaining if/when reftable becomes a default.

PR #60, opened 2026-07-03
docs(adapters): pin new_window command contract as shlex-joined argv

### LINKED ISSUE #59: Multiplexer seam: new_window's command parameter semantics differ between backends and are unspecified
## Summary

The `Multiplexer.new_window` abstract method (`src/bmad_loop/adapters/multiplexer.py`) takes a `command: str`, but its docstring never pins down what that string is — a shell command line, or a POSIX-quoted argv join. The two backends currently interpret it differently:

- **tmux path** (`tmux_base.py` `_window_launch`): hands `command` to tmux verbatim as the window's shell command. Shell operators like `&&` or `|` would be interpreted by the shell.
- **psmux path** (`psmux_backend.py` `_window_launch`): `shlex.split`s the string and re-emits each token individually quoted for pwsh. Shell operators become literal arguments (this behavior is test-pinned).

## Impact

Low today: both existing callers (`generic.py` `build_command`, `probe.py` `_ProbeLauncher.start`) build `command` with `" ".join(shlex.quote(a) for a in argv)`, so the two paths behave identically in practice.

But the contract gap means a future caller passing a genuine shell line (e.g. `"foo && bar"`) would work on the POSIX/tmux backend and silently break on Windows/psmux only — the kind of divergence that surfaces late and platform-specifically.

## Proposed fix

One-line contract clarification in the `new_window` docstring: `command` is a shlex-joined argv (POSIX-quoted), shell operators are unsupported. Optionally assert the same in the POSIX leaf's tests so the tmux path is pinned symmetrically with the psmux one.

dracic:
## Summary

`Multiplexer.new_window`'s `command: str` parameter had unspecified semantics: the POSIX/tmux backend hands it to a shell (operators like `&&` would be interpreted) while a backend that re-splits the string treats operators as literal argv. Both existing callers build the command with `shlex.quote`-joins, so nothing is broken today — but a future caller passing a real shell line would break platform-specifically.

Docs-only contract pin, no behavior change:

- **`multiplexer.py`** — the `new_window` docstring now states the contract: `command` is a POSIX shlex-joined argv string, not a shell line; shell-operator behavior is backend-defined and callers must not rely on it.
- **`test_multiplexer.py`** — one new test pins the POSIX leaf's side of the contract: an operator-bearing shlex-joined command reaches tmux verbatim as the single trailing argument (full argv asserted, symmetric with the existing byte-identity tests).

## Testing

- `ruff check` + `ruff format --check` clean
- `pytest tests/test_multiplexer.py` — 12 passed

Closes #59

### DIFF
--- a/src/bmad_loop/adapters/multiplexer.py
+++ b/src/bmad_loop/adapters/multiplexer.py
@@ -80,7 +80,12 @@ class TerminalMultiplexer(ABC):
         self, session: str, name: str, cwd: Path, env: dict[str, str], command: str
     ) -> str:
         """Create a window running ``command`` (with ``env`` layered on) in
-        ``session``, rooted at ``cwd``. Returns the backend-native window id."""
+        ``session``, rooted at ``cwd``. Returns the backend-native window id.
+
+        ``command`` is a POSIX shlex-joined argv string, not a shell line:
+        shell-operator behavior (``&&``, ``|``, ...) is backend-defined —
+        one backend may hand the string to a shell, another may shlex
+        re-split it into literal argv — so callers must not rely on it."""
 
     @abstractmethod
     def new_parked_window(
--- a/tests/test_multiplexer.py
+++ b/tests/test_multiplexer.py
@@ -9,6 +9,7 @@ role for the transport axis.
 
 import json
 import os
+import shlex
 import subprocess
 
 import pytest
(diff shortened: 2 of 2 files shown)


pbean (approved):
Validated — approving. This is a clean, minimal, docs-only contract pin (+ one pinning test) with **no runtime behavior change**, and it does not risk the core or the other-OS backend path.

**Verification performed:**
- **Docstring is accurate.** `tmux_base.py` `new_window` passes `command` verbatim as the single trailing arg to `tmux new-window` (run via a shell → operators would be interpreted); the eventual re-splitting backend treats them as literal argv. The docstring describes both correctly.
- **Existing callers already comply** with the documented contract — both build `command` as a shlex-quoted join, so no raw shell line ever reaches the seam:
  - `generic.py` `build_command` → `" ".join(shlex.quote(a) …)`
  - `probe.py` `_ProbeLauncher.start` → `" ".join(shlex.quote(a) …)`
  The PR pins the **de-facto** contract; nothing is retrofitted or broken.
- **New test is correct — empirically confirmed.** Reproduced its logic against the real `tmux_base.py`: the asserted argv is byte-identical to `new_window`'s actual output, and the operator-bearing shlex-joined command reaches tmux verbatim as one trailing argument. Symmetric with the existing `test_new_window_posix_argv_byte_identical`.
- **No core / other-OS risk.** Additive documentation on an abstract method plus a passing test. It *reduces* future risk by pinning the contract before the Windows/psmux backend lands — exactly the cross-backend divergence #59 flags.

Nice defensive close on #59. LGTM. 👍

PR #79, opened 2026-07-06
fix(verify): stop treating ledger/spec-only stories as "no changes since baseline"

Pawel-N-pl:
## What

`verify_dev`/`verify_dev_bundle`'s proof-of-work gate now excludes only the session's own spec file and the sprint-status ledger from `has_changes_since`, via a new `verify_dev_exclude_relpaths(paths, spec_path)`, instead of the whole `implementation_artifacts`/`planning_artifacts` folders.

## Why

The old exclusion existed so a session couldn't fake completion by only flipping its own spec's `status:` field, but it hid *everything* under those folders, including sibling content like the deferred-work ledger. A story whose entire authorized scope is ledger/spec reconciliation (no src/tests touched, by design) therefore always false-negatived "no changes since baseline" and got rolled back or deferred, no matter how much real, reviewed work landed.

## How

- New `verify_dev_exclude_relpaths`: excludes `paths.sprint_status` and the claimed `spec_path` only (file, not folder, granularity).
- Wired into both `has_changes_since` call sites, replacing `artifact_relpaths(paths)` (which stays as-is for `Engine`'s rollback-protection use, a different job).
- `output_folder` deliberately stays out of the new exclude set: it's the parent dir of `implementation_artifacts` in the standard layout, so excluding it as a prefix would swallow that folder's content right back out of view.

## Testing

Seven new tests in `tests/test_verify.py`: exclude-helper granularity plus a `..`-segment normalization guard on the helper, a direct `has_changes_since` repro (ledger content counts, own-spec + sprint bookkeeping alone doesn't), both the `verify_dev` and `verify_dev_bundle` gates passing end-to-end on a ledger-only diff, and regression guards that a bare own-spec status flip still retries as "no changes" — both directly and when the spec is claimed via a `..`-laden path. Full suite: 1412 passed, 7 skipped; ruff clean.

## Related

Independent of the other pending PR (`fix/resolve-baseline-advance`); different files, either can land first.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened the “no changes since baseline” check so it only ignores specific bookkeeping files instead of whole artifact folders.
  * Prevents false positives/negatives when validating work that includes ledger updates or status changes.
  * Improved path handling so claimed files are recognized consistently even when referenced through different relative paths.

* **Tests**
  * Added coverage for ledger-only work, file-level exclusions, and status-only changes to verify the new behavior.

### DIFF
--- a/src/bmad_loop/verify.py
+++ b/src/bmad_loop/verify.py
@@ -955,11 +955,19 @@ def verify_dev_exclude_relpaths(paths: ProjectPaths, spec_path: Path) -> tuple[s
     dirs — the deferred-work ledger, other stories' specs — is deliberately left
     un-excluded, so a story whose entire authorized scope is ledger/spec
     reconciliation registers as real work instead of a permanent false "no changes
-    since baseline"."""
+    since baseline".
+
+    `spec_path` comes from a session-reported (untrusted) `spec_file` string, so
+    it is `.resolve()`d before deriving the relpath, same as `spec_within_roots`:
+    an un-normalized `..`/`.` segment would still resolve to the real on-disk
+    file (the OS resolves it), but as a raw string it wouldn't match git's own
+    normalized path output, silently defeating this exclude and letting a bare
+    status flip on the session's own spec count as real work."""
     out: list[str] = []
+    project = paths.project.resolve()
     for path in (paths.sprint_status, spec_path):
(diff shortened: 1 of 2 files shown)


Pawel-N-pl:
@coderabbitai review

PR #81, opened 2026-07-07
fix: bright-yellow SGR crashes TUI log render

### LINKED ISSUE #80: TUI log pane blank: bright-yellow SGR -> 'bright__yellow' ColorParseError
## Description

`bmad-loop tui` crashes its log-render worker with `rich.color.ColorParseError: 'bright__yellow' is not a valid color` (note the **double** underscore). The dashboard log pane then never draws — it just shows blank — for any run whose captured pane output contains a bright‑yellow SGR (SGR 93), which is common in coding‑CLI/agent output.

The bug is in `_rich_color` / `_PYTE_COLOR_FIX` in `src/bmad_loop/tui/data.py`:

```python
# pyte names SGR 33 "brown"; aixterm brights carry no underscore.
_PYTE_COLOR_FIX = {"brown": "yellow", "brightbrown": "bright_yellow"}

def _rich_color(value: str) -> str | None:
    ...
    value = _PYTE_COLOR_FIX.get(value, value)     # "brightbrown" -> "bright_yellow"
    if value.startswith("bright"):
        return "bright_" + value[6:]              # "bright_" + "bright_yellow"[6:] = "bright_" + "_yellow"
    return value
```

pyte names SGR 93 `brightbrown`. The fix‑table remaps it to `bright_yellow`, which **already** carries the `bright_` prefix — but it still matches `startswith("bright")`, so `_rich_color` re‑prefixes it and produces `bright__yellow`, which `rich` rejects.

## Steps to reproduce

```python
from bmad_loop.tui import data
print(data._rich_color("brightbrown"))   # -> 'bright__yellow'
```

Or, live: run `bmad-loop tui` on any run whose agent window emits bright‑yellow text; the `_poll` worker raises and the log pane stays blank. Full traceback tail:

```
File ".../bmad_loop/tui/data.py", line 297, in _char_style
    style = _style_cache[key] = Style(color=_rich_color(fg), ...)
File ".../rich/color.py", line 451, in parse
    raise ColorParseError(f"{original_color!r} is not a valid color")
rich.color.ColorParseError: 'bright__yellow' is not a valid color
```

## Expected behavior

`_rich_color("brightbrown")` returns the valid rich color `bright_yellow`, and the TUI log pane renders bright‑yellow output normally.

## Actual behavior

Returns `bright__yellow`; the render worker raises `ColorParseError` and the dashboard log pane is blank.

## Fix

Keep the remap target in pyte's underscore‑free namespace so the downstream `bright` → `bright_` transform yields a valid color:

```python
_PYTE_COLOR_FIX = {"brown": "yellow", "brightbrown": "brightyellow"}
```

Then `brightbrown` → `brightyellow` → `bright_yellow`. PR with a regression test to follow.

## Environment

- **bmad-loop version:** 0.8.0 (also present in v0.8.1 and on `main` — line unchanged since the `bmad-auto` → `bmad-loop` rename)
- **Coding CLI:** claude
- **OS:** Linux (CachyOS, kernel 7.1)
- **Python:** 3.12

thesleepingsage:
## What

Fixes the TUI log pane rendering blank due to a `ColorParseError` on bright‑yellow output. Closes #80.

## Why

pyte names SGR 93 `brightbrown`. `_PYTE_COLOR_FIX` remapped it to `bright_yellow` — a value that already carries the `bright_` prefix. `_rich_color` then applies its `"bright"` → `"bright_"` transform unconditionally, doubling the underscore into `bright__yellow`, which `rich` rejects:

```python
data._rich_color("brightbrown")   # before: 'bright__yellow'  -> ColorParseError
```

The `_poll` worker raises on the first bright‑yellow line (common in agent output), so `bmad-loop tui`'s dashboard log pane never draws.

## The fix

Keep the remap target in pyte's underscore‑free namespace so the downstream transform produces a valid color:

```python
-_PYTE_COLOR_FIX = {"brown": "yellow", "brightbrown": "bright_yellow"}
+_PYTE_COLOR_FIX = {"brown": "yellow", "brightbrown": "brightyellow"}
```

Now `brightbrown` → `brightyellow` → `bright_yellow`. I also expanded the comment above the table to record the invariant (remap values must stay underscore‑free) so it can't be reintroduced.

## Test

Added `test_rich_color_maps_pyte_names_to_valid_rich_colors` to `tests/test_tui_data.py`: asserts the pyte→rich mapping for the brown/bright‑brown/bright/hex/default cases **and** that every non‑None result parses via `rich.color.Color.parse` — so a future invalid‑color regression fails the test instead of the TUI.

```
$ uv run --all-extras pytest tests/test_tui_data.py -q
59 passed
$ uv run --all-extras ruff check src/bmad_loop/tui/data.py tests/test_tui_data.py
All checks passed!
```

## Scope

One‑line behavior change + comment + one test. No control‑loop logic touched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal UI color handling by remapping pyte “bright” color names to valid Rich display colors.
  * Hardened style rendering so unparseable color values no longer raise errors, falling back gracefully while preserving other text attributes (e.g., bold/underline).
* **Tests**
  * Added tests to confirm pyte-to-Rich color translations always produce parseable values.
  * Added coverage to ensure unparseable colors degrade safely and caching behavior remains consistent.

### DIFF
--- a/src/bmad_loop/tui/data.py
+++ b/src/bmad_loop/tui/data.py
@@ -22,6 +22,7 @@ from pathlib import Path
 from typing import Any
 
 import pyte
+from rich.color import ColorParseError
 from rich.style import Style
 from rich.text import Text
 
@@ -273,7 +274,12 @@ _HISTORY_LINES = 2000  # matches the dashboard RichLog max_lines
 # must stay in pyte's own (underscore-free) namespace: _rich_color re-applies
 # the "bright" -> "bright_" transform after this remap, so a "bright_"-prefixed
 # value would double up into "bright__yellow" (an invalid rich color).
-_PYTE_COLOR_FIX = {"brown": "yellow", "brightbrown": "brightyellow"}
+# "bfightmagenta" is pyte 0.8.2's own BG_AIXTERM[105] typo (SGR 105).
+_PYTE_COLOR_FIX = {
+    "brown": "yellow",
+    "brightbrown": "brightyellow",
+    "bfightmagenta": "brightmagenta",
+}
 _HEX_DIGITS = set("0123456789abcdef")
 
 
@@ -297,15 +303,21 @@ def _char_style(key: tuple) -> Style:
     style = _style_cache.get(key)
     if style is None:
         fg, bg, bold, italics, underscore, strikethrough, reverse = key
-        style = _style_cache[key] = Style(
-            color=_rich_color(fg),
-            bgcolor=_rich_color(bg),
+        attrs = dict(
             bold=bold or None,
             italic=italics or None,
             underline=underscore or None,
             strike=strikethrough or None,
             reverse=reverse or None,
         )
+        try:
+            style = Style(color=_rich_color(fg), bgcolor=_rich_color(bg), **attrs)
+        except ColorParseError:
+            # A color name rich can't parse (e.g. a new pyte table entry) must
+            # degrade to an uncolored run, not kill the poll worker — and with
+            # it the app (textual workers default to exit_on_error=True).
+            style = Style(**attrs)
+        _style_cache[key] = style
     return style
 
 
--- a/src/bmad_loop/tui/data.py
+++ b/src/bmad_loop/tui/data.py
@@ -269,8 +269,11 @@ _PANE_COLUMNS = 220
 _PANE_LINES = 50
 _HISTORY_LINES = 2000  # matches the dashboard RichLog max_lines
 
-# pyte names SGR 33 "brown"; aixterm brights carry no underscore.
-_PYTE_COLOR_FIX = {"brown": "yellow", "brightbrown": "bright_yellow"}
+# pyte names SGR 33 "brown"; aixterm brights carry no underscore. Values here
+# must stay in pyte's own (underscore-free) namespace: _rich_color re-applies
+# the "bright" -> "bright_" transform after this remap, so a "bright_"-prefixed
+# value would double up into "bright__yellow" (an invalid rich color).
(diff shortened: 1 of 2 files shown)


pbean:
Thanks @thesleepingsage — great catch and a clean diagnosis; the regression test was exactly the right shape. Confirmed the repro on `main` (`_rich_color("brightbrown")` → `bright__yellow`) and validated the fix.

While reviewing I swept every name in pyte 0.8.2's color tables through `_rich_color` and found one more crash of the same class just out of this PR's reach: pyte's own `BG_AIXTERM[105]` entry is misspelled **`bfightmagenta`** (an upstream typo in pyte), which passes through unmapped and raises the same `ColorParseError` on any background bright-magenta (SGR 105). Since maintainer edits were enabled I pushed a completion commit (e592059) onto your branch rather than round-tripping a review:

- adds `"bfightmagenta": "brightmagenta"` to `_PYTE_COLOR_FIX`
- extends your test with an exhaustive sweep of pyte's `FG`/`BG`/`FG_AIXTERM`/`BG_AIXTERM` tables, so a future pyte bump that adds or renames entries fails in CI instead of in the dashboard
- guards `_char_style` so an unparseable color degrades to an uncolored run instead of killing the poll worker — textual's `@work` defaults to `exit_on_error=True`, so that exception can take down the whole TUI, slightly worse than the blank pane described in #80

Merging once CI is green.

pbean (approved):
Issue confirmed and reproduced on main; fix validated locally (full suite + an end-to-end LogView render of raw SGR 93/105 bytes). Completed the failure class in e592059 (pyte's bfightmagenta typo + exhaustive table sweep + graceful degrade in _char_style).

PR #93, opened 2026-07-08
feat(sweep): extend intent-gap patch-restore re-drive to deferred-work bundles

### LINKED ISSUE #75: Extend intent-gap patch-restore re-drive to the sweep workflow
## Summary

The intent-gap patch-restore re-drive (added for **story runs**) should be extended to the **sweep workflow** (`SweepEngine` deferred-work bundles).

## Background

BMAD-METHOD [PR #2564](https://github.com/bmadcode/BMAD-METHOD/pull/2564) ("preserve the attempted change on intent-gap halts") makes `bmad-dev-auto` save the reverted attempt as a patch file before halting on an `intent gap` at the review step. The orchestrator now has a sanctioned recovery for **story** escalations: `bmad-loop resolve` (interactive, or `--no-interactive --restore-patch <path>`) re-arms the escalated spec to `in-review`, latches the patch onto the task, and the engine re-applies it onto the baseline after every reset of the re-drive — so the re-driven dev session resumes **review** on the restored diff instead of re-implementing from scratch.

This landed on branch `feat/intent-gap-patch-restore` (plan: `i-received-a-message-unified-journal.md`). Scope was deliberately limited to **story runs**; sweep-bundle escalations still take a from-scratch re-drive.

## What's missing

`SweepEngine` bundle escalations should honor a `restore_patch` latch the same way story runs do:

- re-arm the **bundle spec** to `in-review` (mirroring `runs.rearm_escalation`'s restore branch), rather than the bundle's from-scratch equivalent;
- re-apply the saved patch after the **bundle's** baseline reset, before dispatching the bundle dev session (mirroring `Engine._restore_patch`, called from the dev loop on `feedback is None`);
- clear the latch on the **bundle commit** (mirroring `Engine._commit` clearing `task.restore_patch = None` next to `resolved_redrive = False`).

Note the sweep path has its own `_drive_story` / `_rollback_or_pause` / `_safe_reset` call sites (`sweep.py`), and bundles have no sprint-status entry, so the story-run wiring can't just be inherited unchanged — the seams need porting.

## Design references

- `model.StoryTask.restore_patch` — the persisted latch (already exists; bundle tasks reuse `StoryTask`).
- `verify.apply_patch(repo, patch_path)` — the `git apply` helper (already exists; reusable as-is).
- `Engine._restore_patch` — the apply/journal/escalate-on-failure helper and its `feedback is None` gating in `_dev_phase` (the pattern to mirror in the sweep dev loop).
- `runs.rearm_escalation(..., restore_patch=...)` — the `in-review` vs `ready-for-dev` re-arm branch (the sweep bundle re-arm needs the equivalent).

The sweep resolve/re-arm entry point (however bundle escalations are resolved) would need a `restore_patch` passthrough analogous to `cli.cmd_resolve`.

pbean:
Closes #75.

Extends the intent-gap patch-restore re-drive (built for sprint/story runs in #86) to `SweepEngine` deferred-work bundles. An escalated bundle can now be resolved with `bmad-loop resolve <run> --restore-patch <path>`, which re-arms the bundle spec to `in-review` and re-applies the saved patch on the re-drive, so the corrected session resumes review on the restored diff instead of reimplementing from scratch.

## Why this was small

Investigation showed almost the whole mechanism was already inherited or pre-wired for sweep — the apply seam (`_restore_patch` inside the inherited `_dev_phase`), latch-clear on commit (`_commit`), the T4 proof-of-work exclusion (`verify_dev_bundle`), `runs.rearm_escalation` (run_type-agnostic), and the CLI `resolve`→`resume` routing (`_resume_paused_run` builds `SweepEngine` for `run_type=="sweep"`). The gap was that `SweepEngine` overrides `_loop` wholesale and hand-rolls recovery in `_run_bundle`, so it never reached the base `_finish_inflight` re-drive wiring.

## The two production changes (`src/bmad_loop/sweep.py`)

- **`_generic_bundle_prompt`** — a restore re-drive now points at the in-review bundle spec (`task.spec_file`) instead of the fresh `intent.md`, so upstream step-01's spec-pointer intent check early-exits *before* its version-control sanity check would HALT `blocked` on the restored dirty tree. Mirrors the sprint F1 fix (`engine.py:2133-2141`).
- **`_run_bundle` recovery** — latches `resolved_redrive` and passes `cause="resolved"` for a re-armed task, mirroring `_finish_inflight` (`engine.py:1151-1157`). This is a **correctness fix**: without the latch, a non-critical exhaustion mid-re-drive silently *DEFERs* the human-resolved escalation (`escalation.py:110` → `PAUSE if resolved_redrive else DEFER`) — filing it as fresh deferred work and rolling back the correction. It also repairs the pre-existing **from-scratch** sweep re-drive, which had the same latent bug.

## Tests

- **Engine (`tests/test_sweep.py`)** — restore-prompt unit; restore re-drive → done + both latches cleared + prompt-points-at-spec + `attempt-restored`; pause-not-defer exhaustion for both the restore and from-scratch paths. New escalating-dev helper in `conftest.py`.
- **E2E scenario 8 (`tests/test_stories_e2e.py`)** — drives the real `sweep`/`resolve --restore-patch`/`resume` binaries through tmux (a triage branch added to the fake CLI; the bundle-dev session reuses the existing sprint branch). This is the only coverage of the sweep-specific CLI resume path.
- Every new test verified to **fail without the fix**. Full suite **1695 passed / 1 skipped** (Windows-only); `trunk check` clean.

## Residual (out of scope)

If `triage.json` fails to reload on resume, a re-armed bundle can be orphaned by a fresh triage plan whose bundle names differ — pre-existing sweep fragility, not introduced here. Worth a follow-up issue.

## Summary by CodeRabbit

* **New Features**
  * Added clearer recovery guidance for interrupted review flows, including restoring from a saved patch and continuing review instead of restarting work.
  * Extended the same recovery flow to deferred-work sweep runs, including resumable re-drive behavior after escalation.

* **Bug Fixes**
  * Improved handling of restored runs so recovery now resumes in the correct state and re-applies the saved patch consistently.
  * Refined pause vs. defer behavior for exhausted restore/re-drive scenarios.


pbean:
Filed the out-of-scope residual (triage.json reload failure orphaning a re-armed bundle) as #94.

pbean:
@coderabbit review
augment review

Issue #61, opened 2026-07-03
Hardening: reconcile seam is unreachable for stalled sessions — a finished-but-lost-Stop session cannot be recovered as completed


pbean:
### Description

Follow-up surfaced while validating #53 (fixes #48). With Stop/window-death as the only authoritative completion signals, one recovery scenario is left with no automated path:

A dev/review session that **genuinely finished its work** (terminal artifact written, frontmatter `done`) but whose final Stop event was lost, and which then fails to answer any stall nudge, now ends `status="stalled"` with `result_json=None` (`_final(accept_result=False)`, `adapters/generic.py`). That verdict is the right call at the adapter layer — an artifact under a live window that ignores nudges is untrustworthy by construction (#48). The gap is downstream:

- `_reconcile_generic_terminal_status` (`engine.py` ~1576) is called from exactly one site, inside `if result.status == "completed":` in `_dev_phase` (~1271). It is **structurally unreachable** for `stalled`/`crashed`/`timeout`.
- `decide_dev` (`escalation.py` ~71) treats every non-completed status identically: RETRY while budget remains, else DEFER/PAUSE. On RETRY the verify `outcome` is always `None` (the whole verify block is completed-gated), so the engine goes straight to `_rollback_or_pause`.

Net effect: finished-but-unvouched work is never inspected before disposal. Current mitigations are real but passive — `scm.rollback_on_failure` defaults OFF (pause for manual recovery, tree untouched), and when it is ON, `_preserve_attempt_commits`/`_preserve_attempt_worktree` park the work under `attempt-preserve/*` refs. But nothing ever re-reads the on-disk state to notice the session actually succeeded.

### Likelihood

Low: it requires a lost/undelivered Stop **and** total nudge-unresponsiveness (default 2 nudges × `dev_stall_grace_s`). A responsive agent answers a nudge, ends its turn, and the fresh Stop completes normally. But hook-delivery failure modes (hook misconfiguration, events.jsonl write failure, hard-wedged CLI after finishing) make it a plausible long-tail on flaky environments — precisely where the native-Windows backend will live.

### Proposed direction

A guarded post-kill reconcile for the `stalled` verdict — trust is re-established by window death rather than prose:

1. Engine kills the session as today (`run()`'s `finally`).
2. For a `stalled` result only, after the kill (window now provably dead), re-run the artifact scan (`find_result_artifact` + `synthesize_result`) against the session's launch floor.
[...]

PR #98, opened 2026-07-09
fix(platform): harden win32 atomic replaces and path segments

### LINKED ISSUE #74: Windows: engine silently dies ("engine gone - run was interrupted") - save_state os.replace() WinError 5 sharing violation with TUI reader
## Summary

On Windows a long-running engine can **die silently** mid-run. The TUI then shows **"engine gone — run was interrupted — press e to resume"**, the run is frozen at `phase=dev-running`, and `state.json` has `crashed=false` / `stopped=false` (no crash recorded). Root cause: `save_state()`'s atomic write `os.replace(state.json.tmp → state.json)` raises `PermissionError: [WinError 5] Access is denied` because the **TUI poller holds an open read handle on `state.json`** while the engine tries to rename over it.

This is a Windows-only defect — it does not reproduce on Linux/WSL/macOS.

## Environment

- Windows 11, Python 3.x, native-Windows multiplexer + `WindowsProcessHost`
- Reproduced against a live `bmad-dev-auto` run

## Root cause

`src/bmad_loop/journal.py`:

```python
def save_state(run_dir: Path, state: RunState) -> None:
    ...
    tmp.write_text(json.dumps(state.to_dict(), indent=2), encoding="utf-8")
    os.replace(tmp, target)          # <-- WinError 5 here
```

`load_state()` reads `state.json` via `Path.read_text()`, i.e. `open()` **without `FILE_SHARE_DELETE`**. On Windows `MoveFileExW` (what `os.replace` calls) fails with `ERROR_ACCESS_DENIED` when another handle on the target is open. The TUI dashboard polls `state.json` on every tick via `load_state()`, so writer and reader collide stochastically.

On POSIX `rename(2)` over an open fd always succeeds (the reader keeps the old inode), which is why this only bites on Windows.

## How it was found

1. A frozen run showed a dead engine pid but `state.json` had just been written (mtime matched the last `Stop` event) — yet `journal.jsonl` had **no `session-end` line**, and `crashed=false`.
2. That pinpointed the death to the window in `_run_session` between the durability `self._save()` (`engine.py:~1994`) and the `journal.append("session-end")` a few lines later — the code comment there *already anticipates a host kill in that window*.
3. First occurrence was silent (uncaught); on **resume** the same collision surfaced as a real crash:

   ```
   PermissionError: [WinError 5] Access is denied:
   '...\runs\<id>\state.json.tmp' -> '...\runs\<id>\state.json'
   ```

   which confirmed the `os.replace` sharing violation as the mechanism.

## Broader hazard class

This is one instance of "POSIX-safe idiom, Windows-unsafe". A short research pass surfaced a ranked checklist we intend to work through (each with a grep detection heuristic + fix pattern):

- **H1 — atomic replace over a concurrently-read file** (this issue). *pip wraps its wheel `replace()` in retries; npm graceful-fs backs off up to 60s for EACCES/EPERM/EBUSY.*
- **H2 — `os.kill(pid, 0)` is a destructive `CTRL_C_EVENT` on Windows** (it maps to `GenerateConsoleCtrlEvent`, not an existence probe). Already mitigated by `WindowsProcessHost` (psutil); worth a guard against stray callers.
- **H3 — console close/logoff/shutdown events are not deliverable to Python signal handlers** → silent engine death; only resumable checkpoints save you.
- **H4 — `taskkill /F /T` tree blast radius + Windows PID reuse** (guard force-kills with a create-time identity match).
- **H5 — `CREATE_NEW_PROCESS_GROUP` / `CTRL_C_EVENT` cannot target a single process** (hits the whole console).
- **H6 — AV/Defender/Search-indexer transient handle holds** widen the H1 window beyond just the TUI reader.
- **H7 — long paths (>260), UTF-8 default encoding, CRLF, reserved filenames.**

## Proposed fix

Route `save_state` through a retry helper — retry `os.replace` on `PermissionError` with short jittered backoff, gated to `sys.platform == "win32"` so a genuine POSIX `EACCES/EPERM` still surfaces immediately:

```python
def _atomic_replace(tmp: Path, target: Path) -> None:
    for attempt in range(_REPLACE_RETRIES):
        try:
            os.replace(tmp, target)
            return
        except PermissionError:
            if sys.platform != "win32" or attempt == _REPLACE_RETRIES - 1:
                raise
            time.sleep(_REPLACE_BACKOFF_S * (attempt + 1))
```

Follow-ups: factor this into one shared `atomic_replace` for **every** `os.replace`/rename-over site (audit needed), widen the backoff to survive AV/indexer holds (H6), and add guards/tests for H2/H4/H5.

## Plan / offer

I have this implemented and unit-tested locally and am happy to open a PR (the `save_state` retry first, then the broader audit as separate PRs). Self-assigning to track it.

Repro-once check: keep the TUI open on a running story on Windows; the collision is timing-dependent but recurs within a few state writes under active polling.

dracic:
## Problem

`os.replace` over a file another process holds open fails on Windows with WinError 5 (Python readers take no `FILE_SHARE_DELETE`), which froze a live run when the TUI poller collided with the engine's state write. POSIX rename-over-open never raises, so none of this bites on Linux/macOS. The same audit surfaced two more Windows defect classes with no cover: no filename sanitizer for user-derived path segments, and an unguarded force-kill.

## Changes

- **`atomic_replace`**: win32-gated, jittered-backoff retry around `os.replace` (~5 s worst case, sized for AV/indexer handle holds); every rename-over-open site routes through it — state, decisions, sweep answers, TUI settings, run archives.
- **`safe_segment`**: Windows-filename sanitizer (illegal/control chars, reserved device names incl. `COM0`/superscripts/`CONIN$`/`CONOUT$`, trailing dot/space, length cap, digest suffix on any change for practical collision resistance; identity for clean input) applied wherever story/unit keys become path or window segments: task ids, feedback files, resolve dirs, worktree and failed dirs, deferred artifacts. Composed task ids (key + plugin label) are sanitized as one segment so two individually capped parts can't exceed the filename limit.
- Agent MCP configs read BOM-tolerantly, failing closed on undecodable content; unity teardown explains identity-guard skips.
- `os.kill` usage gated outside the process host (test tripwire).

## Testing

- New unit tests: retry/backoff matrix (incl. POSIX zero-retry), sanitizer contract (reserved names, collision suffix, length cap), composed task-id cap, dirty-key resolve dir, save_state regression under a transient `PermissionError`.
- Full suite on Windows: 1707 passed; only the pre-existing skill-sync baseline failures.
- Lint verified locally at CI-pinned versions (ruff, black 26.5.1, isort 8.0.1, bandit 1.9.4 with repo config).

Closes #74

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Unity config parsing when JSON files include a UTF-8 BOM.
  * Hardened shutdown and run persistence to avoid issues from reused process IDs and improve reliability of atomic file writes (with Windows retry behavior).
  * Sanitized story/unit task and directory naming to prevent failures from invalid or reserved filename characters.

* **Tests**
  * Added/expanded coverage for BOM parsing, atomic replacement retries, path sanitization/collision handling, safer teardown behavior, and the portability guard for restricted `os.kill` usage.

### DIFF
--- a/src/bmad_loop/data/plugins/unity/unity_setup.py
+++ b/src/bmad_loop/data/plugins/unity/unity_setup.py
@@ -304,7 +304,7 @@ def _local_url(worktree: Path) -> str | None:
         return override.strip()
     cfg = worktree / ".mcp.json"
     try:
-        data = json.loads(cfg.read_text())
+        data = json.loads(cfg.read_text(encoding="utf-8-sig"))
         servers = data.get("mcpServers", {})
         entry = servers.get(_MCP_SERVER_NAME) or next(iter(servers.values()), {})
         url = entry.get("url")
@@ -425,10 +425,10 @@ def _verify_agent_isolation(agent: str, worktree: Path, url: str) -> bool:
     port = url.rsplit(":", 1)[-1].strip("/")
     cfg = worktree / rel
     try:
-        text = cfg.read_text()
-    except OSError:
+        text = cfg.read_text(encoding="utf-8-sig")
+    except (OSError, UnicodeDecodeError):
         print(
-            f"unity_setup: agent {agent!r} MCP config {rel} missing after setup; "
+            f"unity_setup: agent {agent!r} MCP config {rel} missing or unreadable after setup; "
             "cannot guarantee per-worktree isolation",
             file=sys.stderr,
         )
--- a/src/bmad_loop/data/plugins/unity/unity_teardown.py
+++ b/src/bmad_loop/data/plugins/unity/unity_teardown.py
@@ -176,11 +176,20 @@ def _force_kill_lingering(worktree: Path) -> int:
             break
         time.sleep(0.5)
     for pid in pids:
-        if host.is_alive(pid) and ident[pid] is not None and host.identity(pid) == ident[pid]:
-            try:
-                host.force_kill(pid)
-            except OSError:
-                pass
+        if not host.is_alive(pid):
+            continue
+        if ident[pid] is None or host.identity(pid) != ident[pid]:
+            print(
+                f"unity_teardown: pid {pid} survived terminate but its identity is "
+                "missing or changed since teardown began; skipping force-kill to "
+                "avoid hitting a reused pid",
+                file=sys.stderr,
+            )
+            continue
+        try:
+            host.force_kill(pid)
+        except OSError:
+            pass
     return len(pids)
 
 
--- a/src/bmad_loop/decisions.py
+++ b/src/bmad_loop/decisions.py
@@ -27,6 +27,7 @@ import re
 from pathlib import Path
 
 from . import bmadconfig, deferredwork, runs, verify
+from .platform_util import atomic_replace
 from .sweep import Decision, DecisionOption, validate_triage
 
 STORE_REL = Path(".bmad-loop") / "decisions.json"
@@ -58,7 +59,7 @@ def _write_store(project: Path, data: dict) -> None:
     path.parent.mkdir(parents=True, exist_ok=True)
     tmp = path.with_suffix(".tmp")
     tmp.write_text(json.dumps(data, indent=2, sort_keys=True), encoding="utf-8")
-    tmp.replace(path)
+    atomic_replace(tmp, path)
 
 
 def record_pre_answer(project: Path, dw_id: str, option: DecisionOption, *, date: str) -> None:
--- a/src/bmad_loop/engine.py
+++ b/src/bmad_loop/engine.py
@@ -39,6 +39,7 @@ from .model import (
     SessionRecord,
     StoryTask,
 )
+from .platform_util import safe_segment
 from .plugins import HookBus, HookContext, PluginRegistry
 from .policy import Policy
 from .runs import kill_session
@@ -1175,7 +1176,7 @@ class Engine:
             role, seq = "review", task.review_cycle
         else:
             return None
-        task_id = f"{task.story_key}-{role}-{seq}"
+        task_id = f"{safe_segment(task.story_key)}-{role}-{seq}"
         for record in reversed(task.sessions):
             if record.task_id != task_id:
                 continue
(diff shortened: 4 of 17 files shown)


dracic:
@pbean please see the outside diff comment...

pbean:
@dracic Confirmed — the outside-diff critical is real, and it's a good catch by the rabbit. I traced the consequence: for any story key that `safe_segment` actually changes (dirty chars, or a clean key whose composed id overflows the 120-char cap), `_resumable_session`'s part-only composition never matches the stored `SessionRecord.task_id`, so `_finish_inflight` falls through to the destructive resume-restart (worktree discard / baseline rollback) — silently defeating the #61 post-kill rescue for exactly the keys this PR hardens.

Rather than round-trip, I pushed the fix to your branch (c768e9d): both sites now route through one `_session_task_id` helper — `SessionRecord` stores no `seq`, so exact-id matching is load-bearing, and a shared composition point beats mirroring the formula (composing inline is how the two diverged). Plus a parity regression test (dirty key + cap-overflowing clean key) that's red on your previous head and green now. Full suite + trunk clean locally.

On the rest of the review sweep:

- **CodeRabbit's `noqa: S311` nitpick is moot** — ruff's `S` family isn't enabled in this repo (no `[tool.ruff.lint] select`); bandit runs separately via trunk, so your `# nosec B311` is exactly right. No change needed.
- **Non-blocking, take or leave:** `resolve.py`'s `f"{safe_segment(story_key)}-resolve-1"` uses the part-only style. It's harmless today (write-only id, no sibling composition to disagree with), but if you're touching the file anyway, `safe_segment(f"{story_key}-resolve-1")` keeps the whole-composition principle uniform.
- **Non-blocking:** CodeRabbit's suggested BOM/undecodable-config tests for `_verify_agent_isolation` would be cheap coverage for the two paths you added; fine as a follow-up too.

I also ran a full consumer audit of every `safe_segment` site: everything else is consistent (readers use stored values — `task.worktree_path`, `SessionRecord.task_id`, returned `Path`s — funnel through a single chokepoint like `resolve._story_dir`, or match by prefix+mtime like the marker discovery), and `atomic_replace` covers both TUI-polled rename targets (state.json, decisions store). Two adjacent pre-existing gaps the audit surfaced are now #101 (`_stash_deferred_artifacts` `shutil.move` → `FileExistsError` on Windows re-defer) and #102 (`unit_branch_name` passes the raw key to git as a ref) — neither blocks this PR.

Will approve once CI is green on the new head. Thanks for a genuinely thorough piece of work — the win32-gated retry sizing rationale and the reserved-name coverage (COM0/superscripts/CONIN$) are better than what most of the ecosystem ships.

pbean (approved):
Validated end to end: full consumer audit of every `safe_segment` site and every rename-over site came back clean (details in my comment above), the one critical (resume task-id divergence) is fixed on-branch with a red/green parity test, and CI is green on the new head including both Windows jobs. Follow-ups tracked in #101/#102.

PR #103, opened 2026-07-09
fix(platform): overwrite-safe deferred stash + ref-safe unit branch names

### LINKED ISSUE #101: Windows: _stash_deferred_artifacts shutil.move raises FileExistsError when a story re-defers
## Summary

`_stash_deferred_artifacts` (engine.py, ~line 2338) moves a deferred story's spec into `run_dir/deferred/<story_key>/<spec_name>` via `shutil.move`. On the same filesystem that falls back to `os.rename`, which on Windows raises `FileExistsError` when the target already exists — i.e. when the **same story defers a second time** with the same spec filename. POSIX `rename(2)` silently overwrites, which is why this has never bitten on Linux/macOS.

This is another instance of the "POSIX-safe idiom, Windows-unsafe" class from #74 (H-list), surfaced while auditing PR #98 — it's the one rename-over site in the package that #98's `atomic_replace` sweep did not cover (the target is a possibly-existing *file*, not a fresh temp-rename).

## Fix direction

Give the move overwrite semantics and the win32 retry in one shot: copy/write into a temp path inside the dest dir, then route through `platform_util.atomic_replace` (which wraps `os.replace` — overwriting — with the win32 sharing-violation retry from #98). Alternatively `os.replace` directly if the source is same-filesystem-guaranteed.

## Notes

- The stash is a write-only forensic artifact (no reader recomposes the path), so the blast radius is the defer flow crashing mid-run on Windows, not data corruption.
- The two stdlib-only hook scripts' `os.replace` calls (`data/bmad_loop_hook.py`, `data/bmad_loop_probe_hook.py`) were also checked and are fine as-is: time_ns-unique target names, never concurrently read, and they can't import the package by design.

Refs: #74, #98

### LINKED ISSUE #102: Dirty unit keys still reach git as raw ref names (unit_branch_name) after #98's dir sanitization
## Summary

PR #98 sanitizes the unit **worktree directory** (`workspace.open_unit_workspace`: `unit_worktrees_dir(run_dir) / safe_segment(unit_key)`) and the **failed dir**, but `unit_branch_name` (workspace.py:62-67) still builds the git branch `bmad-loop/{run_id}/{unit_key}` from the **raw** key. A unit key containing git-ref-illegal sequences (`:`, `?`, `[`, `\\`, space, `..`, `@{`, a trailing `.lock`, leading/trailing `/`…) now gets past the filesystem layer only to fail at `branch_exists`/`worktree_add` with an invalid-ref error.

Not a reader-mismatch bug: the branch is stored on `task.branch` at open time and every consumer (`_merge_local`, `discard_worktree`, `close_unit_workspace`) uses the stored value — the failure is purely at creation time. Pre-existing behavior, but after #98 the dir/branch pair is inconsistently hardened, which can read as a false sense of safety.

## Fix direction

Sanitize the branch segment too. Note git ref rules ≠ Windows filename rules, so `safe_segment` alone is not sufficient (e.g. `..`, `@{`, trailing `.lock` are ref-illegal but filename-legal): either a dedicated ref-segment sanitizer (same identity-for-clean-input + digest-suffix contract as `safe_segment`), or `safe_segment` plus a `git check-ref-format --branch` validation pass with the digest fallback.

Surfaced by the consumer audit of PR #98.

Refs: #74, #98

pbean:
Closes #101, Closes #102

Both issues came out of the **consumer audit of PR #98** — the win32 hardening pass that added `platform_util.atomic_replace` and `safe_segment`. That audit asked a narrow question of the merged code: for every site #98 hardened, is every *consumer* of that site hardened too? Two gaps fell out. #98 swept the package's rename-over sites but skipped the one whose target is a possibly-existing *file* rather than a fresh temp-rename (the deferred-artifact stash), and it sanitized the unit **worktree directory** while leaving the unit **branch** built from the raw key — hardening the dir/branch pair inconsistently, which reads as a false sense of safety. This PR closes both.

## #101 — deferred stash

`_stash_deferred_artifacts` moved a re-deferred story's spec straight over the prior stash with `shutil.move`. Now it stages a `copy2` inside the destination dir and routes through `atomic_replace`.

**The issue's stated mechanism is wrong and the PR does not reproduce it.** #101 claims `shutil.move` "raises `FileExistsError`" on Windows when the target exists. It does not: `move` wraps `os.rename` in a bare `except OSError` and falls back to `copy2` + `unlink`, which overwrites. Verified against the CPython source and by emulating Windows rename semantics. Two *real* hazards ride on that fallback instead, and they are what this change fixes:

1. **Sharing violation.** An AV/indexer handle on the target makes `os.rename` fail with WinError 5/32; `move` falls back to `copy2`, which opens the same locked target and fails too, propagating `PermissionError` and crashing the defer flow. `atomic_replace` retries the replace until the handle clears.
2. **Non-atomicity.** The `copy2` fallback truncates the live target before writing, so a crash mid-copy leaves a torn stash. Stage-then-replace never exposes a partial file.

Both halves of the move are retried, not just the replace (added in review): Windows denies a *delete* against an open handle exactly as it denies a rename-over, so an unretried `spec_path.unlink()` would fail after the stash had already landed — and `_defer` calls this before the rollback, the `story-deferred` journal append, and `_save()`, so that raise aborts the entire deferral. `atomic_replace`'s backoff is extracted into `_retry_on_sharing_violation` and reused by a new `platform_util.retrying_unlink`. Suppressing the unlink instead would silently leave the stale in-review spec this function exists to remove.

Test consequences worth flagging for review: the plan's suggested "pre-create the dest file → stash replaces it" test is a **characterization** test on every platform, not a regression guard — old code passes it too. The two tests that actually fail on `shutil.move` are `test_stash_deferred_artifacts_survives_a_win32_sharing_violation` (simulated WinError 32, asserts the retry lands) and `..._keeps_source_and_cleans_tmp_on_replace_failure` (source survives a failed replace; no `.tmp` residue). Both were confirmed red against the pre-change body.

## #102 — ref-safe branch names

New `platform_util.safe_ref_segment`, applied to **both** segments of `unit_branch_name`. Same contract as `safe_segment` — identity for clean input, `-<hex8>` digest suffix whenever anything changed, never raises, pure Python (git can only validate, not sanitize) — but on git's alphabet, which is disjoint from Windows' **in both directions**: `CON` is a legal ref and an illegal filename; `a..b` is the reverse. A consumer deriving both a directory and a branch from one key must therefore run both sanitizers, and the module docstring says so.

Sanitizes control chars/space/DEL and ``~^:?*[\/`` → `_`, `..` → `__`, `@{` → `_{`, leading `.` → `_`, lone `@` → `_`, caps length. Trailing `.` and trailing `.lock` are ref-illegal but get **no rewrite** — flagging them un-clean routes them to the digest path, and the `-<hex8>` suffix is itself the fix. A leading `-` is deliberately preserved: it is legal *in a component*, and the porcelain's separate "must not start with `-`" check reads the whole name, which callers always prefix.

Unlike #101, this issue's mechanism reproduces exactly as written — reverting `unit_branch_name` fails the new test with `fatal: 'bmad-loop/test-run/story/1:2..3@{now}.lock' is not a valid branch name`.

**Oracle test.** `git check-ref-format --branch` is run over a 42-entry corpus (clean keys, one entry per coercion rule, adversarial combinations like `story/1:2..3@{now}.lock`, `....`, `..lock`, `a.lock.lock`, `@{u}`, 500 chars), in each of the three positions `unit_branch_name` places a segment — `bmad-loop/rid/{}`, `bmad-loop/{}/1-1-a`, `bmad-loop/{}`. Pure-Python sanitization is only as good as its agreement with git; this pins it. Plus an end-to-end test where **real git** mounts a worktree for a dirty unit key.

The clean-key branch-name tests (`test_branch_per_story_naming`, `test_branch_per_run_naming`) are unchanged and still assert byte-identical names.

## Verification

- `.venv/bin/python -m pytest` — 1922 passed, 1 skipped.
- Full `trunk check` (no filter) — clean.
- New tests confirmed **red** against the pre-change implementations for both fixes.
- #101's target platform is Windows; the windows CI leg is the one that matters here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Worktree and branch creation now handle run IDs and unit keys with special characters more reliably, preventing invalid branch-name failures.
  * Deferred artifact storage is now safer and more resilient, reducing the risk of corrupted or partial copies during interruptions and on Windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,8 +42,9 @@ breaking changes may land in a minor release.
   re-stashes the same spec filename over the previous one. `shutil.move` fell back to a non-atomic
   `copy2` there — which tears the stash on a mid-copy crash and fails outright on Windows when an
   AV/indexer handle turns the rename into a sharing violation. The stash now stages a copy inside the
-  destination dir and routes through `platform_util.atomic_replace`, inheriting its win32 retry.
-  (closes #101)
+  destination dir and routes through `platform_util.atomic_replace`, inheriting its win32 retry; the
+  source removal gets the same retry via a new `platform_util.retrying_unlink`, since Windows denies a
+  delete against an open handle exactly as it denies a rename-over. (closes #101)
 - **A finished session whose final `Stop` hook was lost no longer loses its work.** A dev/review
   session that wrote its terminal spec but never delivered the `Stop` ended `stalled` — or `timeout`,
   when hooks were misconfigured and no event ever arrived — and the on-disk result was discarded.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,19 @@ breaking changes may land in a minor release.
 
 ### Fixed
 
+- **Unit keys with git-ref-illegal characters no longer break worktree runs.** `unit_branch_name`
+  built `bmad-loop/<run_id>/<unit_key>` from the raw ids, so a key or `--run-id` carrying `:`, `..`,
+  `@{`, a space or a trailing `.lock` cleared the (already-sanitized) worktree dir only to die at
+  `git worktree add` with _"is not a valid branch name"_. Both segments now go through a new
+  `platform_util.safe_ref_segment` — identity for clean ids, `-<hex8>` digest suffix otherwise, on
+  git's alphabet rather than Windows' (`CON` is a legal ref; `a..b` is a legal filename). A
+  `git check-ref-format` oracle test pins the agreement. (closes #102)
+- **The deferred-artifact stash overwrites its target atomically.** A story deferring a second time
+  re-stashes the same spec filename over the previous one. `shutil.move` fell back to a non-atomic
+  `copy2` there — which tears the stash on a mid-copy crash and fails outright on Windows when an
+  AV/indexer handle turns the rename into a sharing violation. The stash now stages a copy inside the
+  destination dir and routes through `platform_util.atomic_replace`, inheriting its win32 retry.
+  (closes #101)
 - **A finished session whose final `Stop` hook was lost no longer loses its work.** A dev/review
   session that wrote its terminal spec but never delivered the `Stop` ended `stalled` — or `timeout`,
   when hooks were misconfigured and no event ever arrived — and the on-disk result was discarded.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,8 @@ breaking changes may land in a minor release.
   `git worktree add` with _"is not a valid branch name"_. Both segments now go through a new
   `platform_util.safe_ref_segment` — identity for clean ids, `-<hex8>` digest suffix otherwise, on
   git's alphabet rather than Windows' (`CON` is a legal ref; `a..b` is a legal filename). A
-  `git check-ref-format` oracle test pins the agreement. (closes #102)
+  `git check-ref-format` oracle test pins the agreement; the `attempt-preserve` recovery-ref slugs
+  now reuse the same sanitizer instead of their own inline one. (closes #102)
 - **The deferred-artifact stash overwrites its target atomically.** A story deferring a second time
   re-stashes the same spec filename over the previous one. `shutil.move` fell back to a non-atomic
   `copy2` there — which tears the stash on a mid-copy crash and fails outright on Windows when an
--- a/src/bmad_loop/engine.py
+++ b/src/bmad_loop/engine.py
@@ -39,7 +39,7 @@ from .model import (
     SessionRecord,
     StoryTask,
 )
-from .platform_util import safe_segment
+from .platform_util import atomic_replace, safe_segment
 from .plugins import HookBus, HookContext, PluginRegistry
 from .policy import Policy
 from .runs import kill_session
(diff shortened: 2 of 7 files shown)


pbean:
@coderabbitai Addressed in f2b965a, though with a different remedy than suggested — worth writing down why.

The finding lands: after `atomic_replace` succeeds, `spec_path.unlink()` was unretried, and Windows denies a *delete* against an open handle exactly as it denies a rename-over. So the staged move was half-hardened. It matters more than "the run fails" suggests: `_defer` calls `_stash_deferred_artifacts` **before** `_rollback_or_pause`, before the `story-deferred` journal append, and before `_save()` (engine.py:2313 vs :2318-2334) — a raise there aborts the entire deferral, not just the stash.

One correction to the framing: this is not a regression the PR introduces. `shutil.move`'s own Windows path is `copy2(src, dst)` + `os.unlink(src)`, with that unlink equally unguarded, so old code could already land the copy and then die on the delete. The PR neither created nor fixed that — it just made the omission incoherent, since the function now advertises win32 hardening.

I did **not** take the suggested fix (suppress / log the unlink failure). Silently leaving the source spec in place recreates precisely the condition this function exists to prevent — "a leftover in-review spec would confuse the next attempt" — and it would do so quietly, on Windows only, at the exact moment a human is already being asked to look at a deferral. Trading a loud crash for a silent stale artifact is a bad trade here.

Instead: `atomic_replace`'s win32 backoff is extracted into `_retry_on_sharing_violation` and reused by a new `platform_util.retrying_unlink`, so both halves of the move ride out the same transient lock. A *permanent* failure still propagates, preserving today's semantics. Replace-then-unlink ordering is now documented as deliberate — a failure must be able to leave a duplicate spec, never a hole where the work was.

Covered by `test_stash_deferred_artifacts_retries_a_locked_source_spec` (WinError 32 on the source delete, retried, stash intact) plus `retrying_unlink` unit tests for the retry, the POSIX no-retry path, and non-sharing-violation errors propagating unswallowed. All confirmed red against a plain `spec_path.unlink()`.

Issue #91, opened 2026-07-08
Patch-restore cleanup batch: centralize latch preconditions, one path normalizer, test-fixture dedup


pbean:
Non-bug cleanups from the PR #86 final review, batched:

- **Centralize the restore-latch preconditions in `runs.rearm_escalation`** (single seam): the worktree/containment checks live in `cli._resolve_restore_patch` while the sentinel + spec-less rejections live in `rearm_escalation` — any future caller of `rearm_escalation` (TUI restore parity, programmatic) bypasses the CLI half. The CLI's pre-session fail-fast can call the same shared validator.
- **One restore-path normalizer**: the maybe-relative→absolute normalization of `task.restore_patch` exists in three copies resolving against two different bases (`engine._restore_patch` → `workspace.root`; `verify.verify_dev_exclude_relpaths` → `paths.project`; `cli._resolve_restore_patch` → `project`). The CLI always latches an absolute resolved path today, so the relative branches are dead — either delete them and document the latch as absolute-only (model.py currently documents 'repo-relative-or-absolute'), or extract one named helper.
- **Vacuous phase guard in `engine._restore_patch`**: at its sole call site the task is always DEV_RUNNING, so `if task.phase == Phase.DEV_RUNNING` is always true — and the comment's PENDING case wouldn't survive `_escalate` anyway (PENDING→ESCALATED is not a legal transition). Replace with an unconditional advance + call-site-invariant comment.
- **T4 altitude**: `verify_dev_exclude_relpaths`'s `restore_patch` param is threaded through three call sites above `_verify_shared_gates` — the anti-drift seam that already receives `task`. Deriving the exclusion inside the shared gate removes the default-None foot-gun for a future fourth mode.
- **Test dedup**: third `_escalated_run` fixture copy (test_runs / test_resolve / test_cli — promote a parameterized builder to conftest); E2E scaffolds hand-roll the skill list three times (loop over `install.DEV_BASE_SKILLS` like `tests/test_install.py` does); `tests/test_engine.py::_restoring_dev_effect` re-builds the result_json envelope that conftest's `dev_effect` factory owns.

PR #105, opened 2026-07-09
fix(verify): baseline-untracked subtraction + skill's baseline key in the shared gate

### LINKED ISSUE #88: Proof-of-work gate: untracked halt residue can vacuously satisfy 'changes since baseline' (from-scratch re-drives)
## Context

Found by the PR #86 final review (adversarially verified). Pre-existing on `main` since #79's file-granular excludes + upstream BMAD-METHOD#2564 (which saves intent-gap patches regardless of orchestrator version); #86 narrows it by excluding the **latched** patch (T4) but cannot close the from-scratch case.

## The gap

`verify.has_changes_since` (verify.py:141-155) counts **every** untracked file and never subtracts the `baseline_untracked` snapshot (contrast `attempt_dirty` directly below it, which does). The saved intent-gap patch is untracked residue under `implementation_artifacts`, which `_protected_relpaths` shields from every reset.

So after ANY intent-gap halt, a **from-scratch** re-arm (`restore_patch=None` — the common "attempted reading was wrong" case) leaves the stale patch on disk, un-excluded (the orchestrator never learns its path: it only exists in escalation-detail free text). A re-driven session that produces zero code changes but flips the spec to the expected status passes the proof-of-work gate on the patch file's mere presence, and `finalize_commit`'s `add -A` sweeps it into the story commit.

## Why path exclusion can't fix it

`verify_dev_exclude_relpaths(…, restore_patch=…)` (the T4 fix) keys on the latch; on a from-scratch re-arm there is no latch and no reliable path. The only mechanical close is snapshot subtraction in the gate itself.

## Fix shape

Subtract `task.baseline_untracked` from the untracked set in the proof-of-work gate — with a **None fallback that keeps counting all untracked files** (pre-snapshot runs must not weaken the gate, unlike `attempt_dirty`'s ignore-all semantics). Both the re-arm snapshot (runs.py) and the dispatch-time re-capture (engine.py) run while the patch is on disk, so it lands in the snapshot and stops counting.

### LINKED ISSUE #89: Baseline-match verify gate is dead code for generic bmad-dev-auto sessions (baseline_commit vs baseline_revision)
## Context

Found by the PR #86 final review (adversarially verified). Pre-existing on `main`; masked by test fixtures.

## The gap

`_verify_shared_gates` reads `fm.get("baseline_commit", "")` from the spec frontmatter and **skips the whole baseline-match check when the key is absent** (verify.py ~1055-1061). The generic `bmad-dev-auto` skill never writes `baseline_commit` — its step-03 stamps **`baseline_revision`** ('Capture `baseline_revision` (current HEAD…) into `{spec_file}` frontmatter'). The only `baseline_revision → baseline_commit` translation lives in `devcontract`'s result-synthesis, which the gate does not consult (it re-reads frontmatter directly).

Net: in production the baseline-match gate never fires for generic-skill sessions — a spec claiming a stale/foreign baseline sails through.

## Why the tests didn't catch it

`tests/conftest.py`'s `write_spec` (and the engine-test session fakes) emit `baseline_commit:` directly, fabricating exactly the key the real skill never writes — so the gate looks alive and matching under test.

## Fix shape

- Gate reads both keys (`baseline_commit`, falling back to `baseline_revision`) or a single normalizing reader.
- Make the fixtures honest: `write_spec` should stamp `baseline_revision` like the real skill, so a regression here fails a test.
- Note the interaction with #86's patch-restore path: re-arm now re-stamps `baseline_revision` (review F2), so once the gate reads it, restore re-drives stay consistent by construction.

pbean:
Closes #88, Closes #89

Both bugs were surfaced by the **PR #86 final review** (the intent-gap/patch-restore audit), where each was adversarially verified against `main` and filed rather than folded into that PR — they are pre-existing, not regressions from #86. They land together because both fixes are in the same ~15-line block: `_verify_shared_gates`, the workflow-tag / expected-status / baseline-match / proof-of-work gates shared verbatim by `verify_dev`, `verify_dev_bundle` and `verify_dev_stories`.

### #88 — the proof-of-work gate now subtracts the baseline snapshot

`has_changes_since` counted **every** untracked file and never subtracted `baseline_untracked` (contrast `attempt_dirty` directly below it, which does). After any intent-gap halt the saved patch is untracked residue under `implementation_artifacts`, which `_protected_relpaths` shields from every reset. #86's T4 fix excludes the *latched* patch by path, but a **from-scratch** re-arm (`restore_patch=None` — the common "the attempted reading was wrong" case) has no latch and never learns the path. A re-driven session that produced zero code changes but flipped the spec status passed the gate on the patch file's mere presence, and `finalize_commit`'s `add -A` then swept it into the story commit.

`has_changes_since` gains a keyword-only `baseline_untracked`, subtracted when non-None; the sole production caller passes `task.baseline_untracked`. Both snapshot writers (re-arm in `runs.py`, dispatch-time re-capture in `engine.py`) run while the patch is on disk, so it lands in the snapshot and stops counting.

**`None` deliberately keeps counting all untracked files** — the *opposite* of `attempt_dirty`'s `None` = ignore-all. Pinned in the docstring, because it reads like an inconsistency and isn't: the two gates fail open in opposite directions. A proof-of-work gate must fail open toward "work happened" (a pre-snapshot run must not have its gate silently weakened into never seeing new files); a rollback gate must fail open toward "nothing to remove" (never delete a file it cannot prove this attempt created).

### #89 — the baseline-match gate reads the key the skill actually writes

The gate read `fm.get("baseline_commit", "")` and **skips the whole check when the key is absent**. The generic `bmad-dev-auto` skill's step-03 stamps **`baseline_revision`**; `baseline_commit` exists only in the `result.json` `devcontract` synthesizes, which the gate does not consult (it re-reads frontmatter directly). Net: in production the baseline-match gate never fired for generic-skill sessions, and a spec claiming a stale or foreign baseline sailed through.

The gate now reads both keys via the exact idiom already in `devcontract.synthesize_result`. It composes with #86's review-F2 re-stamp (re-arm rewrites `baseline_revision`), so patch-restore re-drives stay consistent by construction.

**Fixture honesty.** The bug was masked because `conftest.write_spec` emitted `baseline_commit:` — fabricating precisely the key the real skill never writes, which made the gate look alive under test. `write_spec` now stamps `baseline_revision` like the skill, and `_spec_baseline` reads either key. This is the load-bearing part: with the fixture honest but the reader unfixed, the two **pre-existing** tests `test_verify_dev_lying_baseline` and `test_verify_dev_stories_lying_baseline` fail — they had been passing vacuously. No other suite breakage surfaced, so no spec stamps both keys.

### Tests

- `has_changes_since`: snapshot subtraction, `None` counts everything, post-snapshot file counts, tracked edit counts regardless of snapshot.
- Gate level, #88: baseline-era untracked residue + a bare status flip → `retry("no changes…")` with no latch in play; `baseline_untracked=None` → still passes (the fallback pinned); real work still passes with the snapshot in place.
- Gate level, #89: a spec carrying only a *mismatched* `baseline_revision` fires the gate (zero coverage before — this was the masked path); a matching one passes.

Mutation-checked: reverting either source fix fails exactly the corresponding new tests (plus the two now-honest pre-existing ones).

### Verification

`.venv/bin/python -m pytest` — 1930 passed, 1 skipped. Full `trunk check` (no filter) clean.

### Follow-up ordering note

Per the bundling plan, **#91(d) must land after this**: it rewrites the exact `has_changes_since` call this PR modifies (renaming `proof_exclude` → `extra_exclude` and deriving the restore-patch exclusion inside the gate). The `baseline_untracked=task.baseline_untracked` argument added here has to survive that refactor verbatim.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed verification so pre-existing untracked files no longer count as new work in baseline checks.
  * Improved baseline matching to recognize both revision-style and commit-style baseline fields, reducing false failures in dev verification.
  * Tightened proof-of-work checks so leftover untracked artifacts are handled correctly.

* **Documentation**
  * Updated the changelog with the latest fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,22 @@ breaking changes may land in a minor release.
 
 ### Fixed
 
+- **Baseline-era untracked residue no longer vacuously satisfies the proof-of-work gate.**
+  `has_changes_since` counted every untracked file. After an intent-gap halt the saved patch is
+  untracked residue under the artifact dirs every reset deliberately protects, so a from-scratch
+  re-arm — which never learns the patch's path — let a re-driven session that produced nothing but a
+  spec status flip pass the gate on that file's mere presence, and `finalize_commit`'s `add -A` swept
+  it into the story commit. The gate now subtracts the task's `baseline_untracked` snapshot. A `None`
+  snapshot (a pre-upgrade run) still counts every untracked file — deliberately the opposite of
+  `attempt_dirty`'s ignore-all, because a proof-of-work gate has to fail open toward "work happened".
+  (closes #88)
+- **The baseline-match verify gate was dead code for generic dev sessions.** The gate read the spec's
+  `baseline_commit` and skipped itself when that key was absent — but `bmad-dev-auto` stamps
+  `baseline_revision`; `baseline_commit` exists only in the orchestrator's synthesized `result.json`.
+  In production the check never fired, so a spec claiming a stale or foreign baseline sailed through.
+  The gate now reads either key, the idiom `devcontract` already used. The test fixture stamps
+  `baseline_revision` like the real skill does, so it can no longer fabricate the key that hid this.
+  (closes #89)
 - **Unit keys with git-ref-illegal characters no longer break worktree runs.** `unit_branch_name`
   built `bmad-loop/<run_id>/<unit_key>` from the raw ids, so a key or `--run-id` carrying `:`, `..`,
   `@{`, a space or a trailing `.lock` cleared the (already-sanitized) worktree dir only to die at
--- a/src/bmad_loop/verify.py
+++ b/src/bmad_loop/verify.py
@@ -138,7 +138,13 @@ def same_commit(a: str, b: str) -> bool:
     return a.startswith(b) or b.startswith(a)
 
 
-def has_changes_since(repo: Path, baseline: str, exclude: tuple[str, ...] = ()) -> bool:
+def has_changes_since(
+    repo: Path,
+    baseline: str,
+    exclude: tuple[str, ...] = (),
+    *,
+    baseline_untracked: list[str] | None = None,
+) -> bool:
     """True if tracked changes since baseline OR untracked files exist.
 
     `exclude` is repo-relative posix dir prefixes whose changes don't count —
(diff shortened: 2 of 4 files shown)


pbean:
Addressed the CodeRabbit nitpick in 86972e9: the baseline-mismatch retry message now says `spec baseline` instead of hardcoding `baseline_commit`, since the value can come from either key (and after this PR typically comes from `baseline_revision`). No test changes needed — assertions on this path only check the `does not match` substring.

Issue #90, opened 2026-07-08
Abandoned patch-restore attempt: untracked patch files become 'pre-existing' at the next re-arm and are swept into the corrected story's commit


pbean:
## Context

Found by the PR #86 final review (adversarially verified, chain confirmed link-by-link). An interaction between #86's restore latch (which supplies the untracked payload) and #78's re-arm baseline machinery (which supplies the absorption). Tracked patch hunks are safe — this is untracked-files-only.

## The chain

1. A restore re-drive escalates AFTER the patch applied (session error / CRITICAL / exhausted-mid-redrive / pre-commit veto / commit-fail — all PAUSE without a rollback; `_escalate` deliberately preserves the tree). Note: review non-convergence is a DEFER path and canNOT feed this chain.
2. The human re-resolves and re-arms **from scratch** (the documented recovery). `rearm_escalation` refreshes `task.baseline_untracked` from the current tree (runs.py ~655-662) — the applied patch's **new untracked files** are captured as 'pre-existing'.
3. Every later rollback preserves them (`safe_rollback` subtracts the snapshot), the corrected implementation proceeds around them, and `finalize_commit`'s `add -A` sweeps the abandoned attempt's files into the corrected story's commit.
4. The only remaining defense is upstream step-01's prose dirty-tree check — a probabilistic LLM judgment, exactly the guarantee class the restore design elsewhere refuses to rely on. (Tracked modifications ARE handled: parked on a preserve ref + reset at the resume's rollback.)

Worse variant: commits the escalated re-drive session made sit **below** the re-arm's refreshed `baseline_commit` and become the re-drive's permanent starting point — no reset revisits them.

## Fix shape

When a re-arm REPLACES a previously-latched restore (the escalated task's old `task.restore_patch` is non-None), clean the abandoned application first: the old latch names the patch file, so its file list is mechanically recoverable (`git apply -R` for a clean reverse, or exclude its named new files from the refreshed snapshot / delete them). Alternatively (or additionally) warn at re-arm when the tree carries untracked files that arrived after the original baseline snapshot.

PR #107, opened 2026-07-09
refactor(restore): one latch validator, one path normalizer, one exclusion site

### LINKED ISSUE #84: test helper: make_engine does not persist max_stories on RunState (resume tests must set it manually)
## Summary

The `make_engine` test helper in `tests/test_stories_engine.py` constructs its `RunState` **without** the `max_stories` field, even though production (`cli.py:404`, `cmd_run`) persists `max_stories` on `RunState` so `resume` can rebuild the cap (`cli.py:746`). The helper only passes `max_stories` to the `StoriesEngine` constructor.

Consequence: any test that runs a stories engine with `--max-stories`, pauses, and then calls `resume_engine(...)` gets `state.max_stories == None` on resume — so the resumed engine has **no cap**, diverging from production behavior. This surfaced while writing the A5 regression test (`test_max_stories_survives_a_checkpoint_pause_resume`), which had to work around it with:

```python
engine, _ = make_engine(project, [stories_dev_effect()], max_stories=2)
engine.state.max_stories = 2  # cli.py persists this on RunState so resume rebuilds the cap
```

## Why it matters

- Tests that resume with a `--max-stories` cap silently test the *wrong* (uncapped) behavior unless every author remembers the manual `engine.state.max_stories = N` line.
- It is an easy trap: `make_engine(..., max_stories=N)` *looks* like it fully configures the cap, but the durable half (RunState) is missing.

## Proposed fix

Make `make_engine` mirror production by seeding the field it already accepts as a kwarg. Roughly, when building the `RunState` inside `make_engine`, also set `max_stories` (and, for symmetry, `story_filter`/`epic_filter`) from the same kwargs it forwards to the engine — so `resume_engine` reconstructs the real launching scope without per-test boilerplate.

After the helper is fixed, drop the manual `engine.state.max_stories = 2` workaround in `test_max_stories_survives_a_checkpoint_pause_resume`.

## Context

Found during PR #77 review remediation (finding **A5** — durable `--max-stories` gate, commit `e56f9db`). Not a product bug; a test-fidelity gap that could mask future resume-durability regressions.

## Acceptance

- [ ] `make_engine` persists `max_stories` (and scope selectors) on the `RunState` it creates, matching `cmd_run`.
- [ ] The workaround line in `test_max_stories_survives_a_checkpoint_pause_resume` is removed and the test still passes.
- [ ] A `resume_engine` round-trip test asserts `state.max_stories` survives without manual seeding.

### LINKED ISSUE #91: Patch-restore cleanup batch: centralize latch preconditions, one path normalizer, test-fixture dedup

pbean:
Closes #91, Closes #84

The last of the four PRs bundling the open issues from the **PR #86 final review** (after #103, #105, #106). Both issues here are non-bug cleanups that review raised: #91 is the restore-seam duplication it flagged across `runs`/`cli`/`engine`/`verify`, #84 is the test-fidelity gap found while writing PR #77's A5 regression test. Behavior-neutral by intent, with one deliberate exception noted below.

## #91(a) — one latch validator

`runs.validate_restore_latch` now holds every patch-restore precondition: the sentinel wedge and spec-less escalation (previously in `rearm_escalation`) and the worktree-isolation rejection (previously only in `cli._resolve_restore_patch`).

**The one intentional behavior change:** `rearm_escalation` now rejects a worktree-isolation restore as well. That split was the actual bug in #91 — any non-CLI caller (TUI restore parity, a script) could latch a patch the re-drive can never honor, because it discards the unit's worktree and re-mounts a fresh one. The CLI still passes `worktree_isolation` from the *live* policy, the one input run state cannot carry, so a policy edit between escalation and resolve can't skew the guard.

Path resolution and `spec_within_roots` containment stay CLI-side — they need `--project` and the loaded bmad config.

## #91(b) — one path normalizer

`verify.resolve_restore_path(raw, root)` replaces the four copies of the maybe-relative→absolute join (`engine._restore_patch` vs `workspace.root`, `verify.verify_dev_exclude_relpaths` vs `paths.project`, `cli._resolve_restore_patch` vs `project`, and the fourth #90 added in `runs._stale_restore_residue`). Each consumer resolves against the base it actually reads the tree from, so `root` stays caller-supplied. The relative branch is kept — `model.StoryTask.restore_patch` promises repo-relative-or-absolute — and documented as CLI-latched-absolute in practice. The CLI keeps its own `.resolve()` on top, since it is the one consumer feeding a containment check.

## #91(c) — vacuous phase guard

`engine._restore_patch`'s `if task.phase == Phase.DEV_RUNNING` was always true at its sole call site (`_dev_phase` advances to `DEV_RUNNING` immediately before dispatch), and the comment's PENDING case couldn't survive `_escalate` anyway. Replaced with an unconditional advance + call-site-invariant comment.

## #91(d) — exclusion derived at the gate

`_verify_shared_gates`' `proof_exclude` becomes `extra_exclude`: the gate now derives `verify_dev_exclude_relpaths(paths, spec_path, task.restore_patch)` from the `task` it already receives, and each mode passes only what it adds — `()` for sprint and bundle, the story record + manifest for stories. Threading the restore patch in from three call sites left a default-`None` foot-gun for a future fourth mode, which would silently let a restore re-drive pass proof-of-work on the patch file's mere presence. `extra_exclude=None` still skips the gate outright for the plan-halt leg, and #88's `baseline_untracked=task.baseline_untracked` argument is preserved verbatim.

## #91(e) — test dedup

The three `_escalated_run` copies were **not** near-identical (different defaults, different return tuples, one unique kwarg each), so conftest gets a **parameterized superset** builder returning an `EscalatedRun` record; each module keeps a thin adapter for its own shape. Every field a caller relied on is still reachable — nothing is weakened down to a lowest common denominator. The E2E scaffolds loop `install.DEV_BASE_SKILLS` (and the `STORIES_PROBE_*` constants) instead of hand-rolling the skill list three times, so a newly required skill fails them loudly. `conftest.dev_effect` absorbs `test_engine._restoring_dev_effect` via a `seen` hook + `write_src=False` flag.

## #84 — `make_engine` persists the launching scope

`tests/test_stories_engine.make_engine` now seeds `max_stories`/`story_filter`/`epic_filter` on the `RunState` it builds, mirroring `cli.cmd_run`. `resume` rebuilds the cap and filters from run state alone, so the old helper made every `resume_engine` test silently run **uncapped** unless the author remembered `engine.state.max_stories = N`. The workaround in `test_max_stories_survives_a_checkpoint_pause_resume` is dropped and a round-trip test pins the scope surviving unseeded.

## Verification

- `.venv/bin/python -m pytest` — **1950 passed, 1 skipped** (5 new tests; `main` was 1945 + 1).
- Full `trunk check` (no filter) — clean.
- Every test-dedup diff reviewed hunk by hunk for silently weakened assertions. One was caught in the act: an edit initially truncated `test_rearm_rejects_restore_patch_without_a_spec_file`'s tail assertions; restored and re-verified.

New coverage: the stories gate excludes a latched restore patch (guards the #91(d) derive), `resolve_restore_path` join/no-normalize semantics, `rearm_escalation` rejecting a worktree-executed task's latch, `validate_restore_latch`'s clean-pass + live-policy legs, and the #84 resume round-trip.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved restore-patch handling so restore paths are resolved more consistently and worktree-isolated runs are rejected when restore isn’t supported.
  * Strengthened validation around escalation re-arming and restore flows to prevent invalid or stale patch reuse.
  * Resume behavior now preserves story selection limits and filters more reliably across runs.

* **Documentation**
  * Updated the changelog with the latest restore, escalation, and test-fixture improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,9 @@ breaking changes may land in a minor release.
   programmatically could latch a patch the re-drive can never honor; it now rejects it too.
   `verify.resolve_restore_path` replaces four copies of the maybe-relative→absolute join, and the
   shared verify gate derives the restore-patch proof-of-work exclusion from the task instead of
-  threading it in from three call sites. Otherwise behavior-neutral. (closes #91)
+  threading it in from three call sites. The resolve context's `restore_supported` signal is now the
+  validator's verdict too, so the agent never negotiates a restore for a sentinel-wedged or spec-less
+  escalation either. Otherwise behavior-neutral. (closes #91)
 - **Test helper fidelity.** `make_engine` seeds the launching scope (`max_stories`, `story_filter`,
   `epic_filter`) on `RunState` like `cmd_run` does, so resume tests no longer silently ran uncapped;
   the three `_escalated_run` fixtures collapse into one parameterized conftest builder. (closes #84)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,19 @@ breaking changes may land in a minor release.
   `customize.toml` in `bmad-dev-auto` (its review-layer config, BMAD-METHOD#2535/#2550). A pre-July bmm
   install missing any is reported with remediation before a run stalls.
 
+### Changed
+
+- **The patch-restore seam is now one validator, one path normalizer, and one exclusion site.**
+  `runs.validate_restore_latch` holds every latch precondition (sentinel wedge, spec-less escalation,
+  worktree isolation) — the worktree check lived only in the CLI, so `rearm_escalation` called
+  programmatically could latch a patch the re-drive can never honor; it now rejects it too.
+  `verify.resolve_restore_path` replaces four copies of the maybe-relative→absolute join, and the
+  shared verify gate derives the restore-patch proof-of-work exclusion from the task instead of
+  threading it in from three call sites. Otherwise behavior-neutral. (closes #91)
+- **Test helper fidelity.** `make_engine` seeds the launching scope (`max_stories`, `story_filter`,
+  `epic_filter`) on `RunState` like `cmd_run` does, so resume tests no longer silently ran uncapped;
+  the three `_escalated_run` fixtures collapse into one parameterized conftest builder. (closes #84)
+
 ### Fixed
 
 - **An abandoned patch-restore no longer smuggles its files into the corrected story's commit.**
--- a/src/bmad_loop/cli.py
+++ b/src/bmad_loop/cli.py
@@ -811,6 +811,7 @@ def _resolve_restore_patch(
     story_key: str,
     args: argparse.Namespace,
     pol,
+    state,
     task,
 ) -> tuple[str | None, str | None]:
     """Determine the intent-gap patch-restore latch (BMAD-METHOD #2564) for a re-arm.
@@ -825,7 +826,12 @@ def _resolve_restore_patch(
     outside the trusted roots, the run can't restore in place, or the restore
     input itself is corrupt (unreadable resolution.json, empty/non-string value)
     — the caller aborts strictly rather than silently re-driving from scratch
-    when a restore was (or may have been) asked for."""
+    when a restore was (or may have been) asked for.
+
+    The run-state preconditions come from ``runs.validate_restore_latch``, shared
+    verbatim with ``runs.rearm_escalation``; only the CLI-side halves live here —
+    path resolution against ``--project`` and trusted-roots containment, which need
+    the loaded bmad config."""
     raw = getattr(args, "restore_patch", None)
     if raw is not None and not raw.strip():
         # `--restore-patch ""` is a classic unset-shell-var slip. Treating it as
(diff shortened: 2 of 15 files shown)


pbean:
Thanks — applied in 0130b18. The suggestion was worth more than a nitpick: `StoriesEngine.__init__` explicitly nulls the `story_filter`/`epic_filter` **constructor kwargs** (a flat story list has no E-S refs), so a reader could reasonably worry that construction clobbers the durable scope on `RunState` too. It doesn't — the base `Engine` parks the filters on `self`, never back onto `self.state` — and now the test pins that instead of leaving it to inspection.

Both suggested assertions added, plus a comment naming the invariant. Full suite still 1950 passed / 1 skipped; `trunk check` clean.

Issue #100, opened 2026-07-09
[BUG] bmad-loop resume requests rollback of already-committed work when orchestrator dies after   successful dev session


guilherme-argentino:
### Description

bmad-loop resume requests rollback of already-committed work when orchestrator dies after successful dev session

### Steps to reproduce

  1. Run `bmad-loop` — let it drive a dev story to completion
  2. Dev session finishes: commit pushed, `sprint-status.yaml` updated to `done`
  3. Orchestrator process is killed before it processes the `Stop` event (crash, OOM, or
  signal)
  4. Run `bmad-loop resume`
  5. Resume emits rollback instruction and treats the session as a failed attempt

### Expected behavior

  Resume should check `sprint-status.yaml` before issuing any rollback. If the story status is
  `done`, treat the session as complete and advance — skip rollback entirely.

### Actual behavior

  Resume reads only `state.json` (stale — reflects pre-crash state), ignores
  `sprint-status.yaml`, and instructs `git reset --hard <baseline>`, which would undo
  already-pushed commits.

### Which area is this for?

Orchestrator / control loop

### bmad-loop Version

0.8.1

### Which coding CLI are you using?

Claude (claude)

### Operating System

Linux

### Relevant log output

```shell

```

### Confirm

- [x] I've searched for existing issues
- [ ] If I attached a diagnose dump, I reviewed it and confirmed it contains no PII, secrets, or project-specific information
- [x] I'm using the latest version

guilherme-argentino:
## Diagnostic data from a live reproduction

Reproduced on 0.8.1 / Linux / Claude adapter.
 
### What state.json showed on resume

The session had completed successfully (`result_json.status: "done"`), but `commit_sha` was
`null`:

```json
  "phase": "dev-verify",
  "baseline_commit": "9d59b70a3f6846f1b5d42ac06753df32f4f0c3d7",
  "commit_sha": null,
  "sessions": [{
    "status": "completed",
    "result_json": {
      "status": "done",
      "baseline_commit": "",
      ...
    }
  }]
```

  The paused_reason asked to git reset --hard 9d59b70 — the baseline before the story's work.
  Git HEAD was already 3 commits ahead of that, all pushed to origin.

### Root cause (narrowed down)

  Stop and SessionEnd events fired 0.2 ms apart:

  ...395074 — Stop
  ...395269 — SessionEnd

  The orchestrator read the Stop event, extracted result_json (status: done), then died before 
[...]

PR #117, opened 2026-07-10
fix(engine): finish the commit on resume for tasks persisted in the COMMITTING window

### LINKED ISSUE #115: resume-restart discards verified work when the host dies in the COMMITTING window (no session record to replay)
## Context

Sibling of #100, split out of its fix (same hazard class, different machinery needed).

#100's fix extended the resume replay matcher (`Engine._resumable_session`) to the `DEV_VERIFY`/`REVIEW_VERIFY` phases, so a completed session's durably-recorded result is replayed through the normal verify/decide pipeline instead of falling through to resume-restart's rollback. That works because those windows have a session record to replay.

## The remaining window

`_commit` persists `phase: committing` (the save right after `advance(task, Phase.COMMITTING)`) **before** `verify.finalize_commit` runs and the `DONE` save stamps `commit_sha`. A host death in that window leaves a task at `COMMITTING` whose work is fully verified (dev + review + deterministic gates all passed) and sitting above baseline — either still in the worktree or already squashed by a `finalize_commit` that completed just before the death.

On resume, `COMMITTING` matches **no** arm in `_finish_inflight` — including the new replay arm, because there is no COMMITTING-keyed session record to replay (the `_skip_review_and_commit` path reaches `COMMITTING` having run no review session at all). It falls through to resume-restart → `_rollback_or_pause`.

## Mitigation already in place

The #100 fix also hardened the rollback-OFF manual-recovery notice: when the attempt has commits above baseline, the pause notice now names them, states they are intact at HEAD and may already be pushed, and never instructs a bare `git reset --hard` — so the worst symptom (an operator following a destructive instruction) is de-fanged for this window too. The work is still not auto-recovered, though.

## Suggested shape of a fix

A `_commit` re-drive keyed on phase + on-disk state rather than record replay: for a task persisted at `COMMITTING` with `task.spec_file` set, resume could re-enter `_commit(task)` (idempotence considerations: the `pre_commit_gate` workflows re-run, and `finalize_commit`'s add/soft-reset/commit sequence must tolerate both the pre-squash and post-squash on-disk states).

pbean:
## Problem (#115)

`_commit` persists `phase: committing` (the save right after `advance(COMMITTING)`) **before** `verify.finalize_commit` runs and the DONE save stamps `commit_sha`. A host death in that window leaves a task whose work is fully verified — dev, review, and deterministic gates all passed — sitting above baseline. On resume it matched no arm of `_finish_inflight`: not the spec-approval arm (wrong phase), not the #100 replay arm (there is no COMMITTING-keyed session record to replay; the `_skip_review_and_commit` path reaches COMMITTING having run zero review sessions). It fell through to resume-restart → `_rollback_or_pause`, rolling back or pausing over finished work.

## Fix

**Refactor:** the post-gate body of `_commit` moves verbatim into `_finalize_commit_phase(task)`, preconditioned on a persisted COMMITTING phase. `_commit` is unchanged in behavior: gate check → `advance(COMMITTING)` → save → helper.

**New resume arm** (`_finish_inflight`, between the replay arm and resume-restart): a task persisted at COMMITTING journals `resume-commit` and re-enters `_finalize_commit_phase` directly (with the same isolated-worktree reopen/integrate wrapper as the other arms). Design points:

- **`pre_commit_gate` workflows are deliberately NOT re-run.** The COMMITTING save lands only after the gate loop returns clean, so the persisted phase is durable proof they passed; the tree is unchanged since (crash — no sessions ran). A re-run would double-charge the session budget through `_run_session`, and a blocking failure would need an illegal `COMMITTING→DEFERRED` move.
- **The `pre_commit` hook DOES re-fire** — the commit message is regenerated, so a plugin rewrite reaches the squash, and a pause veto still escalates (`COMMITTING→ESCALATED` is legal). TEA's gate enforcement re-reads its on-disk artifacts fail-open, so re-emitting is safe.
- **`finalize_commit` needs no change.** It is content-idempotent across both crash states — pre-squash (skill commit chain above baseline) squashes normally; post-squash (squashed commit at HEAD, clean tree) re-squashes to an identical-content commit. Now pinned by a run-twice characterization test.
- `commit_sha` persisted at COMMITTING is always empty (stamped only by the DONE save) and is write-only — never routing.

**Sweep parity:** `SweepEngine._recover_inflight_bundle` had the identical hole (a COMMITTING bundle hit the rollback+PENDING reset). It gets the identical arm and closes the bundle instead of restarting it.

## Tests

- `test_resume_committing_finishes_commit_to_done` — the issue's environment (rollback OFF): resume → DONE, one squashed commit above baseline, `commit_sha` stamped, zero new sessions, no `resume-restart`/`rollback-manual-required`.
- `test_resume_committing_post_squash_still_one_commit` — finalize completed just before the death: re-drive converges on exactly one commit, no stacked second squash.
- `test_resume_committing_git_error_escalates` — a rejecting hook during the re-drive → ESCALATED with the attempt chain intact at HEAD.
- `test_worktree_resume_committing_finishes_and_merges` — isolated flavor: finished inside the still-mounted worktree and merged back.
- `test_pre_commit_gate_workflow_not_rerun_on_commit_resume` — a blocking `pre_commit_gate` workflow is not re-charged on the re-drive.
- `test_pre_commit_hook_fires_on_commit_resume` / `test_pre_commit_pause_veto_on_commit_resume_escalates` — hook re-fires; rewrite reaches git; pause veto escalates legally.
- `test_finalize_commit_rerun_is_content_idempotent` — run-twice pin (tree identity, single commit; no sha-inequality assertion — same-second re-runs re-mint the same sha).
- `test_resume_committing_bundle_finishes_commit` — sweep flavor, production-faithful: a real run crashed on the `pre_commit` emit, resumed to DONE with the ledger closed.

Shared fabricator `committing_crash_state` in conftest persists the exact state.json shape from the issue (attempt commit above baseline — only the work file, keeping the sprint board untracked as production does; spec `done`; sprint synced at dev time; `review_cycle=0` for skip-review fidelity).

Full suite: 2067 passed, 1 skipped. `trunk check` clean.

Closes #115

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resume handling after crashes during commit finalization.
  * Prevented completed stories and sweep bundles from being restarted or losing progress.
  * Ensured resumed commits avoid repeating completed gate workflows and produce a single clean commit.
  * Improved recovery for isolated workspaces and post-squash crash states.
  * Added safeguards for commit failures and pre-commit pauses during recovery.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,15 @@ breaking changes may land in a minor release.
 
 ### Fixed
 
+- **Resume no longer discards a story that already passed its pre-commit gates.** A host death in
+  the COMMITTING window (phase persisted before `finalize_commit` ran and the DONE save stamped
+  `commit_sha`) matched no resume arm — there is no COMMITTING-keyed session record to replay — and
+  fell through to resume-restart, rolling back or pausing over fully-verified work. Resume now
+  finishes the commit in place: the `pre_commit_gate` workflows are not re-charged (the persisted
+  phase is durable proof they passed), the `pre_commit` hook re-fires (message regeneration and
+  pause veto honored), and `finalize_commit`'s content-idempotence covers both the pre- and
+  post-squash crash states. Sweep bundles get the same recovery in `_recover_inflight_bundle` (#115).
+
 - **Resume no longer asks for a rollback of a completed session's committed work.** A host death
   in the post-verify decision window left the task persisted at `DEV_VERIFY`/`REVIEW_VERIFY`,
   where the resume replay matcher (which only knew the `*_RUNNING` phases) missed the
--- a/src/bmad_loop/engine.py
+++ b/src/bmad_loop/engine.py
@@ -1184,6 +1184,27 @@ class Engine:
                     self._integrate_unit(task, unit)
                 else:
                     continuation()
+            elif task.phase == Phase.COMMITTING:
+                # the host died in the commit window: the gate+advance save
+                # landed (pre_commit_gate ran clean) but the DONE save that
+                # stamps commit_sha did not. Finish the commit instead of
+                # rolling verified work back — the gates are deliberately NOT
+                # re-run (see _finalize_commit_phase), the pre_commit hook IS
+                # re-emitted (message regenerated; pause veto still honored),
+                # and finalize_commit tolerates both the pre- and post-squash
+                # crash states (#115).
+                self.journal.append("resume-commit", story_key=task.story_key)
+                if isolated:
+                    unit = self._reopen_unit(task)
+                    prev = self.workspace
+                    self.workspace = unit.workspace
+                    try:
+                        self._finalize_commit_phase(task)
+                    finally:
+                        self.workspace = prev
+                    self._integrate_unit(task, unit)
+                else:
+                    self._finalize_commit_phase(task)
             else:
                 self.journal.append(
                     "resume-restart", story_key=task.story_key, phase=str(task.phase)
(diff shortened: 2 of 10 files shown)


pbean:
@coderabbit review

Issue #140, opened 2026-07-14
herdr: native-Windows launch surface (agent.start launch, pwsh parked dialect, Windows-host re-characterization)


pbean:
The herdr backend (#136 run path, #137 TUI-launch surface) ships end-to-end on POSIX. Its native-Windows launch surface was deliberately deferred ("PR-6" in the PR bodies) and currently has no open tracker — the umbrella #92 is closed. This issue is that residue.

## Remaining work

1. **win32 launch path via herdr `agent.start`** — the POSIX launch types `exec <argv>` into a fresh tab so process-exit == pane-close == window death (`herdr_backend._launch`); `exec` has no Windows equivalent, and `agent.start` (which splits the active tab) is the reserved win32 launch verb. Source: `herdr_backend.py` module-docstring degradation ledger + the `_launch` comment.
2. **pwsh parked-window dialect** — `new_parked_window` types a POSIX `exec sh -c '<argv>; ec=$?; banner; read -r; trailer'` recipe (the trailer `cat`s the per-window return file). The pwsh dialect of the recipe + trailer is deferred alongside the launch path. Source: degradation ledger ("POSIX-only, like the `exec` launch (pwsh dialect is the Windows follow-up)").
3. **Re-run Phase-0 characterization on a real Windows host** — the shipped backend is POSIX-verified / Windows-by-construction; no Windows box was available during #136/#137. Per `docs/ROADMAP.md`: "the Phase-0 characterization must be re-run on a real Windows host before #92 is claimed."

## Non-goals here

- The socket transport replacing subprocess `_HerdrClient` is a separate optional enhancement (the class boundary already isolates it).
- The sibling tmux-family win32 backends (psmux re-land: #58; tmux-windows: #85) are tracked separately — herdr is pairwise-discriminating against them by construction (`which("herdr")`).

## Selection note (no code change needed when this lands)

herdr registers `matches=True` but `_PLATFORM_DEFAULTS` still declares psmux the win32 default, so herdr activates on win32 only via first-match while no higher-priority backend is available — once psmux ships, the default reasserts automatically.

Refs: #92 (closed umbrella), #136, #137, #58, `docs/ROADMAP.md` "Native Windows multiplexer backend", `docs/porting-to-a-new-os.md`.

pbean:
[...]

Issue #58, opened 2026-07-03
Native-Windows psmux multiplexer backend: re-land as a thin leaf over the dialect seams


dracic:
## Context

PR #19 (`feat/psmux-backend`) was the exploratory implementation of a native-Windows terminal-multiplexer backend driving psmux's `tmux.exe` drop-in (ConPTY, no WSL). It proved the approach end-to-end — real dev sessions ran on native Windows — but it was closed rather than merged: it predates the `bmad-loop` rename and the adapter seams, and exercising it surfaced engine-level defects that had to land on `main` first. This issue tracks re-landing the backend as a fresh, thin leaf on top of today's seams.

## What the test PR/branch surfaced

- **Engine died on uncaught adapter exceptions** — a `list-windows` timeout leaked `subprocess.TimeoutExpired` out of the polling loop and killed the engine mid-run. Fixed on main: crash recording + multiplexer seam honesty (0.7.9/0.7.10).
- **Completion misread → destructive auto-rollback** — spec status sampled mid-transition; `git reset --hard` erased commits the attempt itself made. Fixed on main: #29 (preserve attempt commits, follow-ups #32/#49 → PRs #50/#54), #44 (completion contract). Residual completion/durability reports tracked in #34/#48/#52/#57.
- **Windows process/signal hazards** — `os.kill(pid, 0)` is `CTRL_C_EVENT` on win32 (phantom `KeyboardInterrupt`s), PID reuse, unreadable-but-existing PIDs. Fixed on main: `WindowsProcessHost`, persisted PID identity, identity-aware liveness (#12, #30, #33, #38).
- **psmux behavioral divergences** (verified against the 3.3.6 source and live probes, not assumed from docs): cp1252 garbles `-F` output; no `/bin/sh` for the parked-window recipe; the nesting guard silently no-ops `new-session` from inside a pane (exit 0, nothing created); `kill-session -t =name` is silently ignored (plain name works); agent-teams env vars are injected at the server level; no native `cat` for `pipe-pane`.

## Preparation that already landed on main

- #40/#41 — `_run` spawn seam carries output decoding (`_ENCODING` class attr) and a per-call `env`; no `_run` override needed.
[...]

PR #141, opened 2026-07-14
feat(mux): seam-canonical window targets

pbean:
## What

Makes the window-target contract backend-neutral. Core (`runs.py`, `tui/launch.py`, `tui/app.py`) used to hand-assemble tmux-shaped `=session[:window]` exact-match strings in 10 places and push them through the seam — so every non-tmux backend had to reverse-engineer tmux target syntax (herdr's `_parse_target` did; the psmux re-land #58 would have next). The grammar is now owned by the seam:

- **`TerminalMultiplexer.target(session, window=None)`** — new *concrete* encoder, default emits the now-seam-canonical `=session[:window]` token. Backends MAY override to emit native ids; the docstring pins the constraint (the token must stay a stable by-name reference — core formats targets ahead of use, e.g. a parked window's return target, so eager resolution goes stale).
- **`multiplexer.parse_target()`** — the matching module-level decoder: `(session, window|None)` for a canonical token, `None` for a backend-native id. One tested helper instead of per-backend re-derivation.
- Core's 10 f-strings swap to `target()` via two thin helpers that preserve layering: `runs.session_target(run_id)` and `launch.ctl_target(window=None)` (`tui/app.py` keeps importing only `runs`/`launch`, never the mux).
- herdr's `_parse_target` delegates its grammar split to `parse_target()`; its workspace/tab/pane resolution (and strict/tolerant semantics) is untouched. herdr deliberately does **not** override `target()` — pinned by a test.
- tmux backends: zero behavior change — the canonical grammar coincides with tmux exact-match syntax, so `BaseTmuxBackend` keeps passing targets straight through (now noted in its class docstring).
- Target-taking seam methods' docstrings state the two-family rule (canonical token | native id); the adapter authoring guide gains a **Window targets** section.

## Zero behavior change, by construction

The default `target()` emits byte-identical strings to the old f-strings, so **no existing test assertion was edited** — the exact-argv pins in `test_tui_launch.py`, `test_runs.py` (attach_argv with and without `TMUX`), `test_cli.py`, `test_tui_app.py`, and the herdr grammar-acceptance tests in `test_herdr_backend.py` all pass unchanged and double as the compliance proof. New tests in `test_multiplexer.py` cover the encoder grammar, the `parse_target` round-trip/edges/native-passthrough, and herdr's inherit-don't-override choice.

Out-of-tree adapters are unaffected: all string-typed method signatures and semantics are unchanged, `target()` is concrete and inherited. Adapters that already decode the grammar should migrate to `parse_target()` — it is now documented contract, not reverse-engineered tmux syntax.

## Verification

- Full suite green **twice** — with `TMUX` in the env and with `env -u TMUX` (this dev box runs the agent inside tmux).
- `grep -rn 'f"={' src/bmad_loop` → only backend files (`tmux_base.py`) and the seam default itself.
- Full `trunk check` clean.

Follow-up: a small separate PR renames the seam-backed-but-tmux-named helpers (`launch.tmux_available`, `app._tmux_missing`, `runs.tmux_sessions`) so adapter authors aren't misled; kept out of this diff so the "zero assertions changed" property stays reviewable.

Refs: #136, #137 (stacked base), #58, #140.

## Summary by CodeRabbit

* **New Features**
  * Added consistent window-target formatting and parsing across terminal multiplexer backends.
  * Added support for canonical session and window target references, including reusable helpers for attaching to control and agent sessions.
  * Improved backend handling of canonical and native target identifiers.

* **Documentation**
  * Documented target representation, resolution, and backend customization guidance.
  * Added changelog details for the new target contract.

* **Tests**
  * Added coverage for target encoding, decoding, round trips, edge cases, and native identifiers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,15 @@ breaking changes may land in a minor release.
   forced selection. A tmux-less POSIX host still selects `TmuxMultiplexer` and reports it
   unavailable, exactly as before.
 
+- **Seam-canonical window targets.** The `=session[:window]` target grammar is now owned by the
+  `TerminalMultiplexer` seam instead of living as hand-assembled tmux syntax in core: a new
+  concrete `target(session, window=None)` encoder (overridable per backend, tmux inherits the
+  default and passes it straight through) and a module-level `parse_target()` decoder that
+  native-id backends reuse instead of re-deriving the grammar (the herdr backend's
+  `_parse_target` now delegates to it). `runs.py`/`tui/launch.py`/`tui/app.py` format every
+  target via `target()` (new `runs.session_target` / `launch.ctl_target` helpers) — output is
+  byte-identical, so no backend or operator behavior changes; the contract is documented in the
+  adapter authoring guide's new "Window targets" section.
 - **Herdr TUI-launch surface.** The herdr backend now covers everything `tui/launch.py` drives:
   parked orchestrator windows (a typed `exec sh -c '<argv>; banner; read; trailer'` recipe,
   tmux-identical from the operator's seat, with the return-to-origin target mirrored into a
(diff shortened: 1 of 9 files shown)


pbean:
@coderabbit review

pbean:
@coderabbit review

Issue #126, opened 2026-07-12
Worktree seeding drops file modes; verify rc=126/127 is charged as a story failure


gnometa:
**Version:** bmad-loop 0.8.1 · **Component:** `install.py` (`_copy_traversable`) + engine verify-gate classification
**Reproduction:** run `20260709-172451-0187` (Laravel project, claude adapter, worktree isolation, 23 stories, ~20h unattended)

## Problem

`scm.worktree_seed` copies gitignored paths (e.g. `vendor/`, `node_modules/`) into each fresh worktree via `dst.write_bytes(src.read_bytes())`, which discards file modes. Every seeded worktree therefore has non-executable `vendor/bin/*`, and the first verify command dies with rc=127 (`pint: Permission denied`).

The engine cannot distinguish this environment fault from broken story code. Observed impact on the reference run:

- Consumed 1 of `max_dev_attempts = 2` on **4 of 23 stories**, spawning ~16.4M tokens of no-op "repair" sessions that fixed nothing about the code.
- 13+ *passing* stories show dev sessions silently chmod-ing the same fault in-session (wasted tokens on every story).
- On one story it was the but-for cause of a false deferral: the environmental retry exhausted the attempt budget before an unrelated bookkeeping issue surfaced.

## Proposed fix

1. Preserve file modes in the seed copy (`shutil.copy2` / `cp -a` semantics).
2. Classify verify exit codes 126/127 as **environment failures** that are retried without decrementing `max_dev_attempts`.

Workarounds we ship in the meantime: a `chmod +x vendor/bin/*` guard as the first verify command, and invoking tools via `@php vendor/bin/...` so no exec bit is needed.

PR #152, opened 2026-07-15
fix(worktree): detach kept worktree so branch_per=run survives a kept failure (#138)

### LINKED ISSUE #138: `branch_per=run` + `keep_failed=true`: the first kept-failed story's worktree holds the run branch, so every subsequent story in a multi-story run insta-defers on a branch/worktree collision
## Environment
- bmad-loop **0.8.1** via `uv tool install` (latest release, v0.8.1 2026-07-06; no fix on `main` as of 2026-07-13)
- Policy: `branch_per = "run"`, `keep_failed = true`, `isolation = "worktree"`, `target_branch = "dev"`, `max_parallel = 1`
- A run covering multiple stories (observed with `--epic 5`, three stories remaining in the epic)

## Summary
When a run uses `branch_per=run` **and** `keep_failed=true`, the **first** story that ends in a terminal deferred state has its worktree preserved by `close_unit_workspace` — worktree *and* branch left mounted (`workspace.py`, the `keep_failed` early-return). Under `branch_per=run` that branch is the single shared run branch, and git refuses to check out a branch that is already checked out in another worktree. So **every subsequent story's `open_unit_workspace` fails, and the story is deferred with zero dev activity of its own**.

To be clear about what the tool already does right: each collision *is* surfaced — a distinct `worktree-open-failed` journal event, a desktop notification, an ATTENTION line with the full git error, and a `defer_reason` that records the collision verbatim. The problem is that the run **keeps going** and burns every remaining story in milliseconds. `engine.py` (the `except verify.GitError` branch in `_run_isolated`) deliberately defers rather than crashing:

> could not mount a worktree (e.g. branch_per=run with a kept-failed unit still holding the shared branch). Defer this unit rather than crash the whole run; the operator can free the branch and re-run.

That trade-off is sound for an *incidental* mount failure. But under `branch_per=run` this failure is not incidental — once one kept-failed unit holds the shared branch, the collision is **structurally guaranteed for every remaining unit**. The run's own top-line summary then conflates the two failure kinds: this run reported "0 done, 3 deferred", when 1 story was attempted-and-deferred and 2 were never attempted at all.

Net effect: one kept-failure of story 1 converts an N-story run into a 1-story run, with the remaining N-1 stories needing a full re-run.

## Expected
Any of, in order of preference:

1. **Detach the kept worktree to free the branch** (suggested fix, minimal): when `keep_failed` preserves a worktree under `branch_per=run`, run `git checkout --detach` in it. This frees the shared branch name for the next unit while preserving everything inspection needs — the working tree, uncommitted changes, and the branch ref itself (which still points at the kept commit). Subsequent stories then mount the run branch normally and get genuine attempts.
2. **Pause the run on this collision**: since the condition is operator-actionable and guaranteed to repeat for every remaining unit, pausing (as environment faults now do after #126) beats mass-deferring.
3. Failing either, **separate the buckets**: count/report "never attempted (worktree collision)" distinctly from "attempted and deferred" in the run summary and story phase, so the cost of the cascade is at least visible at the top line.

## Actual
Story 1 ends kept-failed → `close_unit_workspace` leaves its worktree checked out on the shared run branch → stories 2..N each fail `git worktree add` (`fatal: '<run branch>' is already used by worktree at '<story-1 worktree>'`) → each is deferred untouched → run completes reporting them all in the same "deferred" bucket.

Confirmed trigger (from source): `close_unit_workspace` keeps the worktree+branch iff the unit ended `not success` **and** `keep_failed=true`. A successful unit's worktree is removed even under `branch_per=run`, freeing the branch — so kept-failure is precisely the poisoning condition.

## Observed in
A real run in a downstream project, run id `20260712-153227-1e77`, scoped `--epic 5` with three stories remaining (5-8 / 5-9 / 5-10):

- **5-8** exhausted `max_dev_attempts=2` (attempt 2 hit the 90-min `session_timeout_min`), deferred with `defer_reason: "dev session timeout"`, worktree kept per `keep_failed=true`.
- **5-9 and 5-10** both recorded `phase: deferred` with `attempt: 0`, `sessions: []`, **zero tokens**, and `defer_reason` containing the git fatal above — both within ~70 ms of 5-8's unit closing. Never attempted.
- Journal sequence: `story-deferred`/`unit-closed` (5-8) → `story-start` + `worktree-open-failed` (5-9) → `story-start` + `worktree-open-failed` (5-10) → `worktree-kept` (5-8) → `run-complete` ("0 done, 3 deferred").

All verified from `.bmad-loop/runs/20260712-153227-1e77/{state.json,journal.jsonl,ATTENTION}`.

## Suggested severity
Medium-high: with `keep_failed=true` (the shipped default posture) and `max_parallel` clamped to 1, any multi-story `branch_per=run` run dies at its first kept failure — the remaining stories are charged as deferred despite never being attempted, and the run summary gives no top-line signal that the two kinds differ.

## Workaround in use
Scope runs narrowly (`--story` / small `--epic` slices) and `--dry-run` first; after a kept failure, assume the rest of the run needs re-running. Manually freeing the branch also works: `git -C <kept-worktree> checkout --detach` before re-running. Not a fix.

pbean:
Fixes #138.

## Problem

Under `branch_per="run"` + `keep_failed=true`, the **first** story to end deferred has its worktree preserved by `close_unit_workspace` — left **checked out on the single shared run branch**. git refuses to check out a branch already held by another worktree, so **every subsequent story's `open_unit_workspace` fails** and is deferred with *zero dev activity* (`attempt=0`, no sessions, 0 tokens). One kept failure turns an N-story run into a 1-story run, and the top-line summary reports "0 done, N deferred" — conflating 1 attempted-and-deferred story with N−1 never attempted.

Confirmed end-to-end: `close_unit_workspace`'s keep-path (`workspace.py`) had no `branch_per` awareness; `unit_branch_name` returns the shared `bmad-loop/<run_id>` for every unit; the next unit's `worktree_add(create=False)` raises `verify.GitError`, which `_run_isolated` catches and defers (`worktree-open-failed`). A deferred unit returns normally, so `_loop` drains every remaining unit into the same collision.

## Fix (issue's option 1 — auto-detach, "minimal")

When `close_unit_workspace` preserves a failed worktree under `branch_per="run"`, it now runs `git checkout --detach` in it. This frees the shared branch **name** for the next story while preserving everything inspection needs — the working tree, uncommitted changes, and the branch ref (still at the kept commit). Subsequent stories mount the run branch normally and get **genuine attempts**.

Why detach over pausing (issue's option 2): unlike the #126 env-fault (unfixable by the tool), this collision **is** tool-fixable in one git command — pausing would just make the operator run that exact command before a resume helps. Detaching restores the *intended* `branch_per=run` continuation and degrades safely: if detach ever fails, the existing `worktree-open-failed` defer still surfaces the collision (no regression).

### Changes
- **`verify.py`** — new `checkout_detach(repo)` helper (`git checkout --detach`), mirroring `checkout_branch`.
- **`workspace.py`** — `close_unit_workspace` gains `detach_kept: bool`; on the `keep_failed` path it detaches the kept worktree (best-effort, swallows `GitError`).
- **`engine.py`** — passes `detach_kept=scm.branch_per == "run"` **only** at the DEFER call site. A boolean flag (not `branch_per`) lets the two callers that must not detach omit it honestly.

### Why the other two `close_unit_workspace` callers don't detach
- **success / `_merge_local`**: removes the worktree anyway → branch already freed.
- **`_keep_branch_and_escalate`**: keeps a DONE-but-merge-conflict branch for **manual merge** and immediately **pauses** the run. That path can't cascade — it halts before any sibling is dispatched, and on resume the re-armed unit's worktree is torn down by the resume-restart `discard_worktree` before any mount. A regression test (`test_branch_per_run_escalation_pauses_without_dispatching_next_unit`) locks this in.

## Tests
- `test_checkout_detach_frees_branch` (`test_verify_worktree.py`) — detach frees the branch for a sibling mount while preserving the ref + dirty tree.
- `test_branch_per_run_kept_failure_detaches_so_next_unit_runs` (`test_engine_worktree.py`) — rewritten from the old test that *encoded the bug* (`deferred==2, done==0`) to the fixed behavior (`deferred==1, done==1`, no `worktree-open-failed`, kept worktree detached).
- `test_branch_per_run_escalation_pauses_without_dispatching_next_unit` — escalate-path scoping guard.

Full suite green (2272 passed, 1 skipped); trunk clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed deferred runs using `branch_per=run` and `keep_failed=true`, preventing kept failed worktrees from blocking subsequent stories.
  * Kept worktrees are now detached while preserving their contents and branch references for inspection.
  * Improved handling of branch collisions and merge-conflict escalations without incorrectly dispatching later work.
* **Documentation**
  * Added a changelog entry describing the worktree and shared-branch fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,6 +145,18 @@ PATH)`, the TUI notifies `multiplexer backend unavailable — launch/attach disa
 
 ### Fixed
 
+- **`branch_per=run` + `keep_failed` no longer poisons a multi-story run after the first kept
+  failure (#138).** The first story to end deferred under `keep_failed=true` left its worktree
+  checked out on the single shared run branch, so every subsequent story's `git worktree add`
+  collided ("branch already checked out") and insta-deferred with zero dev activity — one kept
+  failure turned an N-story run into a 1-story run. A kept worktree under `branch_per=run` now
+  detaches its HEAD (`git checkout --detach`), freeing the shared branch name for the next story
+  while preserving the working tree, uncommitted changes, and the branch ref (still at the kept
+  commit) for inspection; subsequent stories mount the run branch normally and get genuine
+  attempts. Best effort — if the detach ever fails, the existing `worktree-open-failed` defer
+  still surfaces the collision (no regression). The escalate-and-pause path was already safe: it
+  halts the run rather than continuing, and resume frees the kept worktree before any sibling mounts.
+
 - **Dev/review sessions can no longer livelock on their own wake nudges (#149).** The idle
   wake nudge is delivered as a submitted turn, so a session that merely _answers_ it ends in
   another result-less Stop — which refilled the nudge budget, re-armed the grace window, and
--- a/src/bmad_loop/engine.py
+++ b/src/bmad_loop/engine.py
@@ -561,6 +561,7 @@ class Engine:
                 run_dir=self.run_dir,
                 unit_key=task.story_key,
                 delete_branch=scm.delete_branch,
+                detach_kept=scm.branch_per == "run",
                 diff_max_file_bytes=self._failed_diff_max_bytes(),
             )
             self.journal.append(
--- a/src/bmad_loop/verify.py
+++ b/src/bmad_loop/verify.py
@@ -647,6 +647,17 @@ def checkout_branch(repo: Path, name: str) -> None:
         raise GitError(f"git checkout {name} failed in {repo}: {out}")
 
 
+def checkout_detach(repo: Path) -> None:
+    """Detach HEAD at its current commit, leaving working tree + index untouched.
+
+    Frees a shared branch name held by a kept worktree so a sibling worktree can
+    check that branch out (git refuses a branch checked out in another worktree).
+    """
+    rc, out = _git(repo, "checkout", "--detach")
+    if rc != 0:
+        raise GitError(f"git checkout --detach failed in {repo}: {out}")
+
+
 def worktree_remove(repo: Path, path: Path, force: bool = False) -> None:
     args = ["worktree", "remove"]
     if force:
--- a/src/bmad_loop/workspace.py
+++ b/src/bmad_loop/workspace.py
@@ -119,6 +119,7 @@ def close_unit_workspace(
     run_dir: Path,
     unit_key: str,
     delete_branch: bool = True,
+    detach_kept: bool = False,
     diff_max_file_bytes: int | None = None,
 ) -> Path | None:
     """Tear down (or preserve) a unit's worktree.
@@ -129,6 +130,11 @@ def close_unit_workspace(
     happens. On success (or failure without keep_failed) the worktree is removed
     and, if delete_branch, the branch deleted. Returns the patch path it wrote,
     or None.
+
+    detach_kept (branch_per=run only): when keep_failed preserves the worktree,
+    detach its HEAD so the shared run branch it holds is freed for the next unit
+    to mount — otherwise every later unit's `git worktree add` collides on the
+    already-checked-out branch. Best effort; see the keep_failed branch below.
 
     diff_max_file_bytes caps the per-untracked-file size in that forensic patch
     (None = no cap); see verify.capture_diff.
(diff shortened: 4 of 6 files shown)


pbean:
@coderabbit review

pbean:
@coderabbitai Thanks for the review. Declining the single RUF005 nitpick (`[*a, …]` unpacking vs `_defer_script(...) + [...]` at `tests/test_engine_worktree.py:352`), intentionally:

- **Not enforced by this repo.** `pyproject.toml` selects no `RUF` rules, and the PR's `lint (trunk)`/ruff check is green — this is your profile's suggestion, not the project's linter.
- **Consistency.** Line 352 deliberately matches the identical `_defer_script(project, "1-1-a") + [...]` idiom in the sibling test it's modeled on (`test_worktree_defer_then_next_story_succeeds`, line 333). Changing only 352 would diverge from it, and widening the diff to restyle an unrelated existing test isn't warranted for a trivial preference.

Leaving the concatenation as-is.

Issue #154, opened 2026-07-15
bmadconfig: read the four-layer central TOML as primary, legacy bmm/config.yaml as fallback


pbean:
## Context

`bmadconfig.load_paths` resolves `implementation_artifacts` / `planning_artifacts` from the legacy `_bmad/bmm/config.yaml` only. The renderer-era `bmad-dev-auto` (BMAD-METHOD #2587/#2588) resolves the same keys from the four-layer central TOML (`_bmad/config.toml` + `config.user.toml` + `custom/config.toml` + `custom/config.user.toml`), with `[modules.bmm]` beating `[core]` on collision.

The BMAD installer regenerates the per-module yaml from the TOML on each install/update, so the two surfaces normally agree — but a hand-added `custom/config*.toml` override desyncs them until the next install: the skill then writes specs and the deferred-work ledger where the orchestrator is not looking.

## Current mitigation

The renderer-compat PR adds `bmadconfig.toml_artifact_parity`, a validate-time **warning** (never a FAIL) that deep-merges the four layers with the renderer's exact precedence and compares the resolved `implementation_artifacts` against the yaml-resolved path.

## Proposed fix

Make `load_paths` read the four-layer TOML as the primary source (same merge + flatten semantics as `render.py`: layers deep-merge lowest→highest, then `[core]` → `[modules.bmm]` flatten with module precedence), falling back to `_bmad/bmm/config.yaml` for pre-#2285 installs. The parity warning then becomes unnecessary and `toml_artifact_parity` can be retired.

Care points:
- `repo_root` and `output_folder` are also read from the yaml today — decide their TOML home before switching.
- `{project-root}` substitution must match render.py (forward-slash root).
- Keep `BmadConfigError` messages actionable for both surfaces.

pbean:
**Triage: `blocked:pr`.** Filed under #287 with the rest of the v6.10 TOML adoption work.

Sequenced after draft PR #155, which introduces the `bmadconfig.toml_artifact_parity` warning that
this issue proposes to retire — doing this one first would mean writing and deleting the same code.

Confirmed on `main` as of 2026-07-24: `load_paths` still resolves from the legacy surface only —
`src/bmad_loop/bmadconfig.py:70-72` builds `project / "_bmad" / "bmm" / "config.yaml"`, and the
module docstring still reads *"Resolve BMAD artifact paths from _bmad/bmm/config.yaml."* Neither
[...]

PR #155, opened 2026-07-15
fix(worktree): survive the renderer-era bmad-dev-auto (BMAD-METHOD #2587/#2588)

pbean:
## What upstream changed

BMAD-METHOD 64157d39 (#2587) made `bmad-dev-auto`'s SKILL.md a two-line shim — `uv run {skill-root}/render.py`, follow its stdout; on ANY failure "report what it printed and HALT". The old SKILL.md body moved to `workflow.md`, rendered into `{project-root}/_bmad/render/bmad-dev-auto/` with config values and **absolute paths baked in** at render time. `render.py` finds `{project-root}` by walking UP from the session cwd to the first `_bmad/` directory and requires the post-#2285 central `_bmad/config.toml`. Follow-up 1cd4a7f5 (#2588, upstream HEAD) hardened it: a `{{.var}}` referenced by the templates but missing from the merged config, or a present-but-unparseable customization layer, now hard-HALTs.

**The completion contract bmad-loop parses is unchanged** (verified against the diff): `## Auto Run Result`, `Status:` line, frontmatter statuses, `baseline_revision`, the `bmad-dev-auto-result-` fallback prefix, folder+id dispatch, and the `"folder+id dispatch"` stories-probe literal all survive; every preflight marker file still exists. `devcontract.py` is untouched. The breakage is environmental, in two critical shapes:

1. **Walk-up contamination.** Worktrees nest at `<project>/.bmad-loop/runs/<id>/worktrees/…` and never carried a `_bmad/` (default `worktree_seed = []`), so render.py silently adopted the **main checkout's** `_bmad/`: rendered workflows baked main-checkout absolute artifact paths (deferred-work appends, no-spec fallback artifacts, and step-01 context all escaped the worktree — fallback artifacts landing where `_artifact_dirs` never looks), and every parallel session raced on one shared render dir that render.py delete-rewrites per invocation.
2. **HALT before the contract exists.** The shim HALTs before `workflow.md`'s HALT protocol is ever loaded, so a broken environment (no `uv`, no `_bmad/config.toml`, missing `render.py`, or a #2588 config HALT) writes no spec status at all — every story becomes a result-less Stop (the cascade #151's breadcrumbs diagnose, one story at a time).

## What this PR does

- **`_bmad` merge-seed in `provision_worktree`**: per-FILE copy-when-absent of the main repo's `_bmad/` into the worktree (a committed `_bmad/` keeps every tracked file; the gitignored `*.user.toml` layers — whose absence hard-HALTs render.py since #2588 — are filled in), never seeding the regenerated `render/`. Shield rels follow the existing "shield exactly what we wrote" doctrine: one anchored `/_bmad` when the checkout had none, per-file rels when merging.
- **`/_bmad/render/` always excluded** (git-common-dir `info/exclude`), so session-rendered output (machine-absolute paths baked in) can't be swept into story commits by the unit's `git add -A`; `bmad-loop init` also gitignores `_bmad/render/` for in-place (`isolation = "none"`) runs.
- **Renderer preflight** (`install.missing_renderer_support`) in `validate` and at run/sweep/resume start: `render.py` next to the shim, `uv` on PATH, `_bmad/config.toml` present. Gated on the installed SKILL.md actually containing `render.py` — pre-render installs get zero new checks and byte-stable validate output.
- **`validate --render-probe`** (opt-in): executes the real `uv run render.py` against the project; pass = the `read and follow …` dispatch line. Surfaces #2588 render-time HALTs before a run. Side effect documented in `--help` (rewrites the regenerated `_bmad/render/bmad-dev-auto/` cache).
- **Two validate warnings (never FAILs)**: four-layer-TOML vs legacy `bmm/config.yaml` desync on `implementation_artifacts` (`bmadconfig.toml_artifact_parity`, mirroring render.py's exact merge precedence — deep-merge layers, then `[modules.bmm]` beats `[core]`; TOML-first migration tracked in #154), and a one-time "you have `_bmad/render/**` committed — `git rm -r --cached` it" hint (excludes can't shield tracked files).
- **Profiles confirmed unchanged**: claude's native slash invocation and copilot's "LOAD the FULL SKILL.md … follow its directions exactly" both work against the two-line shim; the only failure mode (no uv in-session) is the preflight's job.

## Verification

- Full suite **2295 passed** (+23 new tests), `trunk check` clean. New coverage: seed/merge/shield/symlink-escape, renderer-support matrix (incl. uv-missing reported once across trees), probe pass/HALT via a fake-uv launcher (no test executes real uv — Windows CI safe), an end-to-end engine test proving a committed-`_bmad` project's worktree gets the gitignored user layer seeded **and** the session's render residue never reaches the story commit, and TOML-parity precedence tests.
- **Real smoke** on this repo with the actual upstream skill resynced to 1cd4a7f5: `validate` detects the shim and passes the preflight; `validate --render-probe` ran the real renderer via uv and produced its dispatch line; the rendered step-04 confirms the absolute-path baking the seed protects against.

## Follow-ups

- #154 — bmadconfig: read the four-layer TOML as primary, yaml fallback (retires the parity warning).
- Upstream (BMAD-METHOD): installer should gitignore `_bmad/render/`; render.py's walk-up could stop at a git-worktree/repo boundary as defense-in-depth. Filed separately.

## Summary by CodeRabbit

- **New Features**
  - Added renderer-era preflight checks for `render.py`, `uv`, and centralized configuration.
  - Added `validate --render-probe` to detect render-time failures and configuration issues.
  - Improved isolated worktree setup by seeding required configuration while excluding generated render output.

- **Bug Fixes**
  - Prevented renderer artifacts and user configuration layers from being included in commits.
  - Added warnings for configuration mismatches and already-tracked render output.

- **Documentation**
  - Updated setup, feature, contribution, and changelog documentation with renderer validation and worktree behavior.


pbean:
Closing this draft rather than rebasing it. #447 now carries the renderer preflight that belongs on current `main`.

The remaining concerns are split across their current owners:

- #407 owns renderer execution probing using the installed stub's real argv.
- #408 owns a corrected TOML-parity model.
- #409 completed the tracked-render warning.
- #433 tracks Phase 6H `_bmad` seeding and worktree shielding.

The implementation here is not a viable base for those follow-ups: its `render.py` marker never matches the current `render_skill.py`; its execution-probe argv is guessed rather than derived from the installed stub; its parity model is obsolete; and its provisioning changes target the old `install.py`/shared-exclude seams that current `main` replaced with `worktree_flow.py` and worktree-scoped shielding.

Issue #156, opened 2026-07-16
Uncaught GIT_TIMEOUT in verify.attempt_dirty crashes the whole run from the rollback/pause path


ferrandmathis-afk:
**Version:** bmad-loop 0.8.1 (uv tool, git install) · macOS 15 (Darwin 24.6.0) · git worktree isolation, `adapter.name = "claude"`
**Component:** `verify.py:26` / `verify.py:78-82` (`_git`, `GIT_TIMEOUT_S`) reached from `engine.py:834` (`_rollback_or_pause`)

## Summary

`verify._git` runs every git call with a hard `timeout=GIT_TIMEOUT_S` (120s, a module constant — neither policy- nor env-configurable). `subprocess.TimeoutExpired` is not caught anywhere on the way up, so one slow git call takes down the entire run.

We hit it on the **failure-recovery path**, which is what makes it worth reporting: the dev session had already timed out, `_dev_phase` called `_rollback_or_pause` to handle that gracefully, and the recovery itself killed the run.

```
File "bmad_loop/engine.py", line 1511, in _dev_phase
    self._rollback_or_pause(task)
File "bmad_loop/engine.py", line 834, in _rollback_or_pause
    if task.baseline_commit and not verify.attempt_dirty(
File "bmad_loop/verify.py", line 181, in attempt_dirty
    rc, _ = _git(repo, "diff", "--quiet", baseline, "--", ".", *_exclude_specs(exclude))
File "bmad_loop/verify.py", line 78, in _git
    proc = subprocess.run(
subprocess.TimeoutExpired: Command '['git', '-C', '<worktree>', 'diff', '--quiet',
'<baseline>', '--', '.', ':(exclude).bmad-loop/policy.toml']' timed out after 120 seconds
```

Net effect: `0 done, 0 deferred, 0 escalated` and a `crashed: true` state.json. The story's work was intact in the worktree the whole time — only the orchestrator died.

## Why the git call was slow

The host was under heavy memory pressure (the story had been standing up Docker/Postgres containers). On an unloaded machine the *same* `git diff --quiet` against the *same* worktree completes in ~20s. So 120s is a sane normal-case bound; the defect is that exceeding it is run-fatal rather than a handled outcome.

## What I expected

[...]

PR #162, opened 2026-07-16
fix: accept ancestor baselines for dw bundles; auto-recover unit worktrees on retry (#161)

### LINKED ISSUE #161: Sweep bundle that routes into an existing done spec always fails the baseline gate; recovery message then gives wrong (in-place) instructions in worktree mode
## Environment
- bmad-loop 0.8.1 (uv tool install from git), macOS, tmux 3.7b
- `scm.isolation = "worktree"`, `branch_per = "story"`, `rollback_on_failure = false`
- adapter: dev/triage = claude (opus), review = codex
- upstream skills: bundled `bmad-dev-auto` + review hunters

## Bug 1: baseline gate structurally rejects follow-up-review bundles

A per-epic sweep triaged a deferred-work entry (DW-1: "independent follow-up review of stories 1.1–1.3, cut short by the damping cap") into a one-entry bundle. The dev session completed the work correctly — but the post-session gate rejected the attempt:

```
dev-decision: action=retry,
reason: spec baseline 800adc7deebb does not match orchestrator-recorded baseline 393b55a052dc
```

Sequence (from journal):
1. Sweep worktree opened at baseline `393b55a` (main HEAD after the sweep's own ledger-migrate commit).
2. Bundle intent = follow-up review of already-done stories. `bmad-dev-auto` step-01 routes an invocation pointing at a `status: done` spec into a fresh review pass **on that spec** (`done → step-04`).
3. The session therefore adopted story 1.1's spec as `spec_file` — whose `baseline_revision` is `800adc7` (the story's original dev baseline, and the *correct* diff base for reviewing the stories).
4. `_verify_shared_gates` (verify.py) compares the spec's `baseline_revision` against `task.baseline_commit` (`393b55a`) → mismatch → whole attempt failed as `retry`.

The two rules are individually correct but structurally collide: **any** sweep bundle whose intent is "follow-up review of story X" routes into a spec whose baseline legitimately predates the sweep worktree, so it does ~8M tokens of work and then always fails. The review work itself was sound (we inspected, tested, and manually merged the branch).

Possible fixes: `verify_dev_bundle` could tolerate/skip the baseline gate when the adopted spec is a pre-existing done spec, or the bundle prompt could instruct the session to create a fresh spec for the bundle (our local workaround via `_bmad/custom/bmad-dev-auto.toml` `persistent_facts` does the latter).

## Bug 2: manual-recovery pause message assumes in-place mode under worktree isolation

The retry decision + committed work + `rollback_on_failure=false` produced a `rollback-manual-required` pause whose message told us:

> Your commits are intact at the current HEAD. … Save them first — e.g. `git branch my-rescue HEAD` (the commits are `393b55a052dc..HEAD`). … Only if you decide to discard: `git reset --hard 393b55a052dc` …

Under worktree isolation this is wrong on every point: the main checkout's HEAD **was** `393b55a` (the quoted range was empty, nothing to rescue there), and the attempt's 2 commits lived on the still-mounted unit branch `bmad-loop/<run>/<story>` instead. Following the instructions literally does nothing; a hasty read of step 3 invites a destructive `git reset` on a checkout that was never touched. The message should branch on `scm.isolation` and, in worktree mode, point at the unit branch/worktree path.

## Also noting
`bmad-loop diagnose <run>` refuses to emit for this run (`FAIL: refusing to emit — leak self-check fired: sensitive[2]`, also with `--max-journal-entries 0`), so no sanitized dump is attached; happy to extract specific journal entries by hand on request.

otiunov:
Fixes #161 (both bugs).

## What

**1. `verify_dev_bundle` accepts ancestor baselines.** A deferred-work bundle whose intent is "follow-up review of story X" is routed by `bmad-dev-auto` (step-01, `done → step-04`) into that story's existing done spec — whose `baseline_revision` is the story's original dev baseline, necessarily older than the unit worktree cut for the bundle. The exact-match baseline gate therefore failed every such bundle *after* the session had done all its work (~8M tokens in the incident run). The gate now tolerates a claimed baseline that is an **ancestor** of the orchestrator-recorded one, bundle path only:

- ancestor = the session diffed from an earlier commit on the unit's own history — a superset of the unit's changes, sound for a review pass;
- diverged/unknown baselines still fail;
- sprint mode and stories mode keep the exact-match requirement (a spec authored this session must match);
- new `verify.is_ancestor()` reads any git failure as False, so uncertainty keeps the gate strict.

**2. `_rollback_or_pause` auto-recovers mounted unit worktrees.** The mid-drive dev retry (`_dev_phase`) called `_rollback_or_pause` unconditionally — the only recovery path without an isolation guard (the resume, `_defer`, and sweep paths all have one). With `rollback_on_failure = false` this paused the run with in-place manual-recovery instructions: "your commits are intact at the current HEAD … the commits are `<baseline>..HEAD` … `git reset --hard <baseline>`". Under worktree isolation every clause pointed at the wrong tree — the operator's checkout HEAD *is* the baseline (the quoted range is empty) and the commits sit on the unit branch, so step 3 invites a destructive reset of a checkout the attempt never touched. Since a mounted unit worktree is disposable by design (commits are parked on `attempt-preserve/` refs before any reset, per `_preserve_attempt_commits`), the method now auto-recovers it regardless of `rollback_on_failure`, which the policy docs already scope to `isolation="none"` recovery. Detection is by path (`workspace.root != paths.repo_root`), not by policy — a worktree-mode call that reaches the method *outside* a mounted unit still targets the main checkout and still pauses.

**3. Recovery notices name their tree.** The remaining reachable pauses (e.g. preserve-failure while a worktree is mounted) now say "current HEAD of `<root>`" and use `git -C <root> …` commands, so they can't be misread against the wrong checkout.

## Incident evidence

Run `20260716-152154-3aba` (v0.8.1, `isolation="worktree"`): sweep bundle DW-1 completed a follow-up review of three stories, then failed `spec baseline 800adc7deebb does not match orchestrator-recorded baseline 393b55a052dc` — `800adc7` is an ancestor of `393b55a` in that repo, confirmed against the new gate. Full journal excerpts in #161.

## Tests

- `test_verify_dev_bundle_ancestor_baseline_passes` — the incident shape passes
- `test_verify_dev_bundle_foreign_baseline_still_fails` — relaxation is ancestor-only
- `test_verify_dev_sprint_ancestor_baseline_still_fails` — relaxation is bundle-only
- `test_dev_retry_in_worktree_auto_recovers_instead_of_pausing` — e2e: committed bad attempt in a worktree → auto-recover (commits parked on `attempt-preserve/`), retry succeeds, run completes; no `rollback-manual-required`
- `test_manual_recovery_notice_names_committed_work` updated for the `git -C <root>` command form (ordering assertion unchanged)

`uv run pytest -q`: 2270 passed, 8 skipped. `ruff check` + `black --check` clean on changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic rollback behavior for failed worktree attempts.
  * Clarified manual recovery instructions, including safer commit preservation and reset steps.
  * Allowed deferred development bundles with valid ancestor baselines to pass verification.
  * Continued rejecting unrelated or invalid baselines and enforcing exact baselines for sprint verification.

* **Tests**
  * Added coverage for worktree rollback isolation and baseline verification scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -170,6 +170,25 @@ PATH)`, the TUI notifies `multiplexer backend unavailable — launch/attach disa
 
 ### Fixed
 
+- **Deferred-work bundles that adopt an existing story spec pass the baseline gate (#161).** A
+  "follow-up review of story X" bundle is routed by `bmad-dev-auto` into that story's done
+  spec, whose `baseline_revision` is the story's original dev baseline — necessarily older
+  than the bundle's worktree cut, so the exact-match gate failed every such bundle after the
+  session had already done its work. The bundle gate now accepts a claimed baseline that is
+  an _ancestor_ of the orchestrator-recorded one (the session diffed a superset of the
+  unit's changes); diverged or unknown baselines still fail, any git fault in the probe
+  reads as not-an-ancestor, and sprint/stories modes keep the exact-match requirement.
+
+- **A failed attempt inside a unit worktree auto-recovers instead of pausing with in-place
+  instructions (#161).** The mid-drive dev retry was the only recovery path without an
+  isolation guard: with `rollback_on_failure = false` it paused the run with manual-recovery
+  instructions aimed at the operator's checkout — whose HEAD _is_ the baseline under
+  worktree isolation, while the commits sat on the unit branch, so following them literally
+  did nothing and invited a destructive reset of a tree the attempt never touched. A mounted
+  unit worktree is disposable: the attempt's commits are parked on `attempt-preserve/` refs
+  and the worktree resets regardless of the flag, which gates in-place (`isolation = "none"`)
+  recovery only. The remaining reachable pauses name their tree (`git -C "<root>" …`).
+
 - **A failed worktree teardown no longer crashes the run after the merge landed (#139).** When a
   process the just-ended session left running (e.g. pytest recreating `.pytest_cache`) makes
   `git worktree remove` fail with ENOTEMPTY, git still drops its admin entry, so the `force=True`
--- a/src/bmad_loop/engine.py
+++ b/src/bmad_loop/engine.py
@@ -872,7 +872,16 @@ class Engine:
         if task.baseline_commit and not dirty:
             self.journal.append("rollback-skipped-clean", story_key=task.story_key)
             return
-        if resolved or self.policy.scm.rollback_on_failure:
+        # A mounted unit worktree is disposable by design: its branch never
+        # touches the operator's checkout and the attempt's work is parked on
+        # recovery refs before any reset. `scm.rollback_on_failure` gates
+        # *in-place* (isolation="none") recovery only — pausing here would emit
+        # main-checkout reset instructions for a tree the operator never works
+        # in (#161). Compared by path, not by policy: a worktree-mode call that
+        # reaches this method *outside* a mounted unit (e.g. resume with no
+        # worktree recorded) still targets the main checkout and must pause.
+        in_unit_worktree = self.workspace.root != self.paths.repo_root
+        if resolved or in_unit_worktree or self.policy.scm.rollback_on_failure:
             self.journal.append(
                 "rollback-auto",
                 story_key=task.story_key,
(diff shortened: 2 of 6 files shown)


pbean (approved):
Reviewed against a full code-trace of both bugs — the diagnosis in #161 is exactly right, and both fixes take the approach I'd have taken:

- **Ancestor-only, bundle-only** is the correct shape for the gate relaxation. The sprint/stories exact-match stays sound because resolve re-drives re-stamp `baseline_revision` (runs.py), and a bundle session can only end up recording an ancestor by adopting a pre-existing spec — so the relaxation is effectively scoped to exactly the adoption case.
- **Auto-recover over message-branching** is right for Bug 2: the mounted unit worktree is disposable by design, the policy doc already scopes `rollback_on_failure` to in-place recovery, and your path-based detection is valid at every call site (`self.workspace` is only swapped while a unit is mounted; the resume/defer paths keep their guards).

I rebased onto main and pushed one review commit (8eb74da) on top:

1. **`is_ancestor` now catches `GitError`** — the one real issue found. Since #156 (landed after your branch point), `_run_git` translates `subprocess.TimeoutExpired` into `GitError`, so the original `except (OSError, TimeoutExpired)` would have let a git timeout propagate out of `_verify_shared_gates` and crash the run instead of keeping the gate strict. Added `test_verify_dev_bundle_ancestor_probe_git_failure_stays_strict` pinning it at the gate level.
2. `_rollback_or_pause`'s docstring updated — it still described the old "OFF leaves the tree untouched" invariant; docstrings that justify invariants are load-bearing here.
3. CodeRabbit's nitpick applied: the notice commands quote the path (`git -C "<root>" …`).
4. CHANGELOG entries added (repo convention for fix PRs).

Verified: your four new tests + the updated one fail/pass exactly as claimed with the fix reverted (the worktree e2e reproduces the incident pause verbatim pre-fix); full suite 2479 passed, trunk clean, CI green. The diagnose refusal from your "Also noting" is now tracked as #186.

Thanks for an unusually well-diagnosed issue *and* the fix to go with it.

Issue #169, opened 2026-07-17
antigravity: worktree isolation unusable — agy trust is exact-path, so every worktree hangs on the trust dialog


pbean:
### Problem

`antigravity` (Google's `agy`) gates every workspace on an interactive trust dialog:

```
Do you trust the contents of this project?
> Yes, I trust this folder
  No, exit
```

Spawned sessions cannot answer it, so a pending dialog reads as a session timeout.

### Verified trust semantics (agy 1.1.3, empirical 3-run matrix)

Launch shape matched the real adapter (`agy -i "<prompt>" --dangerously-skip-permissions` in tmux):

| Run | `trustedWorkspaces` contains | Result |
| --- | --- | --- |
| A (control) | neither `/tmp` nor the dir | prompts |
| B (prefix) | `/tmp` | **still prompts** |
| C (exact) | the dir itself | no prompt, turn starts |

Conclusions:

- `~/.gemini/antigravity-cli/settings.json` → `trustedWorkspaces` **is** the gate.
- Matching is **exact-path**. Trusting a parent does **not** cover subdirectories.
- `--dangerously-skip-permissions` does **not** bypass it (it covers tool permissions only).
- No trust-bypass env var exists (searched `AGY_*` / `ANTIGRAVITY_*` / `GEMINI_*` in the binary).

### Impact

- `isolation = "none"` (the default) works — the project root is trusted once, by hand.
- **`isolation = "worktree"` does not work.** Worktrees live at `<project>/.bmad-loop/runs/<run_id>/worktrees/<unit>` — a fresh, untrusted path per run, so every session hangs on the dialog until the stall timeout.
- `probe-adapter antigravity --probe` cannot self-serve either: it launches in a fresh `mkdtemp` that by definition can't be pre-trusted.

Documented as a limitation in the profile's `first_run_note` and the setup guide (see the antigravity finishing work). This issue tracks lifting it.

### Proposal (opt-in, default off)

An opt-in knob (e.g. `[adapter] antigravity_seed_worktree_trust = false`) that appends a worktree path to `trustedWorkspaces` on creation, with:

- **Containment guard** — only ever add a path whose enclosing project root is *already* in `trustedWorkspaces`. The user has already granted trust to that repo; a worktree is the same repo at a path inside it, so this grants no new authority. Never seed for an untrusted root.
[...]

Issue #159, opened 2026-07-16
[BUG] bmad-loop validate and probe fail to recognize registered hooks for antigravity adapter


Nips02:
### Description

When running `bmad-loop validate` or `bmad-loop probe` with the adapter profile set to antigravity (using the `antigravity-hooks-json` dialect), the validation fails with:` FAIL: bmad-loop hooks not registered for antigravity — run bmad-loop init --cli antigravity`

Even if `bmad-loop init --cli antigravity` was successfully run and the hook is present in [.agents/hooks.json](), validation still fails.

### Root Cause

In `bmad-loop`, `the antigravity-hooks-json` dialect keys its configuration by the hook name group` ("bmad-loop")` at the top level instead of a `"hooks"` key. This is correctly handled during the installation process in `install.py` (which writes to the `"bmad-loop"` group):

```

{
  "bmad-loop": {
    "Stop": [
      {
        "type": "command",
        "command": "python3 '/path/to/project/.bmad-loop/bmad_loop_hook.py' Stop",
        "timeout": 60
      }
    ]
  }
```
}
However, both the validate CLI command and the probe utility check for hooks by querying` doc.get("hooks", {}) `which returns an empty dict for antigravity config files.

`validate command location: bmad_loop/cli.py (lines 292–297)`
`probe verification location: bmad_loop/probe.py (lines 341–344)`

### Steps to reproduce

1. In policy.toml, set the default adapter to antigravity:

```
[
adapter
]
name = "antigravity"
```
2. Initialize hooks for the project:

`
bmad-loop init --project . --cli antigravity

`
3. Run the validation command:

`bmad-loop validate`
4. Observe the failure: 
`FAIL: bmad-loop hooks not registered for antigravity`

### Expected behavior

The validation checks should correctly detect that the hook exists under the `"bmad-loop"` group when checking the `antigravity-hooks-json` dialect.

### Proposed Fix
Update the hook-retrieval logic in [bmad_loop/cli.py](url) and [bmad_loop/probe.py](url) to query the correct dictionary key based on the active profile hooks dialect.

`Diff for bmad_loop/cli.py`

```
@@ -291,7 +291,11 @@
         hooks_ok = False
         if hook_config.is_file():
             try:
-                hooks = json.loads(hook_config.read_text(encoding="utf-8")).get("hooks", {})
+                data = json.loads(hook_config.read_text(encoding="utf-8"))
+                if profile.hooks.dialect == "antigravity-hooks-json":
[...]

PR #171, opened 2026-07-17
fix(antigravity): correct payload casing + transcript path; pin trust and usage facts

pbean:
## What

Finalizes the `antigravity` (`agy`) profile against a **live agy 1.1.3 capture**, replacing the guesses it shipped with: correct hook-payload casing, the correct transcript path, and pinned answers to the two questions the profile had left open (worktree trust, and whether a `usage_parser` is possible).

> **Stacked on #170** (the #159 fix) — review that first; this PR's base will retarget to `main` once it merges.

## Why

The profile shipped as "experimental, probe-required" with two placeholders: `usage_parser = "none"` ("deferred") and a `first_run_note` saying *"worktree runs use a different path than the trusted one — verify during probe whether that re-triggers a trust prompt."* `agy` is now available to probe, so both are answerable — and the probe found three things no doc would have revealed.

## How

- **Payload casing.** agy encodes hook payloads as protojson (`conversationId`, `transcriptPath`, `workspacePaths`), but the relay tried only snake_case plus copilot's `sessionId` — so **every agy event recorded a null `session_id`**, and `cwd` was never populated (agy sends no `cwd`, only a `workspacePaths` list). Both the relay and the probe capture hook now try agy's casing.
- **Transcript path.** The shipped glob was community-sourced with the wrong filename: agy writes `transcript_full.jsonl`, not `transcript.jsonl`. ⚠️ Note agy's own `hooks.md` documents a *workspace-relative* `<ws>/.gemini/antigravity/transcript.jsonl` — that is **illustrative**; the live payload's `transcriptPath` is home-rooted under `brain/<conversationId>/`. Corrected from the capture, not the doc.
- **`--probe` prefers the payload's path over the convention glob.** The payload names *this* turn's file; a glob can only take the newest match and may land on an unrelated session. Generic, not agy-specific.
- **`usage_parser = "none"` is permanent, and the profile now says why.** A live turn's transcript carries only `step_index`/`source`/`type`/`status`/`created_at`/`content`/`thinking` — scanning every key recursively finds no usage block, and the probe's own `token_field_candidates` comes back empty. agy *does* count tokens (its TUI shows them) but only in `conversations/<id>.db`, whose `gen_metadata` is an undocumented protobuf blob — outside the `(transcript_path) -> TokenUsage` contract and fragile against an internal schema. Runs work; token columns stay empty.
- **Trust is exact-path**, so `isolation = "worktree"` is unsupported (see below). Documented in the profile, README, FEATURES, and setup guide; lifting it is tracked in #169.
- **Authoring guide**: recorded the two transferable lessons — `none` is a legitimate *terminal* answer for a CLI that doesn't expose usage, and a captured payload beats a documented path.

### Trust matrix (agy 1.1.3, at the adapter's real launch shape)

`agy -i "<prompt>" --dangerously-skip-permissions` in tmux:

| Run | `trustedWorkspaces` contains | Result |
| --- | --- | --- |
| A (control) | neither `/tmp` nor the dir | prompts |
| B (prefix) | `/tmp` | **still prompts** |
| C (exact) | the dir itself | no prompt, turn starts |

So: `trustedWorkspaces` is the gate, matching is **exact-path** (a trusted parent does not cover subdirectories), and `--dangerously-skip-permissions` does **not** bypass it (tool permissions only). ⇒ `isolation = "none"` (the default) works; **`isolation = "worktree"` hangs**, since each worktree is a fresh untrusted path under `.bmad-loop/runs/`.

## Testing

- Live `agy 1.1.3` capture: Stop hook fired; the payload confirms `conversationId` / `transcriptPath` / `workspacePaths` and resolves all three relay fields (previously null). `terminationReason` came back `NO_TOOL_CALL`, not the docs' `model_stop`.
- Corrected glob verified end-to-end against the real transcript agy wrote — found it, scanned 4 entries, `token_field_candidates: []`.
- New `test_infer_antigravity_transcript_has_no_token_fields` pins the captured shape, so `usage_parser = "none"` doesn't get "fixed" blindly; it will start failing if agy ever emits usage.
- New relay tests for agy's casing, incl. an empty `workspacePaths` degrading to `None` rather than `IndexError`.
- Full suite green (2328 passed, 1 skipped); `trunk check` clean. No changes to the probe's temp-dir teardown or sanitization.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Antigravity hook compatibility with varying payload formats.
  * Fixed transcript discovery during Antigravity probing, including support for CLI-reported paths.
  * Prevented incorrect “hooks not registered” reports.
  * Clarified workspace trust behavior and isolation limitations.

* **Documentation**
  * Updated Antigravity requirements to `agy >= 1.1.3`.
  * Documented that `isolation = "none"` is required and token usage remains unavailable.
  * Added guidance for trust dialogs, transcript paths, and adapter verification.


pbean:
@coderabbit review

pbean:
@coderabbit review

PR #173, opened 2026-07-17
fix: keep committed work across a dev retry; accept the baseline it advances past (#172)

### LINKED ISSUE #172: Dev retry after a timeout rolls back commits the attempt already landed, then re-derives them
## Environment

- bmad-loop 0.8.1 (uv tool install), Linux (WSL2), tmux
- `scm.isolation = "none"`, `branch_per = "story"`, `rollback_on_failure = true`
- adapter: claude (opus), `limits.session_timeout_min = 90..180`, `max_dev_attempts = 2`

## The bug

A dev session that ends with a non-completed status (timeout/stalled/crashed) never reaches
`_verify_dev_artifacts`, so its RETRY decision falls straight to `_rollback_or_pause`, which
hard-resets the tree to the story's baseline — **even when the session had already committed
real, working progress before the wall clock cut it off**.

The retry then re-derives the same implementation from scratch, burning a large multiple of
the tokens, and frequently hits the same timeout again before the attempt budget runs out.
The story is deferred despite having been essentially done at the first timeout.

## Steps to reproduce

1. Give a story enough work that dev-1 cannot finish inside `session_timeout_min`.
2. Let dev-1 commit incrementally (any real work above `baseline_commit`).
3. Let the session hit the wall-clock timeout.
4. Observe `rollback-auto` in the journal: dev-1's commits are parked on an
   `attempt-preserve/` ref and the tree is reset to baseline.
5. dev-2 starts from a plain `/bmad-dev-auto <key>` prompt with no knowledge of the work
   that existed, and re-derives it.

## Expected vs actual

- **Expected:** a same-story retry after a non-completed session keeps commits the attempt
  already landed and tells the resumed session to build on them.
- **Actual:** the commits are rolled back and silently re-derived.

## Evidence

md2pdf-auto run `20260709-004907-923b`, story 17-3: dev-1 committed a working table-theme
implementation 9 minutes before its 90-minute timeout. The rollback discarded it; dev-2 spent
~3x the tokens re-deriving the same change and ALSO timed out, exhausting the 2-attempt budget
and deferring a story that was already essentially complete.

## Note on the fix

The obvious fix — keep the tree and advance `baseline_commit` past the kept commits, mirroring
`runs.rearm_escalation` — has a second-order trap worth naming here, because it cost a further
83.6M tokens on md2pdf-auto run `20260716-153340-9183` (story 18.5) while it was carried as a
local patch:

Advancing the baseline splits it from the spec's `baseline_revision`, which the timed-out
attempt already stamped and which nothing re-stamps (correctly — that sha is the diff base the
whole story's review must span). `_verify_shared_gates` then fails the *finished* story on
`spec baseline 68eda05 does not match orchestrator-recorded baseline 1ead148`, and with
`resolved_redrive` latched it re-escalates rather than defers.

Worth noting that the existing engine fixtures cannot catch this: `dev_effect` re-stamps the
spec with the current HEAD on every session, which no resumed production session does (a resume
keeps the attempt's spec — that is the point of it). Same masking as #89.

PR incoming, which fixes both halves.

Pawel-N-pl:
## What

A same-story dev retry after a non-completed session (timeout/stalled/crashed) now keeps
commits the attempt already landed, and the baseline gate accepts the spec baseline that
keeping them leaves behind.

## Why

Fixes #172.

A timed-out session's commits were rolled back and re-derived from scratch by the retry
(~3x tokens, usually into a second timeout). Keeping them requires advancing
`baseline_commit` past them — which splits it from the spec's `baseline_revision`, stamped
by the timed-out attempt and deliberately not re-stamped (it is the diff base the story's
whole review must span). The exact-match gate then failed the *finished* story on that
split alone: 83.6M tokens, re-escalated instead of committed (evidence in #172).

## How

- Keep the tree when `commits_above(baseline)` is non-empty; feedback + prompt tell the
  resumed session its predecessor ran out of time, not that its work was wrong.
- `_advance_baseline_to_head` adopts those commits and latches `baseline_advanced`.
- The gate accepts an **ancestor** baseline only while that latch is set; diverged shas
  still fail, and every other path keeps exact match. `rearm_escalation` clears it.

Note: `verify.is_ancestor()` overlaps #162 — happy to rebase onto whichever lands first.

## Testing

`uv run pytest -q`: 2330 passed, 4 failed (`test_module_skills_sync`, pre-existing on main
in this checkout), 6 skipped. black/ruff clean on changed files. Each new gate test was
re-run with the relaxation disabled to confirm it fails without the fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved development retries after interrupted sessions.
  - Preserves committed work when a session ends after making progress, avoiding unnecessary rollback or duplicated work.
  - Automatically rolls back attempts that produced no durable progress.
  - Resumed sessions now receive clearer instructions to continue existing work.
  - Improved validation for previously recorded progress while preventing invalid or unrelated baselines.
  - Reset escalation state correctly when a task is reactivated.

### DIFF
--- a/src/bmad_loop/engine.py
+++ b/src/bmad_loop/engine.py
@@ -2503,13 +2503,24 @@ class Engine:
         assigned, so a failure partway through can't advance one and not the
         other. Best-effort: on a git failure the old baseline stands — the
         resumed session's verify_dev may then flag a baseline mismatch, which is
-        no worse than today's unconditional rollback."""
+        no worse than today's unconditional rollback.
+
+        Latches ``baseline_advanced`` on success: the kept attempt already
+        stamped the spec's ``baseline_revision`` with the PRE-advance sha and
+        nothing re-stamps it (deliberately — that sha is the diff base the whole
+        story's review must span, so narrowing it to the advanced one would hide
+        the kept attempt's own work from the review). Only the flag tells verify's
+        baseline gate that the resulting spec/orchestrator split is one the
+        orchestrator caused, so it can accept an ancestor there and nowhere
+        else. Set last, and only if both git reads landed: an un-advanced
+        baseline must not claim to have been advanced."""
         try:
             repo = self.workspace.root
             head = verify.rev_parse_head(repo)
             untracked = sorted(verify.untracked_files(repo))
             task.baseline_commit = head
             task.baseline_untracked = untracked
+            task.baseline_advanced = True
         except Exception:  # noqa: BLE001  # nosec B110 - best-effort git read, must not fail retry
             pass
 
--- a/src/bmad_loop/engine.py
+++ b/src/bmad_loop/engine.py
@@ -1510,6 +1510,7 @@ class Engine:
             # session's own tree.
             task.baseline_untracked = sorted(verify.untracked_files(self.workspace.root))
         feedback: Path | None = None
+        resume_note: str = ""
         while True:
             if resume_result is None:
                 # a resumed result replays the attempt it was recorded under, so
@@ -1538,7 +1539,7 @@ class Engine:
                 result = self._run_session(
                     task,
                     role="dev",
-                    prompt=self._dev_prompt(task, feedback),
+                    prompt=self._dev_prompt(task, feedback, resume_note=resume_note),
                     seq=task.attempt,
                 )
             advance(task, Phase.DEV_VERIFY)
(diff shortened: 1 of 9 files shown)


Pawel-N-pl:
Closing — filed prematurely, moving this back to local review before it takes reviewer time. Will re-open once vetted.

Issue #160, opened 2026-07-16
[BUG] The #109 stall-grace fix cannot reach the review leg: a review pass is killed on its first Stop by the dev pass's stale `## Auto Run Result`


ferrandmathis-afk:
# [BUG] The #109 stall-grace fix cannot reach the review leg: a review pass is killed on its first Stop by the *dev* pass's stale `## Auto Run Result`

**Version:** bmad-loop 0.8.1 · claude adapter · `scm.isolation = "worktree"` · `[review] enabled = true, trigger = "always"`

## Summary

A review pass that delegates to sub-agents and yields its turn is killed on its **first** Stop. `dev_stall_grace_s` / `dev_stall_nudges` — the guardrails added for exactly this scenario in #109 — can never arm on the review leg, because the completion marker they key on is already present: the **dev** pass wrote it into the same spec.

The result is a review loop that cannot converge. It burns `max_review_cycles`, then defers a story whose work was complete and verify-green.

## Root cause

`adapters/generic.py:305` — a synthesizable result short-circuits **before** the nudge and the grace window:

```python
if event.event == "Stop":
    result_json = self._result_json(handle, spec, wait=True)
    if result_json is not None:
        return SessionResult(status="completed", result_json=result_json, ...)  # ← returns here
    if nudges_left > 0:                                    # nudge
        ...
    stall_deadline = time.monotonic() + self._stall_grace_s   # #109's grace window
```

The completion marker is the `## Auto Run Result` section (`devcontract.AUTO_RUN_HEADING_RE`). The asymmetry between the two legs:

- **Dev leg:** early Stops happen *before* the section exists → `result_json is None` → nudge + grace arm → the session survives its sub-agents and finalizes.
- **Review leg:** the review pass re-invokes `bmad-dev-auto` on the **same spec**, where the dev pass already wrote `## Auto Run Result` / `Status: done`. So the very first Stop synthesizes a result → the session is ended instantly, mid-flight.

A review pass *always* yields at least once — `bmad-code-review` fans out to sub-agent hunters. So `step-04-review.md` can never reach its `Finalize` (line 95). The spec is left at the transient `status: in-review` that `step-04-review.md:15` sets on entry, the engine reads a non-terminal status, and loops.

This is the same theme as #109 and #48 ("a result observed while work is still alive is advisory — keep waiting"), on a third path: **the marker is stale rather than mid-write**, so no amount of waiting helps — it was already there when the session started.

[...]

PR #179, opened 2026-07-18
fix(tui): pending journal jump survives a starved RichLog flush (#178)

### LINKED ISSUE #178: TUI: direct-path journal jump is lost forever when the RichLog flush outlives the 3s scroll retry budget
## Symptom

\`tests/test_tui_app.py::test_journal_enter_jumps_to_log_position\` flakes on Windows CI (latest: run 29626618672 on PR #160, windows py3.11): \`0 < log.scroll_y < log.max_scroll_y\` never becomes true, and **all three attempts** (initial + 2 rerunfailures reruns) burn the full 10s \`until\` timeout identically.

## Root cause

The test exercises the **direct** jump path in \`_jump_to_log_event\` (dashboard.py:368-370): a same-task jump calls \`_scroll_log_to\` inline and never sets \`self._pending_jump\`, so the poll-tick retry machinery cannot rescue it. Its only completion mechanism is \`_scroll_log_to\`'s internal \`set_timer\` chain — capped at 60 × 0.05s = **3s** (dashboard.py:378-393). The \`#log\` RichLog was hidden in the inactive TabPane, so its writes are deferred; if the post-tab-switch flush outlives the 3s budget (starved xdist Windows runner), the chain exhausts and **silently drops the jump forever** (line 391). Reruns re-roll on the same contended runner and lose the same way.

The deferred path has the same hole after handoff: \`_apply\` (842-849) clears \`_pending_jump\` **before** the scroll lands, handing completion to the same bounded timer.

## History

Patched 5 times without touching the root: b15feec (budget 1s→3s), dd37554 (flaky markers + until 5s→10s), PR #108 (two test-side de-flakes). The fire-once/bounded-timer design never changed.

Not upstream: Textual 8.2.8 (2026-06-30) contains nothing scroll-related; Textualize/textual#6311 is a different symptom (auto-scroll yank on write).

## Fix

Make \`_pending_jump\` the single owner of the jump: set on both paths, cleared only when the scroll actually lands (or the log is provably empty), retried by the existing ≤1s poll tick; an identity guard on the timer chain kills stale/superseded chains so exactly one scroll lands per jump. Then the flaky markers can be removed.

pbean:
## Problem

`test_journal_enter_jumps_to_log_position` kept flaking on Windows CI (latest: run 29626618672 on PR #160) — all three attempts (initial + 2 rerunfailures reruns) burned the full 10s wait identically. Root cause (#178): a same-task journal jump never set `_pending_jump`, so its only completion was `_scroll_log_to`'s 60×0.05s timer chain. The `#log` RichLog sits hidden in the inactive TabPane with its writes deferred; when the post-tab-switch flush outlived the 3s budget on a starved runner, the chain exhausted and **silently dropped the jump forever**. The deferred (cross-task) path had the same hole after handoff: `_apply` cleared `_pending_jump` before the scroll landed. Five prior patches (budget 1s→3s, rerun markers, PR #108's test-side de-flakes) widened or masked the window without removing it. Not upstream: Textual 8.2.8's changes are unrelated, and Textualize/textual#6311 is a different symptom.

## Fix — single-owner pending jump

- `_pending_jump` is set on **both** jump paths and cleared only by `_scroll_log_to` when the scroll actually lands (or the log is provably empty — cleared before the one-shot notify so per-tick retries can't spam it).
- `_apply` no longer pre-clears; its pending block re-attempts the scroll on every ≤1s poll tick, so a flush that outlives the chain delays the jump instead of losing it.
- The timer chain (now 20×0.05s = 1s) covers sub-second snappiness only. Each fire re-checks the jump **identity** (`is`) and the displayed task, so superseded/cancelled/stale chains abort instead of scrolling a re-rendered pane to an old target — this also closes a latent pane-flip stale-scroll bug.
- `action_unpin_log`: esc after a landed same-task jump now restores tail-follow (previously dead — nothing to unpin, `_log_follow_tail` stuck False).

## Tests

- Removed the three `@pytest.mark.flaky(reruns=2)` markers — with the root cause gone they'd only mask regressions of this fix; updated the stale `until()` docstring.
- Near-tail test: replaced the 0.12s late-fire drain with a deterministic wait for `_pending_jump is None` (with the fix, a starved jump lands *late* instead of vanishing, so the old drain could sample the anchor before a post-growth landing).
- New `test_journal_jump_survives_exhausted_scroll_retry_chain`: exhausts the chain deterministically (`attempts=0` against the unflushed hidden pane), asserts the jump survives, then proves a poll tick rescues it once the tab is revealed — reproduces the CI failure mode without needing a contended runner.

## Verification

- Full suite: 2407 passed, 1 skipped (`-n auto`).
- Determinism soak: 30× `pytest -k journal -p no:rerunfailures` — 30/30 green (note: the plugin name is `rerunfailures`; `-p no:flaky` is a no-op).
- `trunk check` clean.

Fixes #178

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved journal-to-log navigation so “jump” and “scroll to task” behavior reliably lands on the correct entry, even during delayed rendering and re-mapping.
  * Ensured pending navigation requests are retained until the UI confirms the scroll outcome (or determines it’s not possible).
  * Updated unpin behavior to better align with tail-following mode.

* **Tests**
  * Added regression tests covering navigation retry survival across skipped polling, delayed line recomputation, and flush-related scroll interruptions.
  * Reduced timing-dependent assertions to improve stability.

### DIFF
--- a/src/bmad_loop/tui/screens/dashboard.py
+++ b/src/bmad_loop/tui/screens/dashboard.py
@@ -378,7 +378,7 @@ class DashboardScreen(Screen[None]):
             self._pin_task = task
         self._tick(force_rescan=False)
 
-    def _scroll_log_to(self, attempts: int = 20) -> None:
+    def _scroll_log_to(self, attempts: int = 20, finalize: bool = False) -> None:
         jump = self._pending_jump
         if jump is None or jump[0] != self._displayed_log_task:
             # jump cancelled or superseded, or the pane re-rendered another
@@ -415,7 +415,22 @@ class DashboardScreen(Screen[None]):
             return
         viewport = max(1, log.scrollable_content_region.height)
         log.scroll_to(y=max(0, target - viewport // 2), animate=False)
-        self._pending_jump = None  # landed: release so later renders don't re-jump
+        if finalize:
+            self._pending_jump = None  # landed: release so later renders don't re-jump
+            return
+        # The open height gate proves the reveal flush *wrote* — not that its
+        # scroll drained: RichLog.on_resize replays the deferred writes, whose
+        # scroll_end goes through call_after_refresh, so a tail-scroll can
+        # still be queued on the log's pump while scroll_to (immediate on a
+        # ScrollView) already landed. Releasing here would let that queued
+        # scroll stomp the jump with nothing left to re-attempt it. Route the
+        # release through the same queue instead: FIFO puts the finalize fire
+        # after any queued stomp, and it re-scrolls to the recomputed target
+        # before letting go. If its guards bail (log reset mid-window), the
+        # jump stays pending for the next ≤1s tick.
+        log.call_after_refresh(
+            lambda: (self._scroll_log_to(finalize=True) if self._pending_jump is jump else None)
+        )
 
     def action_unpin_log(self) -> None:
         if self._resize_mode:  # Escape leaves resize mode before it unpins the log
--- a/src/bmad_loop/tui/screens/dashboard.py
+++ b/src/bmad_loop/tui/screens/dashboard.py
@@ -370,7 +370,7 @@ class DashboardScreen(Screen[None]):
         # retry chain delays the jump instead of losing it.
         self._pending_jump = (task, pos)
         if task == self._displayed_log_task and self._log_index is not None:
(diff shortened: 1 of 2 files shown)


pbean:
@coderabbit review

Issue #192, opened 2026-07-19
`bmad-loop list` has no --json, so its table format is a de-facto API


pbean:
**Version:** bmad-loop 0.8.1 · **Component:** `cli.py` `cmd_list`

## Problem

Same class of problem as #190, one command over: `bmad-loop list` prints a fixed-width `REF / TYPE / STATUS / RUN ID` table (`cmd_list`, cli.py) with no machine-readable form. It is the command that hands out the run refs every other command consumes, so scripts that want to enumerate runs (e.g. "stop everything interrupted", "archive everything finished") have to parse the column layout — which makes that layout an API nobody agreed to, exactly the hazard #190 closed for `status`.

## Proposed fix

Add `bmad-loop list --json` emitting an array of run entries, reusing the conventions `status --json` established (pure JSON on stdout, `schema_version`, additive-only evolution): per run `ref`, `run_id`, `run_type`, `started_at`, and the derived display status the table already computes (`tui/data.py` `RunInfo`/`_classify` — `finished`/`paused`/`interrupted`/…), which is richer than the raw state booleans because it accounts for liveness.

With that in place the table stays best-effort, matching the `status` doctrine.

Issue #189, opened 2026-07-19
Resumed runs enforce the current policy but display the launch-time one


pbean:
**Version:** bmad-loop 0.8.1 · **Component:** `cli.py` `_resume_paused_run`

## Problem

`state.policy_snapshot` is stamped only at run creation (`cli.py:681` for `cmd_run`, `cli.py:873` for `_start_sweep`). `_resume_paused_run` (`cli.py:997`) loads a fresh policy with `policy_mod.load(...)` and hands it to the rebuilt engine, but never re-stamps the snapshot.

From then on the run has two different policies in play:

- **Enforcement** reads live policy — e.g. the per-story budget check at `engine.py:2015`, and every value threaded onto a `SessionSpec`.
- **Displays** read the snapshot via `RunState.cache_read_weight()` (`model.py:379`) — the TUI, `bmad-loop status`, the run summary, and the `diagnose` bundle.

So edit `limits.cache_read_weight` between launch and resume and the run enforces at the new weight while every display reports the old one. At the legal extremes (0.1 ↔ 1.0, validated in `policy.py:640`) that is a 10x disagreement, silently.

The staleness is not specific to the weight — it applies to every snapshot-read policy value, and to the `policy` block of the diagnose bundle, which claims to describe the run that was executed.

## Proposed fix

Re-stamp `state.policy_snapshot = pol.to_dict()` on resume, after `state.clear_pause()` (`cli.py:1002`).

Worth noting either way: a run resumed across a weight edit accumulated tokens under two different weights, so **no** single weighted total is strictly correct for it. Whichever semantics is chosen should be documented and pinned by a test rather than left implicit.

Surfaced while fixing #129, which deliberately routes every display through the snapshot so the surfaces agree with each other. This issue is the remaining engine-vs-display gap.

Issue #129, opened 2026-07-12
Run summaries report raw tokens while budgets use weighted; neither unit is labeled


gnometa:
**Version:** bmad-loop 0.8.1 · **Component:** run summary / session-end events
**Reproduction:** run `20260709-172451-0187`

## Problem

Token accounting uses two different units and labels neither:

- The ATTENTION run summary reported `716,043,403 tokens` — **raw**, with cache reads at full weight.
- The budget machinery uses **weighted** tokens (`cache_read_weight = 0.1`), by which the same run cost ~110M — a 6.5× discrepancy.
- Session-end events carry raw only, so the weighted figure cannot even be reconstructed per session after the fact.

Every cost judgment an operator makes from the summary figure is inflated ~6.5× against actual billing.

## Proposed fix

Add a `weighted` field to session-end events, and lead run summaries with it:

> ~110M weighted tokens (716M raw incl. cache reads)

pbean:
tokens state the weighted count and raw state the unweighted count in the story TUI at this point in time. Im assuming you mean in the normal cli bmad-loop status command, or perhaps the output of a finished run from the cli? adding the information there was missed and I will update it to reflect those as well

pbean:
Confirmed and fixed in #188.

You were right that the CLI side was missed — `be7836b` fixed the TUI only, explicitly as "a pure observer", so the run summary, `bmad-loop status`, the `session-end` journal entries, and the `diagnose` bundle all kept the old raw-and-unlabelled behaviour. All four now lead with weighted and name both units.

Two things worth flagging beyond the original report:

- **The journal gap was the sharpest part.** `session-end` carried `tokens` as a bare scalar, and the weight can't be backed out of it — so per-session weighted spend genuinely wasn't reconstructible, exactly as you wrote. It now carries `tokens_weighted` alongside, `null` (never `0`) when the usage read failed.
- **The reference run is gone**, so I couldn't verify against `20260709-172451-0187` directly. The arithmetic checks out independently though: 716,043,403 raw against ~110M weighted implies 94% cache reads and a 6.51x ratio, matching your figure — and #128 cites weighted 3.4M–6.5M per story from the same run.

Two adjacent problems the investigation turned up are filed separately rather than folded in: #189 (a resumed run enforces the current policy but every display reports the launch-time snapshot) and #190 (`status` has no `--json`, which is why its text layout ends up carrying parsing weight it shouldn't).

[...]

PR #193, opened 2026-07-19
feat(cli): add status --json stable document (#190)

### LINKED ISSUE #190: `bmad-loop status` has no --json, so its text format is a de-facto API
**Version:** bmad-loop 0.8.1 · **Component:** `cli.py` `cmd_status`

## Problem

`--json` exists on `probe-adapter` (`cli.py:1967`) and `diagnose` (`cli.py:2073`) but not on `status`. Anyone who wants per-story phase / attempt / review-cycle / token data has only the human-readable line at `cli.py:1408-1411`, whose fields are positional and whose trailing `extra` is free text (a defer reason, so it can contain spaces).

That makes the text layout a de-facto API that nobody agreed to. Any formatting change is a silent breaking change for someone's script, with no deprecation path and no test coverage asserting the shape.

\#129 is a concrete instance: it changed both the unit of the token cell (raw → cost-weighted) and its shape (`<raw>t` → `<weighted>t (<raw> raw)`). That was the right fix — the old number overstated spend ~6.5x — but there was no supported way for a consumer to have been reading it, and no way to warn them.

## Proposed fix

Add `bmad-loop status --json` emitting a stable document: run id, run type, status/pause stage, and per-story `phase`, `attempt`, `review_cycle`, `tokens` (raw **and** weighted), `cache_read_weight`, `commit_sha`, `defer_reason`.

With a supported machine-readable surface, the human line can stay explicitly best-effort and evolve freely.

pbean:
Closes #190.

## Problem

`--json` exists on `probe-adapter` and `diagnose` but not `status`. The per-story text line was the only machine surface for phase/attempt/review-cycle/token data, making its layout a de-facto API: #129 changed the token cell's unit **and** shape with no deprecation path, and the line's trailing cell is an ambiguous `defer_reason or commit_sha` — a consumer cannot tell which one it got.

## Change

`bmad-loop status [--json]` now emits a single JSON document on stdout (nothing else printed, so `| jq` just works):

- `schema_version: 1` — the stable contract; changes are additive, breaking changes bump it (same convention as `diagnostics.SCHEMA_VERSION`).
- Run level: `run_id`, `run_type`, `source`, `started_at`, derived `status` (`finished`/`paused`/`crashed`/`stopped`/`in-progress`) beside the raw flags, `crash_error`, `paused_stage`/`paused_reason`/`paused_story_key`, snapshot `cache_read_weight`, and `tokens` (`raw` + `weighted`).
- Per story: `story_key`, `epic`, `phase`, `attempt`, `review_cycle`, `commit_sha`, `defer_reason` (finally separate fields), and `tokens` with the four raw counters plus derived `raw`/`weighted`.

Semantics deliberately match the existing surfaces: the weight comes from the run's **persisted policy snapshot** (never live policy, per #189), and the run weighted total is the per-task sum-of-rounds, same as `Engine.summary` — the document is reproducible from `state.json` alone, which is also why the stories board / sprint backlog / decisions count (read from other files) stay text-only. Error paths (`no runs`, unknown ref) keep stdout empty with exit 1, so a JSON consumer never sees a partial document.

Divergence from the `probe-adapter`/`diagnose` convention (fenced `## JSON` block appended to markdown) is intentional: those are human-facing reports; this is an API surface for scripts. With it in place the human text output is now documented as explicitly best-effort. No scrubbing/pseudonymization (contrast `diagnose`): this is a local surface where the real identifiers are the point.

## Tests

8 new tests in `tests/test_cli.py`, parsing the **whole** of stdout as the purity assertion: document shape + phase serialization, snapshot-sourced weight + sum-of-rounds vs round-of-sum (101+101 cache reads at 0.5 → 100, not 101), numeric-zero weighted at weight 0, paused-run fields, defer-vs-commit disambiguation, stories-mode purity (no board text), and empty-stdout error paths. Full suite: 2521 passed, 1 skipped; trunk check clean. Also live-verified in a sandbox (`--json | python -m json.tool`, `jq '.tasks[].phase'`, text mode byte-identical).

## Follow-up

`bmad-loop list` has the same de-facto-API problem — filed as #192 rather than widened into this PR.

## Summary by CodeRabbit

* **New Features**
  * Added `bmad-loop status --json` for stable, schema-versioned machine-readable status output.
  * JSON results include run state, token totals, cache weighting, and per-story task details.
  * Added support for retrieving status for a specific run.

* **Bug Fixes**
  * JSON error responses now keep stdout empty and provide human-readable messages on stderr.

* **Documentation**
  * Documented JSON fields, scripting usage, error behavior, and best-effort text output.


pbean:
@coderabbitai review

PR #198, opened 2026-07-19
feat(cli): add list --json stable document

### LINKED ISSUE #192: `bmad-loop list` has no --json, so its table format is a de-facto API

pbean:
Closes #192

Phase 2 of the machine-output plan (#197 was Phase 1: the `machine.py` contract).

## What

`bmad-loop list --json` emits a stable, schema-versioned pure-JSON document instead of the text table:

```json
{
  "schema_version": 1,
  "runs": [
    {"ref": "a1b2", "run_id": "20260101-120000-a1b2", "run_type": "story",
     "started_at": "2026-01-01T12:00:00", "status": "finished", "paused_stage": ""}
  ]
}
```

- Pure projection of `discover_runs()` — the same liveness-aware status vocabulary the table shows (`running|paused|finished|stopped|crashed|interrupted|unknown`), oldest first.
- Envelope object (not a bare array) so top-level evolution stays additive; `LIST_SCHEMA_VERSION = 1`.
- `ref` is `runs.short_ref(run_id)` — derived from the id, stable, not positional.
- Unparseable-state runs are included (`run_type` "?", `started_at` "", status "unknown") — enumeration scripts must see them, same as the table.
- `paused_stage` is a bonus field the text table drops; "" unless paused.
- Empty runs dir → `{"schema_version": 1, "runs": []}` with exit 0 — never the text "no runs found" (exit-code parity holds).
- Obeys the `machine.py` pure-document contract; flag registered via `machine.add_json_flag`; text table unchanged.

## Tests

Four new tests via the shared `_machine_json` purity helper (parsing the whole stdout is the purity assertion): populated document with field-exact entries + oldest-first order, valid empty document, corrupt `state.json` → "unknown" entry, paused run carries its stage. Fixtures stick to the deterministic statuses (finished/paused/stopped) — running/interrupted probe pid liveness and would flake.

## Verification

- Full suite: 2525 passed, 1 skipped; full `trunk check` clean.
- Smoked `bmad-loop list --json | jq .` in a scratch project: empty case, finished + paused runs, and the text table unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added `--json` output to `bmad-loop list`/`ls`.
  * JSON results use a stable schema, list runs oldest first, and include run status and pause details.
  * Unparseable runs are included as `unknown`.
  * Empty run directories return a valid empty document with exit code 0.

* **Documentation**
  * Updated the changelog and feature documentation with JSON output details.


pbean:
@coderabbit review

Issue #200, opened 2026-07-19
diagnose --json can raise UnicodeEncodeError on a non-UTF-8 console


pbean:
## Problem

`diagnose --json` can now crash with `UnicodeEncodeError` on a console whose encoding cannot represent a non-ASCII character in the document.

```
$ PYTHONIOENCODING=ascii python -c "from bmad_loop import machine; machine.emit_document('{\"os_release\": \"5.15-café\"}')"
UnicodeEncodeError: 'ascii' codec can't encode character '\xe9' in position 24: ordinal not in range(128)
```

## Why the exposure is new

`diagnostics.render_json` previously used `json.dumps`'s default `ensure_ascii=True`, which structurally guaranteed a pure-ASCII document. #195 changed it to `ensure_ascii=False` at `src/bmad_loop/diagnostics.py:527` and `:537`, so raw non-ASCII can now reach `machine.emit_document`'s `print(rendered)` at `src/bmad_loop/machine.py:66`.

**That change is load-bearing and must not simply be reverted.** It is a safety requirement, not cosmetics: with `ensure_ascii=True` a non-ASCII sensitive value is escaped to `\uXXXX` *before* `sanitize.assert_no_leak` scans the rendered bytes, so the guard misses it entirely. Demonstrated against the real sanitizer:

```python
uni = 'café-user'
sanitize.assert_no_leak(f'u: {uni}', extra=[(uni, 'username')])                              # ['sensitive[username]']
sanitize.assert_no_leak(json.dumps({'u': uni}), extra=[(uni, 'username')])                   # []   <-- evades
sanitize.assert_no_leak(json.dumps({'u': uni}, ensure_ascii=False), extra=[(uni, 'username')])  # ['sensitive[username]']
```

So the guard needs the raw form to scan. The open question is only what gets *written* afterwards.

## Scope and severity

Narrow, and it fails safe:

- Requires non-ASCII that **survived** the guard — i.e. a non-sensitive value such as a localized `platform.release()` — **and** a stdout encoding that cannot represent it, in practice a legacy non-UTF-8 Windows console.
- `str.encode` runs before any write, so nothing is emitted and stdout stays empty. That is a contract-permitted error shape (`machine.py`), not a partial document.
- `--out FILE` is unaffected: `src/bmad_loop/cli.py:1951` writes with an explicit `encoding="utf-8"`.

[...]

Issue #199, opened 2026-07-19
probe-adapter report has no egress leak guard and no pseudonymizer


pbean:
## Problem

`probe-adapter`'s report is advertised as sanitized and is meant to be shared — `docs/adapter-authoring-guide.md:168` says both modes "emit the **same single sanitized report**", and the guide's whole purpose is producing an artifact you attach when contributing a profile. But `probe.py` has neither of the two egress protections `diagnostics.py` relies on for the same sharing expectation.

**1. No final-output leak guard.** `sanitize.assert_no_leak` is never called in `probe.py`. `diagnostics.py` re-scans its rendered bytes at `:678`, `:705`, `:720` and *refuses to emit* on a hit (`LeakDetected`, `:727`). `probe.py` scrubs at collection time — `:170-171` (`scrub_text` on `--version`/`--help`), `:184` (`_redact_location` per path component), `:288` (dynamic JSON keys), `:490` (`scrub_event_payload`), `:654` (log tail) — and then ships whatever those produced. If any one scrubber has a routing gap, nothing catches it. That is precisely the failure mode the diagnose backstop was built for, and #186/#187 showed such gaps are real rather than hypothetical.

**2. No `Pseudonymizer` — identifier-shaped values pass verbatim.** `probe.py` uses only `scrub_json`/`scrub_event_payload`. Per `sanitize.py:242-244`:

> Story keys, branch names, spec filenames and SHAs are identifier-shaped, so `scrub_json` would pass them verbatim and leak the customer's feature names.

That is the stated reason `diagnostics.py` routes those values through `Pseudonymizer.alias` instead. `probe.py` has no equivalent, so any identifier-shaped proprietary value reaching a scrubbed field survives into the report under its real name.

The widest surface is `EventCapture.payload` (`probe.py:120`, comment `# scrubbed`), which embeds an entire captured hook payload dict and is emitted whole into both the markdown report and the JSON document (`probe.py:807+`). `--probe` mode captures these from a live session.

## Severity

Defense-in-depth gap, not a demonstrated leak — I have not shown a specific real payload that leaks. What would settle it: run `probe-adapter <cli> --probe` in a project whose story keys / branch names are distinctive and grep the report for them, mirroring the canary approach in `tests/test_diagnostics.py`. `tests/test_probe.py:312` (`test_scan_report_contains_no_pii`) checks only for an email and a path fragment in SCAN mode, so the identifier class and the `--probe` capture path are both untested at egress.

## Proposal

[...]

PR #201, opened 2026-07-19
refactor(cli): unify diagnose/probe-adapter --json with the pure-document contract

### LINKED ISSUE #195: Unify diagnose/probe-adapter --json with the pure-document contract
## Problem

The CLI now has two incompatible meanings for the same `--json` flag:

- `status --json` (#190) emits a **pure JSON document**: exactly one JSON object on stdout, nothing else; inline `schema_version`; errors go to stderr with empty stdout and exit ≠ 0. This is the contract being codified in `src/bmad_loop/machine.py`.
- `diagnose --json` and `probe-adapter --json` predate that contract: they **append a fenced ```json block to a markdown/text report**. A consumer cannot `json.loads` the stream; it has to scrape the fence out of prose.

Same flag, divergent semantics — scripts that learn one shape break on the other.

## Proposal

Bring `diagnose` and `probe-adapter` onto the pure-document contract. Two options to weigh at pickup:

1. **New flag** (e.g. `--json-doc`, or repurpose `--json` and move the current behavior to `--report-json`) — non-breaking, but leaves the flag soup.
2. **Breaking change behind a schema bump**: `--json` emits the pure document (for diagnose, essentially the existing embedded block promoted to the whole stdout, report text dropped or moved behind the default text mode). `diagnose` already has `diagnostics.SCHEMA_VERSION` to bump.

Either way the human-readable report stays available — it is the artifact users hand to maintainers.

## Notes

- `diagnose --out FILE` interacts with this: the pure-document form should respect it (document to the file, nothing on stdout) or reject the combination explicitly.
- `diagnose`'s refusal path (sanitizer fails closed) must keep the contract's error shape: message → stderr, empty stdout, exit ≠ 0 — never a partial document.

pbean:
Closes #195.

## What

`--json` had two incompatible meanings. `status`/`list` emitted a pure JSON document; `diagnose`/`probe-adapter` appended a fenced ` ```json ` block to a markdown report, so a consumer could not `json.loads` the stream — it had to scrape the fence out of prose.

Both commands now emit the document **instead of** the report:

- stdout parses whole, no fences.
- Every human-facing line — `ok:` trailers, the leak-backstop warning, the `unknown profile` notice — moves to stderr.
- `--json --out FILE` writes the document to the file, leaves stdout empty, and puts the confirmation on stderr.
- Text mode (no `--json`) is unchanged; the human-readable report is still what users hand to maintainers.

`machine.py`'s carve-out naming these two commands as exceptions is deleted, along with its mirrors in `FEATURES.md`, the adapter-authoring guide, `README.md`, and the bug-report issue template (which told reporters to expect a fenced block).

## Schema versions

`diagnostics.SCHEMA_VERSION` **stays at 1**, deliberately. `machine.py` defines that number as a property of the *document* — it bumps when a field is removed or renamed, a type changes, or a value's meaning changes — and none of those happened here. Only the packaging changed. Bumping it would falsely tell a consumer pinned to v1 that the fields it reads are gone, while a consumer genuinely broken by the repackaging finds out immediately: the fence is gone and `json.loads` fails.

`probe` gains `schema_version: 1` as its baseline, alongside its existing `version` key — which still holds the *probed CLI's* `--version` output and is deliberately not renamed.

## Leak-guard hardening (the part worth reviewing closely)

Making `diagnose --json` skip `render_markdown` turned out to weaken the sanitizer, because `json.dumps` escapes. Both of these fire the guard as raw text and evaded it entirely inside the JSON document:

| value | raw text | JSON (before) |
|---|---|---|
| `C:\Users\alice` | `absolute-home-path` | *nothing* — `\` doubled past `_ABS_HOME_RE` |
| non-ASCII username | `sensitive[username]` | *nothing* — escaped to `\uXXXX` |

Previously the markdown pass caught both and refused to emit. Fixed at the source: `_ABS_HOME_RE` now matches the escaped separator, and `render_json` uses `ensure_ascii=False` so values reach the guard as themselves. Both have regression tests that were verified to fail when the fix is reverted.

## Incidental fixes

- `diagnose --json` no longer renders a markdown report it was about to discard — which was double-counting every leak-backstop repair in the operator warning (`x2` for a single occurrence). The count is now pinned by an assertion.
- The probe document is `sort_keys`-stable, so two probes of the same CLI diff cleanly. Pinned by a key-order assertion (a plain `json.loads` round-trip cannot see order).

## Testing

`2541 passed, 1 skipped`; `trunk check` clean across 13 files.

Beyond the suite, each new guarantee was checked by reverting the fix and confirming the test fails — the guard hardening, the repair count, and `sort_keys`. Two pre-existing tests needed retargeting: they monkeypatched `render_markdown` to force the refusal path, which JSON mode no longer reaches, so they would have passed vacuously. They are now parametrized `[text, json]` and patch whichever render the mode actually uses.

Live smoke on both commands: pure document, zero fences, `--out` leaving stdout at 0 bytes, error paths exiting nonzero with empty stdout.

## Deliberately out of scope

- **#199** — `probe.py` has no egress leak guard and no pseudonymizer. `sanitize.py`'s own docstring notes that `scrub_json` passes story keys and branch names verbatim, and probe never re-scans its rendered bytes. Untouched here; this PR changes packaging, not what may appear inside the document.
- **#200** — `ensure_ascii=False` lets raw non-ASCII reach stdout, which can raise `UnicodeEncodeError` on a non-UTF-8 console. Narrow (needs non-sensitive non-ASCII *and* a legacy console) and fails safe: the encode precedes any write, so stdout stays empty, and `--out` pins `encoding="utf-8"`. **Note for reviewers:** do not "fix" it by reverting `ensure_ascii=False` — that flag is what lets the guard see non-ASCII values at all. The issue documents the tradeoffs.

## Summary by CodeRabbit

* **Changed**
  * `--json` for `status`, `list`, `diagnose`, and `probe-adapter` now emits a pure stable JSON document to stdout; when using `--out FILE`, stdout is empty and only the file receives JSON.
  * Human-readable output/`ok:` trailers go to stderr in JSON mode, with consistent JSON across output paths.
* **Bug Fixes**
  * Strengthened protection against leaking escaped home paths and non-ASCII sensitive values; JSON output now preserves UTF-8 reliably.
* **Documentation**
  * Updated the command reference, features/docs, adapter guidance, and the bug-report template to match the new `--json` behavior.


pbean:
Both nitpicks validated. One is a false positive; the other pointed at a real gap, though not the one it named. Pushed as f204c2e.

## `sanitize.py` — forward slashes in `_ABS_HOME_RE`: **declining, the case is already covered**

The premise is that `C:/Users/alice` isn't caught. It is — the `/Users/` alternative matches it as a substring, so the drive-letter arm never has to. Measured, current rule vs. the proposal:

| value | current | proposed |
|---|---|---|
| `C:/Users/alice/proj` | **True** | True |
| `c:/users/bob` | **True** | True |
| `C:\Users\alice` (and JSON-doubled `C:\\Users\\alice`) | **True** | True |
| `/c/Users/alice` (git-bash) | **True** | True |
| `C:\Users/alice` *(mixed separators)* | False | True |
| `D:/data/alice` (non-home drive) | False | False |

The only delta is the mixed-separator oddity, which has no realistic producer — and any string carrying a separator at all is rejected by `looks_like_identifier` upstream (`_IDENTIFIER_RE` admits neither `/` nor `\`) and redacted to `<redacted:str>` before it can reach the backstop.

Worth saying that the finding was reasonable to raise: the comment above the rule explained only the backslash-doubling rationale, so it read as if Windows were handled *solely* by the drive-letter arm. That was a real documentation gap even though the code was right. Fixed both ways — the comment now states why the drive-letter arm is backslash-only, and `test_assert_no_leak_fires` now pins `/Users/`, `/root/` and `C:/Users/…` instead of just `/home/alice/x`. Those cases were correct only by accident; a future "simplification" of the alternation could have silently dropped them. Verified by deleting the `/Users/` arm and confirming both new cases go red.

Added a negative case too (`D:/data/alice`, `/var/lib/alice/x`), because the bound matters in both directions: firing is fail-closed, so a rule that matched every absolute path would turn an ordinary dump into a refusal.

## `cli.py` — centralizing the `--out` write + trailer: **taking the direction, not the signature**

The sketched `write_or_emit(..., ok_message, trailer_stream)` can't express `cmd_probe`, which prints a trailer in *both* branches with different text computed from different data (`finding.mode`/`args.cli` vs. `out_path`), while `cmd_diagnose` prints one only under `--out`. Threading that through would need a second message parameter used by one caller — an abstraction that fits neither site.

But the duplication *was* hiding something. `machine.emit_document` parses a document before printing it, deliberately, so "stdout is either one complete valid document or empty" holds structurally rather than by convention. The `--out` branch bypassed it entirely — plain `write_text` — so `--json --out FILE` could write a malformed document that stdout would have refused to print. That's backwards: the file is the one nobody eyeballs before feeding it to a parser.

So the extracted piece is the contract, not the control flow. New `machine.write_document(path, rendered)` is the `--out` sibling of `emit_document`, sharing a `_validated` parse-or-raise; it refuses before creating the file, and appends the same trailing newline `print` would. `--json --out FILE` and `--json > FILE` are now byte-identical — pinned by a test rather than by an explanatory comment duplicated at two call sites. Trailer logic stays exactly where it was.

Smoke-confirmed live on both commands: `--out` leaves stdout at 0 bytes, the file parses, no fences, and it `diff`s clean against the piped form.

## Docstring coverage (57.78%) — no action

Chronic rather than a regression from this PR: #187 merged at 44.44%, #188 at 60.00%, #191 at 66.67%. Nothing enforces it — `pyproject.toml` has no `[tool.ruff.lint]` table at all (so no pydocstyle `D` rules), there's no `.coderabbit.yaml` opting into the 80% default, and `.trunk/trunk.yaml` enables no docstring linter. This PR's only genuinely new function, `machine.emit_document`, is fully documented; the percentage is measuring pre-existing `cmd_*` handlers that the repo leaves undocumented as house style, with their contracts in argparse help and module docstrings. Clearing it would mean documenting ~7 untouched functions in a style the codebase deliberately doesn't use.

`2547 passed, 1 skipped` (+6 from the 2541 baseline); `trunk check` clean across 14 files.

pbean:
@coderabbit review

Issue #196, opened 2026-07-19
Adopt --json on remaining script-facing commands (decisions --list, clean/cleanup --dry-run, validate)


pbean:
## Tracking: adopt `--json` (pure-document contract) on the remaining script-facing commands

`status --json` (#190) and `list --json` (#192) emit stable, schema-versioned pure JSON documents per the contract in `src/bmad_loop/machine.py`. Other commands whose text output is a de-facto API for scripts are still text-only:

- [x] `decisions --list` (#203) — pending decisions with ids/options; scripts currently scrape the numbered text.
- [x] `clean` / `cleanup --dry-run` (#204) — what *would* be removed (worktrees, runs past retention, tmux sessions); the natural machine surface for pre-flight checks.
- [x] `validate` (#205) — pass/fail plus per-check findings; exit code alone loses the detail.

Now split into the per-command issues linked above — each needs its own `*_SCHEMA_VERSION`, document builder, and purity tests via the shared `machine_json` helper in `tests/conftest.py`. Not a commitment to do all of them; strike any that turn out to have no script consumers.

Related: #195 tracks unifying the older report+embedded-block form of `diagnose --json` / `probe-adapter --json` with this contract.

pbean:
Scope fully shipped — closing. All three per-command issues landed, but none of the PRs used a closing keyword against this tracking issue, so it stayed open.

| command | issue | PR |
| --- | --- | --- |
| `decisions --json` | #203 | #207 (merged) |
| `clean` / `cleanup --json` | #204 | #208 (merged) |
| `validate --json` | #205 | #209 (merged) |

Each shipped its own `*_SCHEMA_VERSION` (all at 1), a document builder, and purity tests through the shared `machine_json` helper in `tests/conftest.py`, as planned above. Nothing was struck — all three had real script consumers.

The successor thread is #210: the TUI's validate modal should consume `validate --json` and render the findings structurally instead of scraping text. That is the first in-tree consumer of one of these documents, and the stepping stone toward a web UI reading the same contract.

Also relevant to anyone picking that up: #211 extracts the six document builders out of `cli.py` into a new `documents.py` library layer, so a non-CLI frontend can import them directly rather than shelling out to the CLI. (#195 remains the separate thread for `diagnose` / `probe-adapter`.)

PR #207, opened 2026-07-19
feat(cli): decisions --json emits the pending decisions as a document (#203)

### LINKED ISSUE #203: Adopt --json on decisions --list (pure-document contract)
## `decisions --list --json`

Part of #196 (split into per-command issues).

`bmad-loop decisions --list` prints unanswered deferred-work decisions as
numbered prose — id, question, then an indented option block per choice with a
`(recommended)` marker (`cli.py:1400-1409`). Scripts that want to pre-answer
decisions non-interactively have to scrape that layout today, including the
recommendation marker, which is a suffix on a free-text line.

**Purpose of the document.** Expose the pending decision set as data: each
decision's `id`, `question`, its `options` (each with `key`, `label`, `effect`),
and which option key is the `recommendation`. That is enough for a caller to
render its own prompt, or to select an option by policy and feed the answer
back, without parsing the human layout.

**What it needs.**
- Its own `DECISIONS_SCHEMA_VERSION` in `cli.py`, alongside
  `STATUS_SCHEMA_VERSION` / `LIST_SCHEMA_VERSION` — each command's document
  versions independently (see the `machine.py` module docstring).
- A document builder separate from the printing path, so the text and JSON
  forms cannot drift.
- Purity tests via the shared `machine_json` helper in `tests/conftest.py`.

**Note on the empty case.** `decisions` returns early with `no unanswered
decisions from past sweeps` and rc 0 when nothing is pending (`cli.py:1397`).
Under `--json` that must still emit a complete, valid document with an empty
decision list at rc 0 — the same call `list --json` made in #192. Empty stdout
is reserved for errors.

Contract: `src/bmad_loop/machine.py`.

pbean:
Closes #203. Part of #196.

## What

`decisions --list` prints unanswered deferred-work decisions as numbered prose, so a script that wants to pre-answer them by policy has to scrape that layout — including the recommendation, which is a `(recommended)` suffix on a free-text line.

Adds `DECISIONS_SCHEMA_VERSION = 1` and a pure `_decisions_document` builder alongside the `status` / `list` exemplars, emitting per the pure-document contract in `machine.py`:

```json
{
  "schema_version": 1,
  "decisions": [
    {
      "id": "DW-1",
      "question": "build the widening?",
      "context": "the parser only accepts ASCII keys",
      "recommendation": "2",
      "options": [
        { "key": "1", "label": "Widen", "effect": "build", "intent": "accept unicode keys",
          "resolution": "", "bundle_name": "unicode-keys", "recommended": false },
        { "key": "2", "label": "Close", "effect": "close", "intent": "",
          "resolution": "ASCII is the documented contract", "bundle_name": "", "recommended": true }
      ]
    }
  ]
}
```

## Why it is more than a reformat

The text form is lossy in two ways a scraper cannot work around: it drops `Decision.context` outright, and shows only key/label/effect per option — hiding `intent`, `resolution` and `bundle_name`, the fields that decide what the next sweep actually builds or writes. The document carries the whole dataclass. `recommended` is derived from `key == decision.recommendation`, so the two can never disagree.

## Two wiring points, both load-bearing

- **`--json` takes its branch regardless of `--list`.** The default path constructs the interactive `DecisionPrompter`, which reads stdin and prints per-answer progress; neither survives a pure document. So `--json` implies the listing rather than falling through to a prompt that would hang a caller's pipeline. The test booby-traps both the prompter and `input`, so a regression fails loudly instead of hanging.
- **The branch sits before the `if not pending` early return.** Nothing pending is a valid empty document at rc 0 — not the text `no unanswered decisions from past sweeps`, which would corrupt the stream. Same call `list --json` made for an empty runs dir in #192. Empty stdout stays reserved for errors, as the untouched `BmadConfigError` path (stderr, rc 1) still is.

## Tests

Five, all via the shared `machine_json` helper from `tests/conftest.py` (parsing the whole of stdout *is* the purity assertion):

- every field round-trips, including the three the text drops
- exactly one option flagged `recommended`, and it is the one the recommendation names
- empty ledger yields a valid empty document at rc 0
- `--json` without `--list` emits the document and never reaches the prompter or stdin
- the config-error path leaves stdout empty at rc 1

`uv run pytest -q`: 2553 passed, 1 skipped. Full `trunk check`: clean.

## Docs

`docs/FEATURES.md` command-reference bullet (linking the contract anchor, and adding `decisions` to the section's list of commands taking the flag), the `README.md` command-table row, and a CHANGELOG `### Added` entry.

## Summary by CodeRabbit

* **New Features**
  * Added `bmad-loop decisions --json` for stable, machine-readable output of pending decisions.
  * JSON output includes complete decision and option details, schema versioning, and recommended-option indicators.
  * JSON mode never prompts for input and returns a valid empty document when no decisions are pending.

* **Documentation**
  * Updated the command reference and feature documentation with JSON usage, output structure, and listing behavior.


pbean:
@coderabbit review

PR #211, opened 2026-07-20
refactor(cli): extract the --json document builders into documents.py

pbean:
Pure relocation, no issue (precedent: #197, #206).

## Why

The six per-command `--json` builders and their schema constants lived as private functions in `cli.py`:

| builder | constant |
| --- | --- |
| `_validate_document` | `VALIDATE_SCHEMA_VERSION` |
| `_decisions_document` | `DECISIONS_SCHEMA_VERSION` |
| `_status_document` | `STATUS_SCHEMA_VERSION` |
| `_list_document` | `LIST_SCHEMA_VERSION` |
| `_cleanup_document` | `CLEANUP_SCHEMA_VERSION` |
| `_clean_document` | `CLEAN_SCHEMA_VERSION` |

That was ~314 lines of the dispatch module, growing by a few hundred with every command that adopts the flag, and reachable only through `argparse`. But the contract they serve (`machine.py`) is not a CLI feature — the planned web app's Python backend needs to import these as a library, not shell out to the CLI and parse its stdout.

`probe.py` / `diagnostics.py` already make this split ("one finding, two render targets"). This generalizes it to the commands whose document is a dict rather than a rendered string.

## What

New `src/bmad_loop/documents.py`: the library-level read-model projection layer — domain object in, contract document dict out, obeying `machine.py`'s pure-document contract. No I/O, no process state, no printing, no exit codes: the caller loads, this layer projects, `machine.emit` writes. New commands add their builder here, not in `cli.py`.

`cli.py` re-imports the names explicitly, so `cli.STATUS_SCHEMA_VERSION` and `cli._status_document` keep resolving for existing callers.

`_run_token_totals` moves too — `_status_document` calls it, and leaving it behind would have made `documents.py` import the CLI. `cmd_status` still uses it for text rendering, through the same re-import.

## This is a pure relocation

Zero behavior change, zero schema change, and no test file was touched. The existing tests are the proof: they pin `doc["schema_version"] == cli.X_SCHEMA_VERSION`, so they only pass if both the re-export and the document shape survived.

- **Verbatim bodies** — AST-extracted each moved definition from `git show HEAD:src/bmad_loop/cli.py` and from the new module, compared as text: all 13 (7 functions + 6 constants) byte-identical, and none still defined in `cli.py`.
- **Full suite green, zero test edits** — 2577 passed, 1 skipped; `git diff -- tests/` empty.
- **Byte-identical output** — `status --json` and `list --json` diffed before vs after on a fixture run covering all four terminal statuses (finished/stopped/paused/crashed), a persisted `cache_read_weight` snapshot, and a deferred task with a defer reason. Identical stdout, rc and stderr.
- **`trunk check` (no filter) clean.**

## One note for review

The six constants are re-exported but not referenced in `cli.py`, so ruff F401 flags them — and its autofix *removes the re-export*, breaking every `cli.*_SCHEMA_VERSION` reference. Caught here by the test suite. Each import carries a `# noqa: F401 — re-export` with a comment saying why; worth keeping when the next command lands.

`import bmad_loop.documents` costs ~35ms and pulls no `textual`, so the library layer is usable from a non-CLI frontend without dragging in the TUI.

## Summary by CodeRabbit

* **Refactor**
  * Centralized JSON output generation across validation, decisions, status, listing, cleanup, and clean commands.
  * Preserved existing command usage and output contracts while improving consistency across JSON responses.
  * JSON responses now include stable schema version information for supported commands.

### DIFF
--- a/src/bmad_loop/cli.py
+++ b/src/bmad_loop/cli.py
@@ -11,7 +11,6 @@ import subprocess
 import sys
 import time
 from pathlib import Path
-from typing import TYPE_CHECKING
 
 from . import (
     __version__,
@@ -33,6 +32,28 @@ from . import (
 )
 from .adapters.base import CodingCLIAdapter
 from .checks import Finding, ValidationReport
+
+# The --json document builders live in documents.py (the library-level projection
+# layer a non-CLI frontend imports). The schema constants are re-exported rather
+# than used here — `cli.STATUS_SCHEMA_VERSION` and friends are the published names
+# consumers and tests already resolve — so each carries a noqa: without it ruff
+# F401 autofixes the re-export away and every `cli.*_SCHEMA_VERSION` reference
+# breaks. Keep the noqa when adding a command.
+from .documents import CLEAN_SCHEMA_VERSION  # noqa: F401 — re-export
+from .documents import CLEANUP_SCHEMA_VERSION  # noqa: F401 — re-export
+from .documents import DECISIONS_SCHEMA_VERSION  # noqa: F401 — re-export
+from .documents import LIST_SCHEMA_VERSION  # noqa: F401 — re-export
+from .documents import STATUS_SCHEMA_VERSION  # noqa: F401 — re-export
+from .documents import VALIDATE_SCHEMA_VERSION  # noqa: F401 — re-export
+from .documents import (
+    _clean_document,
+    _cleanup_document,
+    _decisions_document,
+    _list_document,
+    _run_token_totals,
+    _status_document,
+    _validate_document,
+)
 from .engine import Engine
 from .journal import Journal, load_state, save_state
 from .model import RunState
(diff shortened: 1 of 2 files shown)


pbean:
@coderabbit review

pbean:
@coderabbitai Valid, and the argument is stronger than stated — but deliberately out of scope here. Filed as #212.

You're right that the leading underscore contradicts the module docstring's own stated purpose. What sharpens it: the precedent this layer is modeled on names its whole projection surface public. `probe.py` exposes `scan` / `probe` / `render_markdown` / `render_json` / `SCHEMA_VERSION`; `diagnostics.py` exposes `collect` / `collect_run` / `render_json` / `render_markdown` / `SCHEMA_VERSION`. Not one underscore between them. And in `documents.py` the six `*_SCHEMA_VERSION` constants are already public while the builders are not — a split with no rationale behind it, just an artifact of what the names were inside `cli.py`.

I also checked the constraint your suggestion would run into, and it isn't there: no test calls a builder directly (the only grep hit is prose inside a docstring), so a rename would not force test edits.

Not doing it in this PR, for a reason specific to what this PR is. Its deliverable is a *provable* pure relocation — the bodies were verified byte-identical by AST-extracting each definition from `git show HEAD:src/bmad_loop/cli.py` and comparing as text, with the pre-existing `doc["schema_version"] == cli.X_SCHEMA_VERSION` assertions passing untouched as the proof. A rename in the same commit voids both halves of that and breaks the requirement that `cli._x_document` keep resolving. It is a separable change and reads better as one.

#212 covers it, including the one real judgement call you didn't have the context to see: `_run_token_totals` is also used by `cmd_status` for *text* rendering, so it may belong as a private sibling rather than part of the document API.

PR #213, opened 2026-07-20
fix(cli): emit --json documents to a UTF-8 stdout (#200)

### LINKED ISSUE #200: diagnose --json can raise UnicodeEncodeError on a non-UTF-8 console

pbean:
Closes #200.

## The bug

`diagnose --json` could die with `UnicodeEncodeError` on a console whose encoding cannot carry a non-ASCII character in the document:

```
$ PYTHONIOENCODING=ascii python -c "from bmad_loop import machine; machine.emit_document('{\"os_release\": \"5.15-café\"}')"
UnicodeEncodeError: 'ascii' codec can't encode character '\xe9' in position 24: ordinal not in range(128)
```

The exposure arrived with #195, which changed `diagnostics.render_json` to `ensure_ascii=False`. That flag is a safety requirement, not cosmetics — with the default, a non-ASCII sensitive value is escaped to `\uXXXX` *before* `sanitize.assert_no_leak` scans the rendered bytes, so the guard misses it entirely. The guard needs the raw form. The only open question was what gets *written* afterwards.

Narrow and fail-safe as filed: it needs non-ASCII that survived the guard (so a non-sensitive value — a localized `platform.release()`) **and** a stdout encoding that cannot represent it, in practice a legacy non-UTF-8 Windows console. `str.encode` runs before any write, so stdout stayed empty rather than half-written — a contract-permitted error shape. But `diagnose --json` still died on a machine where `--json --out FILE` would have worked.

## The fix

Option (2) from the issue: `machine.emit_document` switches stdout to UTF-8 before the write, guarded by `hasattr` since a substituted stdout (pytest capture, an exotic stream) need not be a `TextIOWrapper`. On that fallback the pre-existing crash-with-empty-stdout is what the contract already permits.

Rejected alternatives, and why:

- **Guard on the raw form, emit the escaped form** (option 3). Smaller, no encoding risk at all, and escaping provably cannot introduce content. But it breaks the "emit exactly the bytes the leak check verified" invariant that `emit_document` exists to enforce, which is worth more than this edge costs.
- **Write `sys.stdout.buffer` directly** (option 1). Deterministic, but bypasses `sys.stdout` and so breaks `capsys` across every `--json` test.
- **Reverting `ensure_ascii=False`.** Reopens the leak-guard evasion #195 closed.

`reconfigure(encoding=...)` preserves `newline`, `line_buffering` and `write_through`, so it changes nothing but the codec — verified against CPython 3.13 before adopting it.

`write_document` needed no change: `--out FILE` has always written `encoding="utf-8"`. `diagnostics.py:524` already asserted the document "is only ever written with `encoding=\"utf-8\"` … or printed to a UTF-8 stream" — this makes the second half true rather than assumed.

Also adds a note to `machine.py`'s module docstring recording that the per-family serializer flags (`sort_keys`, `ensure_ascii`) differ **on purpose** and are not to be unified: the `emit_document` renderers sort keys for a diff-stable dump, `emit`'s dicts are built in the order they are meant to be read, and only `diagnostics.render_json` needs `ensure_ascii=False`.

## Tests

`test_emit_document_writes_utf8_to_a_console_that_cannot_encode_the_document` drives the emit path against a `TextIOWrapper(encoding="ascii")` stdout with a non-ASCII value, and asserts the bytes decode as UTF-8 to the verbatim document. It fails pre-fix with exactly the reported `UnicodeEncodeError` (confirmed by stashing only `machine.py`), and the issue's original `PYTHONIOENCODING=ascii` subprocess repro now round-trips through a real pipe.

`test_emit_document_survives_a_stdout_that_cannot_be_reconfigured` covers the `hasattr` fallback with a `StringIO` — the shape that proves the guard does not itself become the crash.

The `machine_json` conftest helper stays the purity net for the normal path; it is untouched.

Full suite green (2579 passed, 1 skipped), `trunk check` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `diagnose --json` failures on consoles that cannot encode non-ASCII characters.
  - JSON output now writes UTF-8 content reliably while preserving non-ASCII characters.
  - Output redirected to a file continues to work as before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -257,6 +257,19 @@ PATH)`, the TUI notifies `multiplexer backend unavailable — launch/attach disa
 
 ### Fixed
 
+- **`--json` output survives a console that cannot encode it (#200).** A JSON document is not
+  necessarily ASCII: `diagnostics.render_json` serializes with `ensure_ascii=False` so its leak
+  guard can scan values unescaped (#195, below), which lets a _non_-sensitive non-ASCII field —
+  a localized `platform.release()`, say — reach stdout verbatim. Printing it to a console whose
+  encoding could not carry it raised `UnicodeEncodeError`, in practice a legacy non-UTF-8
+  Windows one. It failed safe rather than silently — the encode runs before any write, so
+  stdout stayed empty instead of half-written — but `diagnose --json` still died on a machine
+  where `--out FILE` would have worked. `machine.emit_document` now switches stdout to UTF-8
+  before writing. Re-serializing the document as escaped ASCII would have been the smaller
+  change and the wrong one: the leak check verified the unescaped bytes, and emitting anything
+  re-derived from them is what that helper exists to prevent. `--out FILE` was never affected;
+  it has always written `encoding="utf-8"`.
+
 - **Leak self-check now matches JSON-escaped values (#195).** Two evasions became reachable
   the moment `diagnose --json` stopped also rendering the markdown report, since that raw-text
   pass was what had been catching them: `json.dumps` doubles backslashes, so a Windows home
--- a/src/bmad_loop/machine.py
+++ b/src/bmad_loop/machine.py
@@ -52,12 +52,19 @@ Two ways in, by what the caller already holds:
 
 :func:`write_document` is the ``--out`` sibling of :func:`emit_document`, taking
 the same already-serialized string to a file instead of stdout.
+
+The serializer flags differ by family on purpose and are not to be unified: the
+renderers behind :func:`emit_document` sort their keys (a diff-stable dump is
+worth more than field order there) while :func:`emit`'s dicts are built in the
+order they are meant to be read, and ``diagnostics.render_json`` alone passes
+``ensure_ascii=False`` because its leak guard has to scan the values unescaped.
 """
 
 from __future__ import annotations
 
 import argparse
 import json
+import sys
 from pathlib import Path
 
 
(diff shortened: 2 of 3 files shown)


pbean:
@coderabbit review

PR #216, opened 2026-07-20
refactor(documents): make the document builders public (#212)

### LINKED ISSUE #212: documents.py: drop the leading underscore from the document builders before a non-CLI consumer depends on them
Follow-up from a CodeRabbit nitpick on #211, deliberately not addressed there (see below).

## The tension

#211 created `src/bmad_loop/documents.py` as the library-level read-model projection layer, with a docstring that says, explicitly, that a future non-CLI frontend (the planned web backend) imports these builders directly. But every builder kept the leading underscore it had as a private function inside `cli.py`:

```
_validate_document   _decisions_document   _status_document
_list_document       _cleanup_document     _clean_document      _run_token_totals
```

A leading underscore conventionally means "internal, do not import from outside this module," which is the opposite of what the module is for. An external consumer writing `from bmad_loop.documents import _status_document` is reaching into privates by every convention a reviewer or linter applies.

## Why the underscore is the odd one out

The precedent this layer is modeled on argues the same way. `probe.py` and `diagnostics.py` — the "one finding, two render targets" split `documents.py` generalizes — name their whole projection surface public:

| module | public surface |
| --- | --- |
| `probe.py` | `SCHEMA_VERSION`, `scan`, `probe`, `discover_transcript`, `infer_token_schema`, `render_markdown`, `render_json` |
| `diagnostics.py` | `SCHEMA_VERSION`, `collect`, `collect_run`, `collect_env`, `summarize_journal`, `summarize_files`, `render_json`, `render_markdown` |

Not one underscore between them. The six `*_SCHEMA_VERSION` constants in `documents.py` are already public; only the builders are not, and that split has no rationale behind it — it is just what the names happened to be when they lived in `cli.py`.

## Why not in #211

That PR's entire deliverable is a *provable* pure relocation: the bodies were verified byte-identical by AST extraction against `git show HEAD:src/bmad_loop/cli.py`, and the pre-existing tests passed untouched as the proof. Renaming in the same commit would have voided both, and would have broken the requirement that `cli._x_document` keep resolving. Separable change, separate PR.

## Scope

Small and entirely in-tree — worth doing *before* a consumer exists, not after:

- Drop the underscore on the seven functions in `documents.py`.
- Update the call sites and the re-import block in `cli.py`. Note the re-imported names would then no longer need `# noqa: F401` for the *builders* (they are referenced in `cli.py`), but the six `*_SCHEMA_VERSION` constants still do — that noqa is load-bearing, see the comment above the import block.
- Decide whether `cli._x_document` aliases stay for backwards compatibility. Nothing in-tree needs them (no test calls a builder directly — verified), so probably not; they would just re-create the private surface the rename is removing.
- `_run_token_totals` is the one genuine judgement call: it is also used by `cmd_status` for *text* rendering, so it is arguably a shared helper rather than part of the document API. Either name it publicly alongside the rest, or leave it private and accept that `_status_document` calls a private sibling within its own module (which is fine — that is what the underscore is actually for).

Blocks nothing, but should land before #210's successor work or any web backend starts importing from here.

pbean:
Closes #212.

`documents.py` is the library-level read-model projection layer, and its docstring says outright that a non-CLI frontend (the planned web backend) imports these builders directly. Every builder nonetheless kept the leading underscore it had as a private function inside `cli.py` — which says the opposite. The six `*_SCHEMA_VERSION` constants beside them were already public; the split had no rationale, it was an artifact of the origin.

Deferred out of #211 because that PR's deliverable was a *provable verbatim relocation* (AST-identical bodies, untouched tests as the proof) — renaming in the same commit would have voided both.

## The one judgement call

`run_token_totals` goes public too. The issue framed this as open, on the grounds that it is a shared helper rather than document API and could stay "a private sibling within its own module." It cannot: `cmd_status` imports it **across the module boundary** (`cli.py:53`) and calls it on the *text* path (`cli.py:1581`), after the `--json` early return. A private name imported by another module is the exact contradiction being removed.

The `run_` prefix stays — it marks run-level vs per-task aggregation, which is the distinction the function exists to get right.

## Decisions

- **No back-compat `cli._x_document` aliases.** Nothing in-tree needs them and they would re-create the private surface this removes. Same call the Unreleased notes already record for `runs.tmux_sessions` → `mux_sessions` ("internal, no deprecation aliases").
- **No `__all__`.** No leaf module in the package has one, including `probe.py` and `diagnostics.py` — the modules this one is modelled on. Adding one here alone would make the absence there newly ambiguous.
- **No new CHANGELOG entry** (matches #211 — nothing user-visible changes: CLI surface, all six schema versions, and every emitted document are byte-identical). The Unreleased entry naming `_run_token_totals` *is* updated, since it is unshipped notes rather than history.

## New: the library path's first coverage

Every other `--json` test reaches the builders through `cli.main`, so nothing pinned the docstring's central promise — that the library path and the CLI path agree. Two parity tests drive one fixture down both and assert the same dict comes back.

Equality is asserted against the **raw builder return, not a json round-trip**: a consumer holds this dict before serializing it, so a tuple where a list belongs is a real defect a round-trip would hide. Verified by injecting exactly that (`"tasks": tuple(tasks)`) and confirming the test fails, then reverting.

Scoped to `status`/`list`, which have clean deterministic fixtures. `clean`/`cleanup` need mutation fixtures and `validate` needs a driven `ValidationReport` — those would balloon this past a rename.

## Notes for review

- `tests/conftest.py` was the surprise: `make_validate_document`, new in #215, already calls the builder directly — so an in-tree consumer was *already* reaching into a private name across a module boundary, which strengthens the issue's case. It needed the rename too.
- **`make_validate_document` contains `_validate_document` as a substring.** The rename is word-boundary anchored, so it and its 13 call sites in `test_tui_app.py` are untouched — verified by dry-run diff before applying.
- `tests/test_cli.py` has one docstring-only cross-ref fix; **zero assertion changes** to any pre-existing test, so #211's "the untouched tests are the proof" story survives.
- `cli.py:36-47` is deliberately byte-identical: all six `noqa: F401` are load-bearing (ruff's autofix silently deletes a re-export), and the comment above them stays accurate. Checked with `trunk check --no-fix`.

## Verification

- `2628 passed, 1 skipped`
- `trunk check --no-fix` — no issues
- Library surface imports with no CLI/argparse involved; `cli.*_SCHEMA_VERSION` re-exports still resolve; `noqa: F401` count still 6

## Summary by CodeRabbit

* **New Features**
  * Exposed reusable document-generation helpers for validation, decisions, status, listing, cleanup, and token totals.
* **Changed**
  * Standardized CLI JSON output generation across supported commands without changing the documented output format.
  * Clarified the machine-readable JSON contract and token-total calculation terminology in the changelog.
* **Tests**
  * Added checks confirming library-generated documents match the corresponding CLI `--json` output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -245,7 +245,7 @@ breaking changes may land in a minor release.
   from #190 — one JSON object on stdout, inline `schema_version`, additive-only evolution,
   errors → stderr with empty stdout — now live in one module with shared `emit`/`add_json_flag`
   helpers; `status --json` uses them (output byte-identical) and the duplicated token-total math
-  folded into `_run_token_totals`. All four `--json` commands share the contract (#195); `--json`
+  folded into `run_token_totals`. All four `--json` commands share the contract (#195); `--json`
   adoption on more commands is tracked in #196.
 - **Backend-neutral naming for the seam-backed helpers and operator messages.** The multiplexer
   seam has non-tmux backends now, so the helpers that wrap it drop their legacy tmux names —
--- a/src/bmad_loop/cli.py
+++ b/src/bmad_loop/cli.py
@@ -46,13 +46,13 @@ from .documents import LIST_SCHEMA_VERSION  # noqa: F401 — re-export
 from .documents import STATUS_SCHEMA_VERSION  # noqa: F401 — re-export
 from .documents import VALIDATE_SCHEMA_VERSION  # noqa: F401 — re-export
 from .documents import (
-    _clean_document,
-    _cleanup_document,
-    _decisions_document,
-    _list_document,
-    _run_token_totals,
-    _status_document,
-    _validate_document,
+    clean_document,
+    cleanup_document,
+    decisions_document,
+    list_document,
+    run_token_totals,
+    status_document,
+    validate_document,
 )
 from .engine import Engine
 from .journal import Journal, load_state, save_state
@@ -504,7 +504,7 @@ def cmd_validate(args: argparse.Namespace) -> int:
     if getattr(args, "json", False):
         # getattr, not args.json: cmd_validate is called directly by tests (and by
         # anything holding a hand-built Namespace) that predate the flag.
-        machine.emit(_validate_document(report, stories_on, spec_folder))
+        machine.emit(validate_document(report, stories_on, spec_folder))
     else:
         report.render()
     return 0 if report.passed else 1
@@ -1521,7 +1521,7 @@ def cmd_decisions(args: argparse.Namespace) -> int:
         # document, not the text line), and regardless of --list: --json *is*
         # the listing. It cannot fall through to the prompter, which reads stdin
         # and prints per-answer progress — neither survives a pure document.
-        machine.emit(_decisions_document(pending))
+        machine.emit(decisions_document(pending))
         return 0
     if not pending:
         print("no unanswered decisions from past sweeps")
@@ -1568,7 +1568,7 @@ def cmd_status(args: argparse.Namespace) -> int:
         return 1
     state = load_state(run_dir)
     if args.json:
-        machine.emit(_status_document(state))
+        machine.emit(status_document(state))
         return 0
     kind = f" [{state.run_type}]" if state.run_type != "story" else ""
     print(f"run {state.run_id}{kind}  started {state.started_at}")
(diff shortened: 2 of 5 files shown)


pbean:
@coderabbit review

Issue #222, opened 2026-07-21
psmux 3.3.8: adopt as installed floor, re-verify the switch-client/select-window paths, and drop the psmux#497 and other workarounds


dracic:
## Summary

This is the single **"when the next psmux tag lands"** tracker for the psmux family under #288. Two separate upstream fixes are sitting on psmux `main` with no release carrying them — 3.3.7 (2026-07-20) is still the newest tag, and the only one our version gate admits:

| upstream | what it fixes | commits on `main` |
|---|---|---|
| psmux/psmux#483 | `switch-client` window/pane targets silently ignored (rc-0 no-op) | `d56d777` + `6c76ff9` |
| psmux/psmux#497 | `select-window -t sess:@N` rejected CLI-side; **and** the server focusing `@N` as *index* on ≥3-window sessions | `33ea109` |

Everything below is what changes for us the day a tag ships both. Sections are split by what each item is actually gated on — the tag alone, the tag plus a code removal here, or something no release will ever fix.

---

## A. Gated on the tag only — live smoke, real attached client, Windows 11

Several forms we rely on (or plan to, see #221) were never live-verified because 3.3.7 cannot exercise them.

- [ ] `switch-client -t %N` **same-session**: the parked-window trailer's return move now actually lands on the originating pane (maintainer verified; confirm through our trailer end-to-end)
- [ ] Unresolvable target exits **non-zero** → the trailer's `switch-client -l` fallback is now reachable; confirm the fallback itself works
- [ ] `switch-client -l` semantics across psmux's per-session servers: does "last session" survive the server hop at all? (never live-tested; both trailers and `switch_client(last_fallback=True)` depend on it)
- [ ] Cross-session `sess:win` selection: TUI/CLI attach to a control-session window from inside psmux (`switch-client -t =bmad-loop-ctl:<win>`) lands on **that** window, not the session's active one (fixed upstream in `6c76ff9`; on 3.3.7 the window part is ignored)
- [ ] `=` exact-match prefix in a qualified target (`=sess:win`): does the new dedicated switch-client handler parse it? (untested upstream form; upstream tests use bare `sess:win`)
- [ ] Whichever return-target form #221 lands (`sess:idx.idx` or `sess:%N`): verify it live cross-session (ctl session → user session)
- [ ] `switch_client()` seam no longer returns a false `True` on a no-op (rc-0 trust — self-heals with the release, confirm)
[...]

PR #225, opened 2026-07-21
fix(tui): record session-qualified parked-window return target

### LINKED ISSUE #221: Parked-window return target is a bare pane id — unresolvable cross-session on psmux (one server per session)
## Summary

The attach return path records a **bare pane id** (`%N`) and later replays it as `switch-client -t %N` from inside the control session. That assumes tmux's one-server model, where pane ids are globally unique and resolvable from anywhere. psmux runs **one server per session**, and upstream has now stated this as a permanent model boundary (psmux/psmux#483, final comment): a bare `%pane`/`@window` id is resolved **only within the current session** — an unqualified id coming from another session is ambiguous and **can collide with a real pane of the current session**.

For us the collision case is worse than a no-op: the trailer runs inside the control session, so a colliding `%N` switches the client to the **wrong control-session pane** with exit 0 — the `switch-client -l` fallback never fires because nothing failed.

## Where the assumption lives

Recording side (both store the TUI/CLI client's own pane — a pane of the **user's** session — as a window option on a **control-session** window):
- `src/bmad_loop/tui/app.py` — `_attach_to_target` records `launch.current_pane_id()` via `set_return_pane`
- `src/bmad_loop/cli.py` — the `attach` command does the same
- `src/bmad_loop/tui/launch.py` — `RETURN_OPTION` / `set_return_pane` / `current_pane_id`

Replay side (all issue `switch-client -t <recorded value>` from control-session context):
- `src/bmad_loop/adapters/tmux_base.py` — `_parked_trailer` (POSIX) and `switch_client`
- `src/bmad_loop/adapters/psmux_backend.py` — `_parked_trailer` (pwsh)
- `src/bmad_loop/tui/launch.py` — `return_attached_client`

## Upstream status

psmux/psmux#483 is fixed on psmux main (d56d777 + 6c76ff9, **not in any tagged release** — 3.3.7 still silently ignores window/pane targets with exit 0). Post-fix behavior, verified by the maintainer:
- full `sess:win.pane` / `@win` / `%pane` targets switch **and** select — including cross-session, where the addressed window/pane selection is carried across the server hop;
- unresolvable targets exit non-zero (so our `-l` fallback finally becomes reachable);
- bare ids stay **session-local by design** — this will not change, so the fix must be on our side.

## Candidate fixes (investigate, pick one)

**A) Record a fully index-qualified target** — `#{session_name}:#{window_index}.#{pane_index}` instead of `#{pane_id}`. This is exactly the form the maintainer verified cross-session (`dstb:1.1`). Standard tmux target syntax, so the tmux path keeps working unchanged and no per-dialect branching is needed. Trade-off: indexes are positional, not stable ids — a window moved/renumbered between attach and return drifts the target (the non-zero exit + `-l` fallback then degrades gracefully).

**B) Record `#{session_name}:#{pane_id}`** — keeps the stable id. **Unverified on both sides**: whether tmux and psmux's new dedicated handler accept a pane id in the window position of a qualified target needs probing before committing.

Either way `RETURN_DETACH` ('detach' sentinel) still never collides with a recorded value.

## Constraints

- The tmux backend must keep working with the new recorded form — a single seam-level recording change beats per-backend branching if the form parses on both.
- On psmux 3.3.7 (the only release passing the version gate today) even a correct target form is a silent no-op — that stays broken by design until the next psmux release and is out of scope here; correctness is verifiable under tmux plus mocked tests now, live psmux confirmation is tracked separately.

Refs: psmux/psmux#483, psmux/psmux#202, #185

dracic:
## Summary

An interactive attach recorded the client's origin as a bare pane id (`%N`) and replayed it as `switch-client -t %N` from inside the control session. Sound under tmux's one-server model, but on psmux (one server per session, upstream-final per psmux/psmux#483) a bare id is session-local: at best unresolvable, at worst colliding with a real control-session pane and landing the client on the wrong one with exit 0 — past the `switch-client -l` fallback.

This is candidate B from the issue thread: the recording seam (`tui/launch.py`) now emits `=session:%N` when both probes succeed, retaining bare `%N` only if the session-name probe fails (a resolvable own pane means the client IS inside the multiplexer, so recording `detach` would strand it). An rc-0-but-empty pane probe is rejected before composing. The replay sides — POSIX and pwsh trailers, `switch_client`, `return_attached_client` — treat the value as an opaque target and are unchanged beyond comment accuracy. `RETURN_DETACH` stays collision-free.

## Verification

- tmux 3.4 (WSL): `=sess:%N` parses and resolves to the exact pane; `switch-client` accepts the form. tmux resolves the qualified form to the same pane a bare id always reached.
- psmux side is source-verified against psmux main (d56d777 + 6c76ff9, see the issue thread); live confirmation stays on the #222 checklist — psmux 3.3.7 (the only release passing the version gate) no-ops all window/pane targets by design, so nothing regresses there.
- Windows targeted suite (launch/cli/tui-app/multiplexer/psmux): 457 passed, 1 known baseline failure (skill-sync drift). Full WSL suite: failures identical to a clean checkout of main (stash-compared), none introduced.
- Recording is covered by probe-composition tests (qualified, outside-mux, transport failure, empty pane, failed/empty session name); the replay fixture now round-trips the qualified value.

Closes #221

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed return-pane selection after interactive attaches to use session-qualified return targets when available.
  * Avoided cross-session misrouting when pane identifiers overlap between sessions.
  * Kept safe fallback behavior when session information can’t be determined reliably, ensuring correct tmux/psmux return from parked windows.
* **Documentation**
  * Updated the adapter authoring guide with clarified parked-window return target and pane identifier formatting rules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -298,16 +298,19 @@ PATH)`, the TUI notifies `multiplexer backend unavailable — launch/attach disa
 
 ### Fixed
 
-- **The parked-window return target is now session-qualified (#221).** An interactive attach
+- **The parked-window return target is now backend-composed (#221).** An interactive attach
   recorded the client's origin as a bare pane id (`%N`) and replayed it as `switch-client -t %N`
   from inside the control session — sound under tmux's one-server model, but on psmux (one
   server per session, upstream-final per psmux/psmux#483) a bare id is session-local: at best
   unresolvable, at worst colliding with a real control-session pane and landing the client on
-  the wrong one with exit 0, past the `switch-client -l` fallback. The recording seam now emits
-  `=session:%N` when both probes succeed, retaining bare `%N` only if the session probe fails;
-  the replay sides treat the value as an opaque target and are unchanged. tmux resolves the
-  qualified form to the same pane it always did (verified on 3.4), and psmux releases carrying
-  the psmux/psmux#483 fix resolve it cross-server.
+  the wrong one with exit 0, past the `switch-client -l` fallback. No single form resolves on
+  every backend (tmux's window resolver rejects a pane id in the `session:%N` slot, and a
+  native-id backend needs its own id passed through untouched), so the recording seam now asks
+  the backend: `TerminalMultiplexer.current_return_target()` defaults to the bare native pane
+  id — tmux and native-id backends behave exactly as before — and psmux overrides it to emit
+  `=session:%N`, which releases carrying the psmux/psmux#483 fix resolve cross-server,
+  degrading to the bare id only if the session probe fails. The replay sides treat the value
+  as an opaque target and are unchanged.
 
 - **A dry run the TUI cannot spawn opens a modal instead of taking the app down (#210).** The
   `run --dry-run` / `sweep --dry-run` workers called `run_captured` unguarded, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -298,6 +298,17 @@ PATH)`, the TUI notifies `multiplexer backend unavailable — launch/attach disa
 
 ### Fixed
 
+- **The parked-window return target is now session-qualified (#221).** An interactive attach
+  recorded the client's origin as a bare pane id (`%N`) and replayed it as `switch-client -t %N`
+  from inside the control session — sound under tmux's one-server model, but on psmux (one
+  server per session, upstream-final per psmux/psmux#483) a bare id is session-local: at best
+  unresolvable, at worst colliding with a real control-session pane and landing the client on
+  the wrong one with exit 0, past the `switch-client -l` fallback. The recording seam now emits
+  `=session:%N` when both probes succeed, retaining bare `%N` only if the session probe fails;
+  the replay sides treat the value as an opaque target and are unchanged. tmux resolves the
+  qualified form to the same pane it always did (verified on 3.4), and psmux releases carrying
+  the psmux/psmux#483 fix resolve it cross-server.
+
 - **A dry run the TUI cannot spawn opens a modal instead of taking the app down (#210).** The
   `run --dry-run` / `sweep --dry-run` workers called `run_captured` unguarded, and
   `@work(thread=True)` defaults to `exit_on_error=True` — so an `OSError` from the spawn itself
--- a/docs/adapter-authoring-guide.md
+++ b/docs/adapter-authoring-guide.md
@@ -97,7 +97,11 @@ families: the **seam-canonical target token** `=session[:window]` — formatted 
 the concrete `TerminalMultiplexer.target(session, window=None)`, decoded by the
 module-level `parse_target()` — or the backend's own **native id** (whatever your
 `new_window` returned). Core never hand-assembles the grammar; it calls
-`target()`. tmux consumes the token natively (it coincides with tmux exact-match
+`target()`. One caller stretches the shape: the parked-window return target is
+formatted as `target(session, pane_id)` — a _pane id_ (`%N`) rides in the
+window slot — so a backend resolving `parse_target()` output must tolerate a
+pane id there when it reaches `switch_client`. tmux consumes the token
+natively (it coincides with tmux exact-match
 syntax), so `BaseTmuxBackend` passes it straight through. A native-id backend
 calls `parse_target()` first — `None` means "already a native id, use as-is",
 otherwise resolve `(session, window)` yourself; the herdr adapter's
--- a/docs/adapter-authoring-guide.md
+++ b/docs/adapter-authoring-guide.md
@@ -88,8 +88,13 @@ seams of a full OS port are in
   `send_text`.
 - **Client / attach** — `attach_target_argv` (argv that reaches a target, nesting-
   aware), `current_pane_id` / `current_window_id` / `current_session`,
-  `detach_client`, `switch_client` (with an optional last-client fallback),
-  `available` (is this backend usable on the current host).
+  `current_return_target` (the value an interactive attach records for the
+  parked-window return hop, replayed opaquely by `switch_client` and the parked
+  trailer; concrete default = the native pane id, so most backends inherit it —
+  override only when your ids do not resolve from another session's context,
+  as psmux does to emit `=session:%N`), `detach_client`, `switch_client` (with
+  an optional last-client fallback), `available` (is this backend usable on
+  the current host).
 
 **Window targets.** The target-taking methods (`kill_window`, `select_window`,
 the window-option trio, `attach_target_argv`, `switch_client`) receive one of two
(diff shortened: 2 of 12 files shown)


dracic:
Ran a multi-aspect review of this branch (general correctness, test coverage, comment accuracy, silent-failure audit). Verdict: the core change is sound — all degradation branches of `current_return_target()` are covered by tests at the `subprocess.run` seam, the replay sides genuinely treat the recorded value as opaque, and no caller of the old `launch.current_pane_id()` was missed. The `=` prefix is also a quiet correctness upgrade: it forces exact session match where tmux would otherwise pattern-match.

One finding was addressed in a0ec7a1:

- **Seam violation:** `current_return_target()` hand-assembled `=session:%N` with an f-string, contradicting the adapter guide's contract that core never formats the target grammar itself, and bypassing backend `target()` overrides. Now routed through `mux.target(session, pane)` (byte-identical output for the tmux family), with a guide note that this caller carries a pane id in the window slot. A few comment-precision fixes rode along (the `RETURN_OPTION` block now names the bare-`%N` degradation; the parked-trailer comment distinguishes the option name from its recorded value).

Verified safe (no action): session names cannot contain `:`/`.` under tmux so the qualified form cannot misparse; a session literally named `detach` yields `=detach:%N`, which the trailer's equality test cannot match; the rc-0-empty-stdout probe answers degrade identically to rc!=0 and both are pinned by tests.

Two pre-existing adjacent issues surfaced by the audit are filed separately as follow-ups (ignored `switch_client` result in `return_attached_client`, and the pre-#483-fix psmux rc-0 no-op having no version gate) — they are out of this PR's scope.

`ruff check` clean; affected test files pass on Windows (434 passed, 1 known skill-sync-drift baseline failure unrelated to this diff).

pbean:
@coderabbit review

pbean (changes requested):
Ran this through a full review with live verification: scratch tmux server with a real pty-attached client, tmux source at 3.4 and master, psmux `cli.rs` at master, and the herdr adapter source. Verdict: **the recording-side structure is right, but `=session:%N` regresses the tmux path this PR promises to leave intact, and breaks the herdr adapter's return hop.** Since maintainer edits are enabled I'll push a fix commit onto this branch (backend-owned composition — sketch at the end) rather than bounce it; review of that commit welcome.

### 1. tmux rejects `=session:%N` at the replay verb

```
$ tmux -V
tmux 3.7b
# scratch -L server: sessions ctl + main, real client attached to ctl via a pty
$ tmux -L S switch-client -t "=main:%1"
can't find window: %1        # rc=1, client does not move
$ tmux -L S switch-client -t "%1"
                             # rc=0, client lands on main:%1 (the pre-PR form)
```

Source-confirmed in 3.4 and master (`cmd-find.c`): with a colon and no period, `cmd_find_target` puts the post-colon part in the *window* slot unconditionally — the find type never matters there — and `cmd_find_get_window_with_session` has no `%` handling at all: it tries `@` ids, `+`/`-` offsets, `!`/`^`/`$`, numeric index, then exact/prefix/fnmatch name matching, then returns `-1` → `no_window:` → `can't find window`. The pane-id checks exist only in the pane helpers (reached via a `.`-separated pane part or the no-separator branch).

Consequence: **every** parked-window return on tmux now exits 1 and rides the `switch-client -l` fallback — origin session usually recovered, exact-pane return gone — and the client hard-strands in the control session whenever `-l` has nothing to fall back to (e.g. the client was attached directly into the ctl session, so it has no last session).

**Why the PR's tmux 3.4 verification passed anyway** — two independent false positives:

- `display-message -p -t '=main:%N' '#{pane_id}'` answers the right pane, but not because the target resolves. Its target is declared `{ 't', CMD_FIND_PANE, CMD_FIND_CANFAIL }`, and `cmd_find_target`'s `error:` path under CANFAIL `return (0)` with the *partially-filled* state: the session half already resolved, and `cmd_find_get_window_with_session` seeds `fs->wl = fs->s->curw` **before** failing. So the answer is "the session's current pane" — indistinguishable from correct resolution precisely when the probed pane is that session's active pane, i.e. exactly the by-hand test from your own pane. `switch-client` (`flags = 0`) takes the strict `return (-1)` and errors. Reproduced both against the same server and target string.
- End-to-end, the `-l` fallback usually lands near the right place (the origin session's curw is where the user left), masking the rc=1.

CI stayed green because every touched test scripts `subprocess.run` with rc=0 — nothing exercises real target resolution.

### 2. herdr's return hop breaks

herdr's `current_pane_id()` is a native id (`w1:p1`). This branch records `target(label, "w1:p1")` = `=label:w1:p1`; herdr `switch_client` → `_parse_target` → `multiplexer.parse_target` yields `("label", "w1:p1")` and then looks for a *tab labeled* `w1:p1` → `None` → silent `False`. Today's bare native id passes `parse_target` untouched (no leading `=`) and works. The guide note added here ("a backend … must tolerate a pane id in the window slot") is a contract change the shipped herdr adapter does not satisfy.

### 3. No uniform composition exists

- `=sess:%N` — precise on psmux (`parse_target`: post-colon `%N` sets `pane_is_id`), hard error on tmux (above).
- `=sess:.%N` — precise on tmux (verified live: rc=0, exact pane — the empty window slot routes `%N` through `cmd_find_get_pane_with_session`, which does handle ids), but psmux silently drops the pane: `.%N` goes down the dot-split path and `"%N".parse::<usize>()` fails, leaving session-only.
- bare `%N` — precise on tmux and herdr, broken on psmux (#221 itself).

So the issue's constraint — "a single seam-level recording change beats per-backend branching **if the form parses on both**" — fails its own premise. The composition has to be backend-owned.

### Fix being pushed

The recorded value is written to and replayed from the *same* backend's server, so the backend composes it: `TerminalMultiplexer.current_return_target()` gains a concrete default (`self.current_pane_id() or None`) — tmux and herdr byte-identical to pre-PR behavior with zero changes on their side — and `PsmuxMultiplexer` overrides it to emit `=sess:%N` (the form your psmux source read verified), degrading to bare `%N` when the session probe fails or answers with a `:` in the name (unchanged rationale: a resolvable own pane means we're inside the mux, so `None`→detach would strand the client). `launch.current_return_target()` becomes a thin delegate and the replay sides stay fully opaque — which preserves everything this PR got right: the rename, the empty-probe degradation guards and their tests, and the `RETURN_DETACH` non-collision reasoning.

pbean:
Pushed 0a427e5 implementing the backend-owned composition from the review, on top of your two commits.

What it does:

- `TerminalMultiplexer.current_return_target()` — new concrete seam method, default `self.current_pane_id() or None`. tmux and native-id backends (herdr) record exactly what they recorded before #225: the bare native pane id, which their own `switch-client` resolves precisely. The `or None` keeps your rc-0-empty-pane guard at the seam.
- `PsmuxMultiplexer.current_return_target()` — the qualified composition moved here: `=session:%N` via `target()`, degrading to the bare id when the session probe fails, answers empty, or answers a name containing `:` (which `parse_target`'s first-colon split couldn't carry). Your degradation rationale and comments are preserved verbatim where they still apply.
- `launch.current_return_target()` is now a thin delegate; `cli.py` / `tui/app.py` call sites and your rename are untouched. Replay sides (`_parked_trailer` sh + pwsh, `switch_client`, `return_attached_client`) stay fully opaque — no code changes, comments reworded from "targets are `=session:%N`" to "backend-composed".
- Adapter guide: the "pane id rides in the window slot" tolerance note is replaced by documenting the optional `current_return_target` override — the seam grammar never has to carry a pane id now, so out-of-tree backends need no changes.
- Tests: your probe-composition/degradation tests moved to `test_psmux_backend.py` (plus a session-name-with-colon case); the launch-level test pins the tmux default back to bare `%N` **and asserts the session probe is never issued there**; your qualified-value round-trip through `_return_fake` is kept — the replay must handle any recorded form.

Verification: full suite 2681 passed / 1 known skip; `trunk check --no-fix` clean on all 12 touched files. Live on tmux 3.7b (scratch `-L` server): `launch.current_return_target()` from inside a real pane answers `%0`, and `switch-client -t %0` with a pty-attached client returns it to the exact pane, rc 0 — the pre-#225 behavior, restored. The psmux side keeps your source-verified `=session:%N` form; live confirmation stays on the #222 checklist as before.

pbean:
@coderabbit review

PR #233, opened 2026-07-21
fix: report worktree_seed entries skipped as no-ops

### LINKED ISSUE #230: A `worktree_seed` directory entry is skipped entirely when any child is tracked
### Description

`provision_worktree` treats a seed entry as one unit: if the destination path exists in the
fresh worktree, the whole entry is skipped. For a **file** that is exactly right, and
`test_provision_worktree_seed_does_not_clobber_existing` states the reason plainly — *"so no
diff is merged back"*.

For a **directory** the same rule is over-broad. A worktree checks out tracked files, so a
seed directory with even one tracked child already exists in the worktree — and the entry is
then skipped whole, including the many children that are **not** present and would clobber
nothing.

Concretely: `worktree_seed = ["_bmad"]`, where `_bmad/custom/` is tracked and the rest of
`_bmad/` is gitignored. That is the natural value to write, and it copies **nothing**, because
`_bmad/` exists the moment `_bmad/custom/` is checked out.

### Steps to reproduce

1. A project with `scm.isolation = "worktree"`.
2. A directory `foo/` that is gitignored except for one tracked child — `foo/tracked/` is
   committed, `foo/ignored/` is not.
3. `.bmad-loop/policy.toml`:
   ```toml
   [scm]
   isolation = "worktree"
   worktree_seed = ["foo"]
   ```
4. `bmad-loop run --story <key>`
5. `ls -la .bmad-loop/runs/<run-id>/worktrees/<story>/foo/`

### Expected behavior

`foo/ignored/` is seeded into the worktree. It exists in the repo, is absent from the worktree,
and copying it clobbers nothing — which is the stated condition for seeding.

### Actual behavior

Only `foo/tracked/` is present. `foo/ignored/` was never copied, and nothing reports it: no
stderr warning, no journal event, nothing in `validate`.

`install.py`, `provision_worktree`:

```python
for rel in seed_files:
    src = (repo_root / rel).resolve()
    dst = (worktree / rel).resolve()
    ...
    if not src.exists() or dst.exists():
        continue
```

Reached from the engine's unit-provisioning step, where `scm.worktree_seed` is appended to the
seed list handed to `provision_worktree`. The local `seeded` list is used only to build the
worktree's git-exclude patterns, so a skipped entry leaves no trace anywhere.

*(Symbols rather than line numbers throughout — I originally wrote this against 0.8.1 and the
lines have already moved on main.)*

### Why the no-clobber rationale doesn't cover this

Two reasons the "no diff is merged back" argument doesn't extend from files to a directory's
missing children:

1. Those children don't exist in the worktree, so copying them overwrites nothing and produces
   no diff against any tracked file.
2. The seeded paths are added to the worktree's local git-exclude by `_worktree_local_exclude`
   immediately afterwards, so they're kept out of the unit's `git add -A` regardless.

`_copy_traversable` is already recursive and already does `dst.mkdir(..., exist_ok=True)` — the
directory case is only blocked by the top-level guard.

### Why it's worth fixing rather than documenting

The failure is invisible and it hides its own cost. In my case the isolated sessions detected
the missing tree and re-created it themselves, once per story, so every run stayed green while
paying for the repair each time. I found it only by opening a worktree for an unrelated reason
and noticing symlinks where seeded directories should have been real ones. There was no bad
outcome to investigate — only a slow leak.

### Two possible fixes

**A — warn only (minimal, no behaviour change).** Emit a warning + journal event when a seed
entry is skipped because the destination exists. The configuration stays a no-op, but it stops
being a *silent* one.

**B — recurse with copy-when-absent per child.** When `src` is a directory and `dst` exists,
walk into it and copy only the children that are absent. This preserves the no-clobber rule at
the level where it actually matters. Note `_copy_traversable`'s file branch currently uses
`copy2` unconditionally, so it would need a per-child `dst.exists()` check to keep that
guarantee.

B is what I'd want as a user, but it is a behaviour change and A is uncontroversial — happy to
open a PR for either, or both, whichever you prefer.

### Confirm

Polloinfilzato:
## What

`provision_worktree` now returns the `seed_files` entries it skipped because the destination
already existed, and the engine journals them as `worktree-seed-skipped`.

No behaviour change — nothing new is copied.

## Why

Fixes #230.

Copy-when-absent is right for a file the checkout legitimately carries, and
`test_provision_worktree_seed_does_not_clobber_existing` states the reason: no diff is merged
back.

For a **directory** entry the same rule is over-broad. A worktree checks out tracked files, so a
seed directory with any tracked child already exists — and the entry is then skipped whole,
including the children that are absent and would clobber nothing. `worktree_seed = ["_bmad"]`
with `_bmad/custom` tracked is the natural value to write, and it copies nothing at all.

Nothing reported it: the local `seeded` list only feeds the git-exclude patterns, so the skip
left no trace in stdout, the journal, or `validate`. The configuration reads as applied and is
not. In my case the isolated sessions re-created the missing tree themselves, once per story, so
every run stayed green while paying for the repair.

## How

- `provision_worktree` gained a `list[str]` return: entries whose source existed but whose
  destination did too.
- The engine journals them under `worktree-seed-skipped`, alongside `worktree-opened`.
- **Reported, not printed.** The docstring is explicit that this function is quiet because it
  runs inside the engine loop under a TUI, so a `print` would corrupt the display — the journal
  is where this signal belongs, next to `worktree-open-failed` and `sprint-status-unknown-keys`.
- `not src.exists()` and `dst.exists()` are now separate branches. A missing source is a
  different case (already covered by its own test) and is not a no-op worth reporting.
- **Glob matches are deliberately not reported.** A plugin's `seed_globs` pattern is expected to
  hit paths the checkout already carries, so that skip is routine rather than a misconfiguration.
  The signal stays specific to hand-written `worktree_seed` entries.

I stopped at reporting rather than the alternative in #230 (recursing into an existing directory
with per-child copy-when-absent). That one changes behaviour and `_copy_traversable`'s file
branch would need a per-child `dst.exists()` guard to keep the no-clobber property — happy to
follow up with it if you'd prefer that shape.

## Testing

Three tests added in `tests/test_install.py`:

- a skipped file entry is reported;
- a seed **directory** with a tracked child is reported and its absent children stay absent —
  the case from the issue, pinned as current behaviour;
- a successful seed reports nothing, and a missing source is not reported either.

Existing seeding tests pass unchanged, including
`test_provision_worktree_seed_does_not_clobber_existing`, so the no-clobber guarantee is intact.

- `uv run pytest -q` → **2678 passed, 23 skipped**
- `trunk fmt` and `trunk check` → no issues

One unrelated note: `tests/test_engine.py::test_rollback_or_pause_resolved_auto_recovers` fails
on my machine because it matches git's English `"did not match"` while my git is localised. It
fails the same way on a clean `main` and passes under `LC_ALL=C`, so it is untouched by this
change — but the suite isn't locale-independent. Happy to open that separately if it's useful.

---

Written with AI assistance, reviewed line by line before submitting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved worktree setup reporting when seed files are skipped because destination paths already exist.
  * Added journal events identifying skipped seed entries for better visibility.
  * Ensured directory seeds are handled consistently without copying unintended files.

* **Tests**
  * Added coverage for skipped, copied, directory, and missing seed-file scenarios.

### DIFF
--- a/src/bmad_loop/engine.py
+++ b/src/bmad_loop/engine.py
@@ -598,13 +598,22 @@ class Engine:
         # config so the worktree's Editor MCP is reachable. Aggregate every loaded
         # plugin's declared seeds.
         seeds.extend(self._registry.seed_files())
-        provision_worktree(
+        skipped_seeds = provision_worktree(
             unit.path,
             profiles,
             self.paths.repo_root,
             seed_files=list(dict.fromkeys(seeds)),  # dedupe, preserve order
             seed_globs=self._registry.seed_globs(),
         )
+        if skipped_seeds:
+            # A seed entry whose destination already exists is a no-op. Harmless for
+            # a file the checkout legitimately carries, but a directory entry is
+            # skipped WHOLE the moment any child is tracked — so a `worktree_seed`
+            # that looks applied can be copying nothing. Journal it; provision is
+            # quiet by contract (it runs under the TUI).
+            self.journal.append(
+                "worktree-seed-skipped", story_key=task.story_key, entries=skipped_seeds
+            )
         self.journal.append(
             "worktree-opened", story_key=task.story_key, branch=unit.branch, path=str(unit.path)
         )
--- a/src/bmad_loop/install.py
+++ b/src/bmad_loop/install.py
@@ -486,7 +486,7 @@ def provision_worktree(
     repo_root: Path,
     seed_files: Sequence[str] = (),
     seed_globs: Sequence[str] = (),
-) -> None:
+) -> list[str]:
     """Make a freshly-created git worktree a self-sufficient bmad-loop project.
 
     A worktree checks out tracked files only, but the skill trees (.claude/skills,
@@ -519,9 +519,14 @@ def provision_worktree(
     seed_files are copied BEFORE the hook step so a seeded settings file that is
     also a hook config_path (.claude/settings.json, .gemini/settings.json) keeps its
     real content and just gets the Stop hook merged in, rather than being created empty.
+
+    Returns the `seed_files` entries that existed in the repo but were skipped
+    because the destination already existed — copy-when-absent turned them into
+    no-ops. The caller journals them: a user-authored `worktree_seed` entry that
+    silently copies nothing reads as applied configuration and is not.
     """
     if not profiles and not seed_files and not seed_globs:
-        return
+        return []
     worktree = worktree.resolve()
     repo_root = repo_root.resolve()
     relay = repo_root / HOOK_SCRIPT_REL
@@ -530,12 +535,24 @@ def provision_worktree(
     # project gitignored MCP/CLI configs: copy from the main repo when absent.
     # Resolve-and-contain guards against an `..`/absolute entry escaping either tree.
     seeded: list[str] = []
+    # Entries that named a real source but whose destination already exists, so
+    # copy-when-absent made them a no-op. Reported to the caller (this function is
+    # quiet by contract — it runs under a TUI) because for a DIRECTORY entry the
+    # no-op is silent and total: a worktree checks out tracked files, so a seed dir
+    # with any tracked child always exists and the whole entry is skipped, including
+    # the children that are absent and would clobber nothing. Glob-expanded matches
+    # are deliberately not reported: a plugin's glob is expected to hit paths the
+    # checkout already carries, so that skip is routine rather than a misconfiguration.
+    skipped: list[str] = []
     for rel in seed_files:
         src = (repo_root / rel).resolve()
         dst = (worktree / rel).resolve()
         if not src.is_relative_to(repo_root) or not dst.is_relative_to(worktree):
             continue
-        if not src.exists() or dst.exists():
+        if not src.exists():
+            continue
+        if dst.exists():
+            skipped.append(str(rel))
             continue
         dst.parent.mkdir(parents=True, exist_ok=True)
         _copy_traversable(src, dst)
@@ -614,6 +631,7 @@ def provision_worktree(
     patterns |= {f"/{p.hooks.config_path}" for p in profiles if not p.hookless}
     patterns |= {f"/{rel}" for rel in seeded}
     _worktree_local_exclude(worktree, sorted(patterns))
+    return skipped
 
 
 def _warn_if_policy_tracked(project: Path) -> None:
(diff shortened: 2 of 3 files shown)


pbean (approved):
Validated line-by-line against main plus a sweep over every consumer surface — approving.

**What I checked beyond the diff:**

- The return-type change is inert everywhere else: `_run_isolated` is the only production caller, and no existing test consumes the return.
- The new journal kind can't break a consumer: there is no closed event-name enum anywhere — `diagnose` histograms kinds via an open `Counter`, the TUI journal view falls back to dim styling for unknown kinds, and the CLI resume readers only special-case their own kinds. No test pins an exact journal-kind sequence. In `diagnose`, `entries` routes through `sanitize.scrub_json` and relative seed paths don't trip the leak guard.
- No false-positive path: the one thing that could have made this event misleading is a re-provision of an already-seeded worktree reporting every entry as skipped. That can't happen — resume reconstructs the workspace via `_reopen_unit` without provisioning, and `open_unit_workspace` always mounts fresh (`git worktree add` refuses an existing path; that failure defers with `worktree-open-failed`). Every reported skip is genuine.
- Cross-platform: a pure `pathlib` branch split, and reporting the TOML-authored relative entries (rather than OS-native paths) is the right call. Full suite green locally (2694 passed) and on CI's Linux + real-Windows legs.

**Two non-blocking notes:**

1. The signal isn't quite "specific to hand-written `worktree_seed` entries": the seed list the engine passes also carries adapter-profile `seed_files` (under `scm.seed_adapter_defaults`, e.g. `.mcp.json` from the claude profile) and plugin seeds — so a project that *tracks* one of those files journals a skip for it once per story. Benign and honest; just noting the commit message's claim only holds vs. globs. Tagging entry origin could be a future refinement if this ever gets noisy.
2. Nothing load-bearing enumerates journal kinds, so no doc change is required. `docs/tui-guide.md`'s journal section is the natural home for a one-liner if you want one — equally happy to take that as a follow-up.

**On your two offers:** filing both as follow-up issues — one for the option-B shape from #230 (per-child copy-when-absent), which is yours if you'd like it; and one for the locale failure you hit. On the latter, it's better than you thought: the English match lives in production, not the test — `safe_rollback`'s preserve-restore tolerates the benign checkout failure by substring-matching `"did not match"`, so under a localized git that benign no-op raises `GitError` in the rollback path. Good catch.

Thanks for an unusually well-argued issue + PR pair.

PR #226, opened 2026-07-21
feat: add Hermes Agent adapter

SauliusDev:
## Summary
- add an interactive, attachable Hermes tmux adapter and bundled Hermes profile
- register the user-scoped Hermes `post_llm_call` relay and translate it into BMAD `Stop` events
- preserve application `config.yaml` files in isolated worktrees; Hermes hooks remain user-scoped
- validate Hermes YAML hook registration, including scan-only probe behavior for its user-scoped config

## Verification
- `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null uv run pytest -q` (2698 passed, 18 skipped)
- `uv run ruff check src tests`
- `uv build`
- live `hermes hooks test post_llm_call` relay check
- real tmux injection/relay integration test

Note: the default environment has a global `core.hooksPath` that bypasses two test-local rejecting hooks; the full suite is run with global Git config isolated so those tests exercise their intended local hooks.

## Summary by CodeRabbit

* **New Features**
  * Added support for running workflows through the Hermes interactive adapter.
  * Added a built-in Hermes profile with user-scoped hook configuration.
  * Added the `bmad-loop relay` command for forwarding Hermes events.
  * Added automatic Hermes hook registration while preserving existing hooks.
* **Bug Fixes**
  * Improved startup validation and error handling when interactive sessions exit unexpectedly.
  * Added safer handling for user-scoped probing and worktree configuration.

PR #238, opened 2026-07-21
feat(relay): shared event-write primitive + `bmad-loop relay <Event>`

pbean:
## What

Factor the relay event-write logic — **already duplicated** between `data/bmad_loop_hook.py` and PR #226's `hermes_hooks.relay_event` (identical env-gating, casing-tolerant field extraction, atomic `events/<ts>-<task>-<event>.json` write) — into **one core primitive** that owns the event schema, and add a general `bmad-loop relay <Event>` command over it.

This is the **general, in-tree counterpart to `bmad_loop_hook.py` for global/user-scoped-hook adapters** — NOT Hermes-specific. An out-of-tree adapter whose CLI registers a single user-scoped hook (rather than copying the per-project script) points that hook at `bmad-loop relay Stop` and needs **zero** per-adapter relay code.

## Changes

- **New `src/bmad_loop/events.py`** — `write_relay_event(event_name, payload, environ) -> Path | None` (+ `build_event`). Owns the canonical schema (`ts`, `event`, `task_id`, `session_id`, `transcript_path`, `cwd`) and the env-gated (`BMAD_LOOP_RUN_DIR` + `BMAD_LOOP_TASK_ID`), atomic write. No-ops outside an active bmad-loop session.
- **`bmad-loop relay <Event>`** in `cli.py` — reads the hook payload from stdin and defers to the primitive. `event` choices come from `CANONICAL_EVENTS` (never a hardcoded set).
- **`data/bmad_loop_hook.py`** — behavior unchanged. It **keeps its own inlined copy on purpose**: it is installed into target projects as a lone, stdlib-only file and run as a bare subprocess where the `bmad_loop` package need not be importable, so it cannot import the primitive. Added a docstring note documenting the deliberate duplication and the sync guard.

## Tests (`tests/test_relay_event.py`)

- primitive: active-env write, no-env / partial-env no-op, camelCase + `workspacePaths` fallbacks, unusable-`workspacePaths` degrades to `None`
- CLI stdin path, no-env no-op, garbage-stdin tolerance, non-canonical event rejected (parse error), every canonical event accepted
- **schema parity**: the standalone script (run as a real subprocess) and the primitive emit byte-identical events across every payload dialect — pins the inlined copy to the primitive so they cannot drift
- existing `tests/test_hook_script.py` behavior unchanged (incl. `test_installed_copy_matches_source`)

## Verification

- Full `pytest` green in **both TMUX states** (TMUX-unset and inside a real tmux session): 2717 passed, 7 skipped. The one failure (`test_tui_app.py::test_validate_findings_multiline_message_keeps_column_alignment`) is **pre-existing and unrelated** — an environment-sensitive TUI rendering-width test that fails identically on untouched `main`.
- Full `trunk check` — no issues.

## Context

Session 2 of the PR #226 review plan (make the coding-CLI adapter axis extensible out-of-tree). Independent of the other sessions. The out-of-tree Hermes package (a later session) is where `hermes_hooks.relay_event` gets deleted in favor of pointing its hook at this command.

### DIFF
--- a/src/bmad_loop/cli.py
+++ b/src/bmad_loop/cli.py
@@ -2242,6 +2242,25 @@ def cmd_diagnose(args: argparse.Namespace) -> int:
     return 0
 
 
+def cmd_relay(args: argparse.Namespace) -> int:
+    """Write one canonical BMAD event from a coding-CLI hook payload on stdin.
+
+    The general, in-tree counterpart to ``data/bmad_loop_hook.py`` for
+    global/user-scoped-hook adapters: such an adapter points its native hook at
+    ``bmad-loop relay <Event>``, which reads the hook's JSON payload from stdin
+    and defers to :func:`bmad_loop.events.write_relay_event`. The write is
+    env-gated to the active task, so it is a no-op outside a bmad-loop session.
+    """
+    from .events import write_relay_event
+
+    try:
+        payload = json.load(sys.stdin)
+    except (json.JSONDecodeError, ValueError):
+        payload = {}
+    write_relay_event(args.event, payload if isinstance(payload, dict) else {}, os.environ)
+    return 0
+
+
 def cmd_init(args: argparse.Namespace) -> int:
     from .install import install_into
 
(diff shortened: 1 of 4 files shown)


pbean:
Triage 2026-08-08 (self-note): conceptually stale since PR #493 hardened data/bmad_loop_hook.py's event write (O_NOFOLLOW dir_fd, link-like refusal, atomic replace, mid-write redirect recheck). As written this PR adds a second, unhardened writer of the same control plane — its own docstring concedes the split ownership. Rework: extract #493's hardened primitive rather than duplicate the old shape. If #494's decision lands on option 3 (move event writes out of the agent-writable tree), that work belongs here too. Also collides with PR #226's own `bmad-loop relay` — sequence this after #239 and the #226 re-cut.

pbean:
Closing in favor of a fresh re-cut: 646 commits behind and conceptually stale — written pre-#493, it adds a second, unhardened writer of the events control plane. The re-cut executes the recorded decisions on #494 (option 3) and #498 (option 2): extract #493's hardened write shape as the primitive, relocate events/ to a user-scoped state root, with a version-skew guard for already-installed hooks. #498's digest move follows as a small PR into the same root.

PR #263, opened 2026-07-22
refactor(cli): extract resume + sweep composition into runsetup.py (#243)

### LINKED ISSUE #243: Extract run-composition logic from cli.py callbacks
Finding F-2 of the "Brownfield CLI/TUI Refactor Assessment" (local report, gitignored under `_bmad-output/implementation-artifacts/`).

## Evidence

`cli.py` (2,588 lines) embeds workflow composition in argparse callbacks:

- `cmd_run` (`cli.py:757-845`) builds RunState, validates the worktree, and wires the Engine inline.
- `_make_adapters` (`cli.py:125`) and `_platform_preflight` (`cli.py:214`) are shared composition helpers living in the CLI module.
- `cmd_sweep` (`cli.py:1056`) and `cmd_resume` (`cli.py:1229`) repeat the pattern.

Run composition is untestable without argparse, and the TUI can only reuse it by argv round-trip.

## Recommendation — two PRs

**PR 1 (cmd_run):** create `src/bmad_loop/runsetup.py` (single module — do NOT create a services/ package) holding the adapter factory, platform preflight, and RunState/Engine composition as typed functions; `cmd_run` becomes parse → call runsetup → render.

**PR 2 (cmd_resume + cmd_sweep):** extend the same pattern to the remaining logic-heavy callbacks. Scope guard (assessment anti-goals): do NOT touch thin render-only callbacks (status/list/decisions/attach/etc.) — that would be pass-through layering.

## Ordering

Sessions 6-7 of the refactor program. PR 2 requires PR 1 merged (verify on main).

## Traps

- The run subcommand's argv surface must not change — `tui/launch.py` round-trips through CLI argv; grep it for the flags it composes before touching anything.
- `tests/test_cli.py` monkeypatches `cli.*` names (e.g. `cli._make_adapters`) — keep those names importable from `cli.py` as re-exports/delegators so tests stay untouched and green.
- ruff F401 autofix deletes re-exports — pin with `noqa: F401`, verify with `trunk check --no-fix`.
- Behavior-preserving moves only; no CHANGELOG (not user-visible).

pbean:
## Summary

PR 2 of the two-PR plan in #243 (F-2 of the brownfield CLI/TUI refactor). PR 1 (#262) lifted `cmd_run`'s composition into `runsetup.py`; this extends the identical seam to the two remaining logic-heavy callbacks so they become **parse → compose → render**:

- **`cmd_sweep` path** — `_start_sweep`'s inline composition (run dir + state + pid + `sweep.json` + adapters + `SweepEngine`) moves to `runsetup.compose_sweep`. `_start_sweep` now composes and renders only.
- **`cmd_resume` path** — `_resume_paused_run`'s adapter build + engine selection (sweep vs stories vs plain, chosen from persisted run state) moves to `runsetup.compose_resume`.

Per the assessment's anti-goals, the thin render-only callbacks (status/list/decisions/attach/…) are left untouched — layering them would be pass-through.

## Design

- **Same injection seam as PR 1.** The engine classes (`Engine`/`StoriesEngine`/`SweepEngine`) and `make_adapters` are passed *into* the composers from `cli`'s own module namespace, so the test suite's `monkeypatch.setattr(cli, "SweepEngine"/"Engine"/"_make_adapters", …)` still bites. All `cli.*` names the tests target stay in place.
- **Resume keeps its load-bearing bookkeeping CLI-side.** The policy-snapshot re-stamp, `clear_pause`, `clear_graceful_stop`, `write_pid`, and `save_state` stay in `_resume_paused_run` because their ordering is documented and intentional (`#189`, graceful-stop, and the write_pid-before-save observer window). `compose_resume` receives the already-re-stamped, already-persisted `state` and lifts only the composition.
- Both composers reuse the existing `ComposedRun` artifact (docstring generalized to cover all three composers).
- Dropped the now-orphaned `from .runs import RUNS_DIR` from `cli.py` (its only use moved to `runsetup`, which already imports it; not a monkeypatch/re-export target).

## Constraints honored

- **argv surface unchanged** — no argparse definitions or callback signatures touched; the TUI's argv round-trip is unaffected.
- **Monkeypatch targets preserved** via the injection seam and unchanged `cli` re-exports.
- **Behavior-preserving** (with one intentional carve-out, below) — the moves are pure; the only reordering is the `sweep {id} starting` print now landing after engine construction (a no-output `__init__`), exactly as PR 1 did for `run`.
- **F401 trap** — no new caller-less re-exports; verified with `trunk check` (no autofix).
- No CHANGELOG (not user-visible).

## Review follow-up — sweep.json hardening (2nd commit)

Review surfaced a pre-existing, latent robustness gap on the relocated resume path (the flagged `json.loads(sweep.json)` is a verbatim move from `main`, not new here): a torn/corrupt `sweep.json` from a crash mid-write would abort resume — the recovery path itself. Folded a small hardening into this PR (kept as a separate commit):

- `compose_sweep` now writes `sweep.json` **atomically** (tmp + `atomic_replace`), matching how `save_state` persists `state.json` — so the corruption can't happen in the first place.
- `compose_resume` **guards the read** (`except (OSError, json.JSONDecodeError): opts = {}`), falling back to the same launch defaults as the existing missing-file arm — mirroring `tui.data`'s tolerant run-dir reads.
- New regression test: a paused sweep run with a corrupt `sweep.json` resumes with default options instead of crashing.

This is the one deliberate behavior change vs. a pure refactor; it only affects the corrupt-file edge case.

## Testing

- Full test suite green.
- `trunk check` clean.

Closes #243.

## Summary by CodeRabbit

* **Refactor**
  * Improved sweep run startup and paused-run resume workflows.
  * Run configuration and restoration are now handled consistently across command-line and other supported entry points.
  * Resume operations better restore previously saved sweep settings, filters, and execution limits.
  * Existing run tracking, stop handling, and summary reporting remain intact.

### DIFF
--- a/src/bmad_loop/cli.py
+++ b/src/bmad_loop/cli.py
@@ -65,7 +65,6 @@ from .journal import Journal, load_state, save_state
 from .model import RunState
 from .platform_util import MAX_SEGMENT
 from .process_host import ProcessHostError
-from .runs import RUNS_DIR
 
 # The run-composition helpers now live in runsetup.py (the library layer a non-CLI
 # frontend imports). They are re-exported under their historical private names —
@@ -789,46 +788,26 @@ def _start_sweep(
     trigger: str,
     run_id: str | None = None,
 ) -> int:
-    run_id = run_id or runs.new_run_id()
-    run_dir = project / RUNS_DIR / run_id
-    journal = Journal(run_dir)
-    state = RunState(
-        run_id=run_id,
-        project=str(project),
-        started_at=time.strftime("%Y-%m-%dT%H:%M:%S"),
-        policy_snapshot=pol.to_dict(),
-        run_type="sweep",
-    )
-    save_state(run_dir, state)
-    runs.write_pid(run_dir)
-    options = {
-        "prompting": prompting,
-        "decisions_only": decisions_only,
-        "max_bundles": max_bundles,
-        "repeat": repeat,
-        "max_cycles": max_cycles,
-        "trigger": trigger,
-    }
-    (run_dir / "sweep.json").write_text(json.dumps(options, indent=2), encoding="utf-8")
-    adapters = _make_adapters(project, run_dir, pol)
-    journal.append("run-start", run_id=run_id, run_type="sweep", trigger=trigger)
-    print(f"sweep {run_id} starting (attach: bmad-loop attach)")
-    engine = SweepEngine(
+    # The composition (run dir + state + pid + sweep.json + adapters + engine)
+    # lives in runsetup; this stays compose -> render. SweepEngine and
+    # _make_adapters are handed in from this module's namespace so the test suite's
+    # `monkeypatch.setattr(cli, "SweepEngine"/"_make_adapters", ...)` still applies.
+    composed = runsetup.compose_sweep(
+        project=project,
         paths=paths,
         policy=pol,
-        adapter=adapters["dev"],
-        review_adapter=adapters["review"],
-        triage_adapter=adapters["triage"],
-        run_dir=run_dir,
-        journal=journal,
-        state=state,
+        run_id=run_id,
         prompting=prompting,
         decisions_only=decisions_only,
         max_bundles=max_bundles,
         repeat=repeat,
         max_cycles=max_cycles,
+        trigger=trigger,
+        make_adapters=_make_adapters,
+        sweep_engine_cls=SweepEngine,
     )
-    summary = engine.run()
+    print(f"sweep {composed.run_id} starting (attach: bmad-loop attach)")
+    summary = composed.engine.run()
     print(summary.render())
     return 0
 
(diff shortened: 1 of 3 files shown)


pbean:
@coderabbit review

pbean:
Addressed in-PR (commit `8d28322`), kept as a separate commit from the refactor.

The flagged `json.loads(sweep.json)` was a **verbatim move** from `_resume_paused_run` on `main` (`f0ad079`) — pre-existing, not introduced here — so rather than only tolerate the corruption I also fixed its root cause:

- **compose_sweep** now writes `sweep.json` **atomically** (tmp + `atomic_replace`), matching how `save_state` persists `state.json` — a crash mid-write can no longer leave a torn file.
- **compose_resume** guards the read as suggested (`except (OSError, json.JSONDecodeError): opts = {}`), falling back to the same launch defaults as the existing missing-file arm — mirroring `tui/data.py`'s tolerant run-dir reads.
- Added a regression test (`test_resume_tolerates_a_corrupt_sweep_json`): a paused sweep run with a corrupt `sweep.json` resumes with default options instead of aborting.

Full suite green; `trunk check --no-fix` clean.

PR #274, opened 2026-07-23
fix(deferredwork): keep flat appends visible after canonical entries

Haven2026:
## Summary

- stop canonical entry spans before the flat `- source_spec:` format emitted by inner BMAD sessions
- keep those later flat findings visible to `parse_legacy()` even after a canonical append
- make `append_entry()` emit the canonical `location:` field with a safe `n/a` default

## Verification

- red before fix: 2 failures (`flat append` invisible; writer missing `location`)
- `uv run pytest tests/test_deferredwork.py -q` — 33 passed
- `GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=core.fsmonitor GIT_CONFIG_VALUE_0=false uv run pytest tests/test_deferredwork.py tests/test_sweep.py -q` — 104 passed
- `uv run ruff check src/bmad_loop/deferredwork.py tests/test_deferredwork.py` — passed
- real `bmad-loop sweep --dry-run` fixture reports 1 canonical open plus 1 visible legacy entry

The process-level Git config only disables a local macOS fsmonitor socket that otherwise cannot be copied by the test fixture; it does not affect parser behavior.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Deferred Work ledger parsing so embedded “flat appender” blocks no longer get included in canonical `### DW-*` entry bodies.
  - Newly generated Deferred Work entries now include a `location` field (default: “n/a”), and `append_entry` writes it when provided.
  - Added stricter input validation for `append_entry`, `mark_done`, and `append_decision` (rejects invalid status/severity and malformed/invalid dates and multiline inputs).
- **Tests**
  - Added regression coverage for canonical vs flat-appender parsing boundaries, including CRLF handling.
  - Added validation tests ensuring invalid inputs are rejected without modifying the ledger.

### DIFF
--- a/src/bmad_loop/deferredwork.py
+++ b/src/bmad_loop/deferredwork.py
@@ -14,19 +14,21 @@ from __future__ import annotations
 import hashlib
 import re
 from dataclasses import dataclass
+from datetime import date as calendar_date
 from pathlib import Path
 
 HEADING_RE = re.compile(r"^### (DW-\d+): (.+?)\s*$", re.MULTILINE)
 ANY_HEADING_RE = re.compile(r"^#{1,6} ", re.MULTILINE)
 FLAT_ENTRY_RE = re.compile(
-    r"^- source_spec:[^\r\n]*\n"
-    r"[ \t]+summary:[^\r\n]*\n"
-    r"[ \t]+evidence:[^\r\n]*(?:\n|$)",
+    r"^- source_spec:[^\r\n]*(?:\r\n|\n)"
+    r"[ \t]+summary:[^\r\n]*(?:\r\n|\n)"
+    r"[ \t]+evidence:[^\r\n]*(?:\r\n|\n|$)",
     re.IGNORECASE | re.MULTILINE,
 )
 STATUS_RE = re.compile(r"^status:[ \t]*(.*)$", re.MULTILINE)
 CANONICAL_STATUS_RE = re.compile(r"^(?:open|done \d{4}-\d{2}-\d{2})$")
 CANONICAL_SEVERITIES = frozenset({"critical", "high", "medium", "low"})
+LINE_BREAK_RE = re.compile(r"[\n\r\v\f\x1c-\x1e\x85\u2028\u2029]")
 
 
 @dataclass(frozen=True)
@@ -98,6 +100,8 @@ def mark_done(path: Path, dw_id: str, date: str, note: str) -> bool:
     Returns False (no write) when the entry is missing or already done."""
     if not path.is_file():
         return False
+    _require_iso_date("date", date)
+    _require_single_line("note", note)
     text = path.read_text(encoding="utf-8")
     entry = _find_entry(text, dw_id)
     if entry is None or not entry.open:
@@ -118,6 +122,9 @@ def append_decision(path: Path, dw_id: str, date: str, label: str, detail: str) 
     """Record a human decision on an entry without changing its status."""
     if not path.is_file():
         return False
+    _require_iso_date("date", date)
+    _require_single_line("label", label)
+    _require_single_line("detail", detail)
     text = path.read_text(encoding="utf-8")
     entry = _find_entry(text, dw_id)
     if entry is None:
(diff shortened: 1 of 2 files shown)


pbean (changes requested):
Thanks for this — and welcome! You found a real one. I reproduced it on current `main`: a flat
appender block after the last canonical `### DW-<n>` entry is absorbed into that entry's
`parse_ledger()` span, `parse_legacy()` masks it out, and the finding becomes invisible to the
sweep — silently lost work. I've filed #304 to track it, and your `parse_ledger` boundary
truncation is exactly the right seam for the fix (it's the single source of spans that
`parse_legacy`'s masking consumes). The strict three-line `FLAT_ENTRY_RE` matching the shape
`step-04-review.md` instructs is a good call too — a looser boundary could truncate mid-entry and
orphan a `status:` line, which fails in the worse direction.

Two asks before this can land:

**1. Rebase onto current `main`.** `deferredwork.py` was rewritten the day after you opened this
(PR #300): `mark_done` is now a thin wrapper over a new `mark_done_many` (single atomic write via
`atomic_write_text`), and new `parse_declaration`/`classify` helpers sit in the middle of the
file. The PR is currently conflicting; the flat-boundary hunk and the tests apply nearly cleanly,
the validation hunks don't.

**2. Trim this PR to the flat-boundary fix (+ the `location:` writer change).** Please keep:

- `FLAT_ENTRY_RE` + the `parse_ledger` boundary truncation, with your three boundary tests
  (including the CRLF one — nice catch);
- the `append_entry` `location:` field with the `n/a` default and its tests — that closes a real
  gap against the documented canonical shape in `deferred-work-format.md`.

And please drop the input-validation commits (`9b2efe2f`, `582e5662`, `9c6d5c86`) from this PR.
The injection vector they target is real, but raise-style validation in the mutators conflicts
with how the orchestrator has to fail on those paths: `sweep.py` passes LLM-authored triage
`evidence` into `mark_done`'s note and `decisions.py` passes triage `resolution`/`intent` as
decision details, and `validate_triage` only strips — so a multiline value from a session would
now crash a live sweep mid-close, where the ledger-close doctrine is "never a gate, never crash
the run". There's also a placement trap post-rebase: the engine's story-close path calls
`mark_done_many` directly, so guards on `mark_done` alone would be inert exactly where they matter
most. I've split this out as #305 with the design that fits (reject at the `validate_triage`
schema boundary where errors become session feedback, sanitize line breaks at the writer
chokepoints) — it credits your commits as prior art, and it's yours if you'd like to take it.

Optional polish while you're in there: the boundary's first line accepts only `- ` while the
legacy recognizer accepts `[-*][ \t]+` bullets (`_BULLET_RE`/`_FLAT_SOURCE_RE`) — widening it
keeps the two in agreement about what counts as a flat block.

If you add `Fixes #304` to the PR description it'll close the tracker on merge. CI on fork PRs
needs a maintainer approval per run — ping here (or just push) once rebased and I'll approve it.

pbean:
Landing this as #322, with your commit `fbf7541` cherry-picked so it carries your authorship — the
find and the seam are yours, and the CHANGELOG entry credits you. Closing this one in favour of
that; two things I owe you first.

**I was wrong about the strict regex.** My review above called the three-line `FLAT_ENTRY_RE` "a
good call" on the reasoning that a looser boundary could orphan a `status:` line. I measured it
this time and both halves of that were wrong:

- The three-line shape covers **2 of 9** flat shapes `parse_legacy` accepts. It misses the
  no-`summary:` and no-`evidence:` blocks — which are not hypothetical, they are exactly what
  `test_flat_appender_missing_summary_falls_back` and `test_flat_appender_in_done_section_is_done`
  already pin on `main` — plus `*` bullets, tab and padded markers, and reordered lines. Your
  *first* commit's `^- source_spec:[ \t]` was closer to right; the CodeRabbit-driven narrowing in
  commit 3 took coverage away.
- And it **does not prevent the orphaned status** it was narrowed for. A flat-shaped bullet inside
  an entry ahead of its `status:` truncates over that status under the strict regex too, and the
  entry then reads as neither open nor done — `open_ids()` drops it, `classify()` calls it
  malformed. Your `test_source_spec_bullet_without_flat_shape_stays_in_canonical_body` passes on
  unpatched `main`, so it wasn't catching that; its fixture just happens to omit `summary:` and so
  never matches the pattern.

The real fix for the orphan is structural, not a narrower pattern: search for the block starting at
the entry's own `status:` line and never above it. #322 does that, and keeps your loose opening-line
match — shared with `_FLAT_SOURCE_RE` so the two can't drift again. Sorry for sending you the wrong
way on that; it cost you two of the four commits.

**The validation commits (`9b2efe2f`, `582e5662`, `9c6d5c86`) are a real finding, and they're
#305.** Not carried into #322 because raising inside the mutators crashes a live sweep — `sweep.py:911`
feeds LLM triage `evidence` straight into `mark_done`'s note and `_close_resolved` is called bare
from `sweep.py:541`, against a contract that says a ledger close is "never allowed to fail the story
or crash the run". #305 has the layered design (schema-boundary feedback + sanitize at the writer
chokepoints + raise only for orchestrator-owned date/status/severity, at `mark_done_many` rather
than the `mark_done` wrapper) and it's still yours if you want it. Your `U+2028`/`U+2029`/`\x1c-\x1e`
line-break set was the sharpest thing in those commits and it carries over intact.

Thanks for this — a silent work-losing bug in the ledger is about the worst kind we have, and you
found it on your first PR here.

PR #282, opened 2026-07-24
fix(notify): native desktop notifications on macOS/Windows + warn when inert (#231)

### LINKED ISSUE #231: `notify.desktop = true` silently does nothing on macOS and Windows
### Description

`[notify] desktop = true` is accepted on every platform, but the only desktop channel is
`notify-send`, which exists on Linux. On macOS and Windows the `shutil.which` guard makes the
call a no-op, so the setting is enabled, valid, and does nothing — with no indication anywhere
that it cannot work.

`gates.py`:

```python
# portability: notify-send is Linux-only; the shutil.which guard makes this a
# no-op on macOS/Windows (no desktop notification), so no platform branch needed.
if policy.notify.desktop and shutil.which("notify-send"):
```

The comment shows this is a known, deliberate simplification. The report is about its
consequence for unattended runs, which I think is larger than it looks.

### Steps to reproduce

1. On macOS (or Windows), `[notify] desktop = true` in `.bmad-loop/policy.toml` (the default).
2. Run anything that calls `gates.notify` — e.g. trigger a worktree-open failure, or reach an
   epic gate with `gates.mode = "per-epic"`.
3. Watch for a desktop notification.

### Expected behavior

Either a native notification (`osascript -e 'display notification …'` on macOS; a PowerShell
toast or equivalent on Windows), **or** a warning at config-validation time saying `notify.desktop`
has no effect on this platform.

### Actual behavior

Nothing appears. The only surviving signal is the `ATTENTION` file appended inside the run
directory — which nobody is tailing, and which is inside `.bmad-loop/` where a user has no reason
to look while a run is healthy.

### Why this matters more on macOS than the line count suggests

Every "attention" path the orchestrator has ends in `gates.notify`. On the two platforms where it
is a no-op, an unattended run that needs a human has effectively no way to reach one. Concretely,
what an overnight run does when it needs you, per platform:

| | Linux | macOS / Windows |
|---|---|---|
| `ATTENTION` file | ✅ | ✅ (unread) |
| desktop notification | ✅ | ❌ silently absent |

I hit this on a mount-failure deferral: the engine did exactly the right thing — journalled
`worktree-open-failed` and called `gates.notify` — and I learned about it the next morning,
because on macOS that notify call wrote a line to a file and nothing else. The engine's behaviour
was correct; the channel just wasn't there.

I ended up wiring my own notifier (`osascript` + a Telegram push) outside bmad-loop to cover
this, which is what prompted the report.

### Suggested fix

Small version: branch on platform in `gates.notify` — `osascript -e 'display notification "…"
with title "…"'` on macOS, and on Windows either a PowerShell toast or simply nothing but an
honest warning. Both stay best-effort and keep the existing try/except shape.

Even smaller version, if the platform branch isn't wanted: warn once — at `validate` or at run
start — when `notify.desktop` is true and no supported notifier is available, so the setting
stops being silently inert.

Happy to open a PR for either.

### Confirm

pbean:
Closes #231

## Problem

`notify.desktop` defaults to `true`, but the only desktop channel was `notify-send` (Linux-only). On macOS/Windows the `shutil.which("notify-send")` guard made the call a **silent no-op** — the setting was enabled, valid, and did nothing, with no indication anywhere. Since `gates.notify` is the single choke every "a human is needed" path funnels through (escalations, deferrals, worktree-open failures, run-finished), an unattended macOS/Windows run that needed a human had no channel but the untailed `ATTENTION` file in `.bmad-loop/`.

## Fix

A small notifier-resolution seam in `gates.py`, consumed by `notify`, `validate`, and the engine:

- **Native per-platform dispatch** in `gates.notify`: `osascript` (macOS), a best-effort WinRT `ToastNotificationManager` PowerShell toast (Windows, dependency-free), `notify-send` (Linux). Resolution is gated on `sys.platform` first (not `which` alone) so PowerShell Core on Linux can't divert Linux off `notify-send`. Windows stays strictly best-effort inside the existing try/except — a failure only ever no-ops.
- **Untrusted title/message are passed via environment variables** (`system attribute` / `$env:`), never interpolated into the osascript/PowerShell command string, so quotes/newlines/metacharacters in story keys or error tails can't break out. `notify-send` keeps taking them as argv (already injection-safe). A test proves the raw text lands in `env`, not `argv`.
- **Honest warning when inert**, in two places so unattended launches that skip `validate` still surface it:
  - `validate` emits a `notify.desktop-unavailable` **warning** finding when `notify.desktop` is set but no notifier resolves on the platform.
  - Run start prints a one-time stderr `warning:` line and journals a `notify-desktop-unavailable` event (gated by `_owns_signals` so nested auto-sweeps don't repeat it).

No new policy field — reuses the existing `notify.desktop` / `notify.file`.

## Tests

- New `tests/test_gates.py`: file sink; `desktop_notifier_kind` per platform + `None`; osascript/PowerShell/notify-send argv+env dispatch; the untrusted-text-via-env-not-argv assertion; no-op when no notifier; error swallow.
- `tests/test_cli.py`: `validate` warns when no notifier / silent when one is available.
- `tests/test_engine.py`: `_warn_desktop_notifier_inert` journals + prints when inert / no-ops when a notifier is present.

Full suite green (2928 passed, 1 skipped); `trunk check --no-fix` clean. The Windows toast path is unverifiable on this repo's Linux/macOS CI and is best-effort by design.

## Summary by CodeRabbit

* **Bug Fixes**
  * Desktop notifications now use platform-native delivery on macOS, Windows, and Linux (best-effort).
  * Untrusted notification title/message text is handled safely to avoid unsafe command construction.
  * Added validation and one-time startup warnings when desktop notifications are enabled but unavailable, with guidance pointing to the run-directory alert behavior.
* **Documentation**
  * Expanded `[notify] desktop = true` policy documentation to reflect platform-specific mechanisms.
* **Tests**
  * Added/updated tests for notifier dispatch, selection behavior, and the new unavailable-notifier warning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,16 @@ breaking changes may land in a minor release.
 
 ### Fixed
 
+- **`notify.desktop` works on macOS/Windows, and warns when it can't (#231).** The desktop
+  notification channel was `notify-send`-only, so on macOS and Windows `notify.desktop` (default
+  `true`) was silently inert — every "a human is needed" path (escalations, deferrals,
+  worktree-open failures) reached no one, leaving only the untailed `ATTENTION` file. `gates.notify`
+  now dispatches natively per platform: `osascript` (macOS), a best-effort WinRT PowerShell toast
+  (Windows), `notify-send` (Linux); the untrusted title/message are passed via environment
+  variables, never interpolated into the command string. When `notify.desktop` is set but no
+  notifier exists on the platform, `validate` emits a `notify.desktop-unavailable` warning and the
+  run start prints a one-time stderr warning plus a `notify-desktop-unavailable` journal event.
+
 - **Scrollable modal dialogs (#275).** The decision, escalation, confirm, sweep-options and
   story-checkpoint TUI dialogs now scroll their bodies and dock their action buttons, so the
   choose/action buttons stay reachable with long content down to the dialog's minimum frame
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,8 +72,9 @@ breaking changes may land in a minor release.
   `true`) was silently inert — every "a human is needed" path (escalations, deferrals,
   worktree-open failures) reached no one, leaving only the untailed `ATTENTION` file. `gates.notify`
   now dispatches natively per platform: `osascript` (macOS), a best-effort WinRT PowerShell toast
-  (Windows), `notify-send` (Linux); the untrusted title/message are passed via environment
-  variables, never interpolated into the command string. When `notify.desktop` is set but no
+  (Windows), `notify-send` (Linux); the untrusted title/message are handed to `osascript`/PowerShell
+  through environment variables and to `notify-send` as argv — never interpolated into the command
+  string. When `notify.desktop` is set but no
   notifier exists on the platform, `validate` emits a `notify.desktop-unavailable` warning and the
   run start prints a one-time stderr warning plus a `notify-desktop-unavailable` journal event.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,10 @@ breaking changes may land in a minor release.
   string. When `notify.desktop` is set but no
   notifier exists on the platform, `validate` emits a `notify.desktop-unavailable` warning and the
   run start prints a one-time stderr warning plus a `notify-desktop-unavailable` journal event.
+  Note: headless CI cannot observe a real toast/notification (there is no macOS job), so the
+  unit tests assert command construction only. To verify visual delivery, set
+  `notify.desktop = true` and trigger a notify path on each OS (macOS/Windows/Linux) — or call
+  `gates.notify` directly — and confirm a notification appears.
 
 - **Scrollable modal dialogs (#275).** The decision, escalation, confirm, sweep-options and
   story-checkpoint TUI dialogs now scroll their bodies and dock their action buttons, so the
--- a/README.md
+++ b/README.md
@@ -343,7 +343,7 @@ session_budget_grace_s = 240     # enforce mode: seconds to wrap up after the nu
 commands = ["pytest -q", "ruff check ."]
 
 [notify]
-desktop = true             # desktop notification on gate pauses / escalations
+desktop = true             # desktop notification on gate pauses / escalations — notify-send (Linux) / osascript (macOS) / PowerShell toast (Windows), best-effort
 file = true                # append the same alerts to the run's ATTENTION file
 
 [review]
--- a/src/bmad_loop/checks.py
+++ b/src/bmad_loop/checks.py
@@ -64,6 +64,7 @@ VALIDATE_CHECKS: frozenset[str] = frozenset(
         "mux.selection",
         "mux.external-backend",
         "host.process",
+        "notify.desktop-unavailable",
         "skills.base",
         "skills.base-missing",
         "skills.base-incomplete",
--- a/src/bmad_loop/cli.py
+++ b/src/bmad_loop/cli.py
@@ -20,6 +20,7 @@ from . import (
     decisions,
     deferredwork,
     envvars,
+    gates,
     install,
     machine,
 )
@@ -257,6 +258,20 @@ def cmd_validate(args: argparse.Namespace) -> int:
         report.fail("git.probe", f"git check failed: {e}")
 
     report.extend(_platform_preflight())
+
+    # #231: notify.desktop defaults to true but only fires when a platform notifier
+    # exists (osascript/PowerShell/notify-send). When none does, the setting is
+    # silently inert — warn so it stops being a no-op nobody can see.
+    if pol is not None and pol.notify.desktop and gates.desktop_notifier_kind() is None:
+        report.warn(
+            "notify.desktop-unavailable",
+            f"notify.desktop is set but no desktop notifier is available on "
+            f"{sys.platform} — desktop notifications are silently skipped; the "
+            f"ATTENTION file in the run directory is the only alert channel "
+            f"(macOS ships osascript; Linux needs notify-send; Windows needs "
+            f"PowerShell). Install one or set notify.desktop = false.",
+            {"platform": sys.platform},
+        )
 
     for tool in dict.fromkeys(p.binary for p in profiles):
         if shutil.which(tool):
--- a/src/bmad_loop/cli.py
+++ b/src/bmad_loop/cli.py
@@ -263,13 +263,21 @@ def cmd_validate(args: argparse.Namespace) -> int:
     # exists (osascript/PowerShell/notify-send). When none does, the setting is
     # silently inert — warn so it stops being a no-op nobody can see.
     if pol is not None and pol.notify.desktop and gates.desktop_notifier_kind() is None:
+        # The ATTENTION file is only a fallback when notify.file is also on; with it
+        # off there is no alert channel left at all, so say so rather than pointing
+        # at a file that is never written.
+        channel = (
+            "the ATTENTION file in the run directory is the only alert channel left"
+            if pol.notify.file
+            else "notify.file is also off, so no alert channel is configured — "
+            "enable notify.file"
+        )
         report.warn(
             "notify.desktop-unavailable",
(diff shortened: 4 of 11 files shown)


pbean:
Thanks @coderabbitai — validated each finding against the code (and current WinRT docs for the Windows one). Addressed in `00168bd`:

- **Windows toast AUMID (gates.py)** — ✅ Valid. `CreateToastNotifier('bmad-loop')` used an unregistered AUMID, which Windows drops without a Start-Menu shortcut. Rather than add install-time shortcut machinery (out of scope for this best-effort path), the toast now reuses **Windows PowerShell's own default-registered AUMID** (`{1AC14E77-…}\WindowsPowerShell\v1.0\powershell.exe`), so it actually displays (attributed to "Windows PowerShell"), no new dependency. Those `_WIN_TOAST_PS` segments are now raw strings so the `\v1.0` backslash stays literal instead of becoming a vertical tab.
- **`_owns_signals` gate (engine.py)** — ✅ Valid, good catch. Switched the run-start warning gate to `if not self._is_nested:`. `_owns_signals` is also False for a top-level run that couldn't install handlers off the main thread (the code's own comment notes this), which was wrongly silencing the warning; `_is_nested` still suppresses it for nested auto-sweeps.
- **ATTENTION-file guidance (cli.py + engine.py)** — ✅ Valid. The message now branches on `notify.file`: only names the ATTENTION file as the remaining channel when file logging is on; when it's off, it says no alert channel is configured and to enable `notify.file`.
- **CHANGELOG wording** — ✅ Valid. Reworded so the env-var claim is scoped to `osascript`/PowerShell; Linux passes text to `notify-send` as argv.
- **test_gates Windows title in argv** — ✅ Valid. The Windows test now uses a PowerShell/shell-sensitive **title and message** and asserts neither reaches `argv` (plus that a registered AUMID is used).
- **test_engine "invoke twice, assert one"** — ⏭️ Skipped. `_warn_desktop_notifier_inert()` is intentionally not idempotent; the once-per-run guarantee is the single guarded call site in `run()`, not the helper. Invoking it twice would legitimately warn twice, so that test would fail correct code. The existing single-call test already asserts exactly one event + one warning.

Docstring-coverage pre-merge warning is a CodeRabbit-wide metric (it counts test functions); this repo gates on `trunk check` (clean) and the new public functions carry docstrings. Full suite green (2928 passed, 1 skipped).

pbean:

**Recommendation: request changes before merge.** I found four implementation defects and three verification gaps in the final PR diff (`3f59ee4...00168bd`).

### Findings

1. **High — embedded NUL text can crash the run** (`src/bmad_loop/gates.py:95`). `subprocess.run` raises `ValueError` for NULs in argv/environment values, but the best-effort boundary catches only `SubprocessError` and `OSError`. Plugin veto reasons are unrestricted, so this is reachable. Reproduced locally as `ValueError: embedded null byte`.

2. **High — a `pre_run` pause bypasses the unavailable-notifier warning** (`src/bmad_loop/engine.py:335`). `_emit_run_boundary("pre_run")` executes before `_warn_desktop_notifier_inert()`. A supported plugin `pause` veto raises immediately, leaving the run paused without the promised stderr warning or journal event.

3. **Medium — Linux notification text is vulnerable to option injection** (`src/bmad_loop/gates.py:80`). The `notify-send` argv has no `--` terminator. GLib parses options even after the title, so an untrusted message such as `--help` produces help output and no notification. Reproduced locally.

4. **Low — nested auto-sweeps can duplicate warnings off the main thread** (`src/bmad_loop/engine.py:497`). `_is_nested` is derived solely from signal ownership. Off-main-thread engines cannot own signals, so both a parent and its synchronous child are classified non-nested and emit the warning/event.

5. **Medium — native macOS/Windows delivery is never exercised** (`tests/test_gates.py:19`). Both native paths mock `subprocess.run`; Windows CI never executes the WinRT command, and there is no macOS job. The PR's central platform behavior therefore remains mock-only.

6. **Low — the PowerShell-Core-only dispatch path lacks coverage** (`tests/test_gates.py:67`). Detection is tested when only `pwsh` exists, but the dispatch test supplies only `powershell.exe`. A regression between detection and executable selection would pass.

7. **Low — the warning policy matrix is incomplete.** Tests do not verify that `desktop=false` suppresses warnings or that `file=false` emits no-channel guidance. Engine tests call the helper directly rather than exercising the new `Engine.run()` wiring.

Two additional candidates were dismissed: discarded nonzero notifier exits are intentional best-effort behavior, and the alleged PowerShell AUMID mismatch was not sufficiently supported as a defect.

### Remediation plan

1. **Harden notification dispatch**
   - Catch `ValueError` at the best-effort subprocess boundary.
   - Add `--` before Linux title/message arguments.
   - Add regression tests for embedded NULs and option-shaped payloads.

2. **Correct the warning lifecycle**
   - Emit the inert-notifier warning before a pausable `pre_run` boundary.
   - Track nested runs explicitly, preferably with context/thread-local run depth rather than signal ownership.
   - Exercise top-level, off-main-thread, nested, and pre-run-veto paths through `Engine.run()`.

3. **Close platform verification gaps**
   - Add a `pwsh`-only dispatch test.
   - Parameterize the `desktop`/`file` warning matrix.
   - Add platform-conditioned command-acceptance smoke tests and a documented visual smoke test where headless CI cannot observe notifications.

4. **Validate**
   - Re-run focused notification/warning tests, full pytest, and `trunk check --no-fix`.

Current validation is otherwise green: all reported PR checks pass; locally, `2928 passed, 1 skipped`, and `trunk check --no-fix` reports no issues.

pbean:
## Addressed in `b9dac20`

Validated all seven findings from the review above against the code — all confirmed — and fixed them on this branch. Scope calls: **fix finding 4 now**, **document finding 5**.

### Implementation defects

1. **Embedded-NUL crash (`gates.py`)** — ✅ Fixed. Broadened the best-effort `except` to include `ValueError`. Confirmed a NUL in argv **and** in an env value both raise `ValueError`, and `subprocess.SubprocessError` is not a `ValueError` subclass, so it was escaping. `gates.notify` now truly never raises.
2. **`pre_run` pause bypasses the warning (`engine.py`)** — ✅ Fixed. Moved `_warn_desktop_notifier_inert()` above `_emit_run_boundary("pre_run")`, so a plugin `pause` veto there can't skip the once-per-run warning + journal event.
3. **`notify-send` option injection (`gates.py`)** — ✅ Fixed. Added a `--` terminator before the title/message, so an option-shaped payload (`--help`, `-t …`) stays positional instead of being parsed by GLib.
4. **Off-main-thread nested-warning duplication (`engine.py`)** — ✅ Fixed (chose to fix now). Replaced the signal-ownership derivation of `_is_nested` with a module-level `contextvars` run-depth counter (`run()` → `_run_inner()` wrapper). Off-main-thread top-level runs (which own no signals) are now correctly depth-0 and warn; a nested auto-sweep — verified to run synchronously in-thread via `sweep_factory` — is depth>0 regardless of signal ownership. Signal ownership now governs only handler install + session teardown.

### Verification gaps

5. **Native macOS/Windows delivery only mocked** — 📝 Documented. Headless CI can't observe a real toast/notification (no macOS job), so the native paths stay unit-tested at the command-construction level; added a manual smoke-test note to the CHANGELOG and the `gates.notify` docstring.
6. **`pwsh`-only dispatch path** — ✅ Test added.
7. **Warning policy matrix** — ✅ Tests added for `desktop=false` suppression, `file=false` no-channel guidance, and the off-main-thread / nested paths exercised through `Engine.run()`. The three existing nested-reraise tests now simulate nesting via the run-depth counter.

The two candidates you dismissed (discarded nonzero notifier exits; the alleged AUMID mismatch) are correctly non-issues — no change.

### Also folded in (unrelated)

Hardened `tests/test_tui_app.py`'s `render()` helper with `no_color=True`. It builds a plain Rich `Console`, so under a color-forcing environment (`FORCE_COLOR`/`CLICOLOR_FORCE`) it emitted ANSI codes that shifted a column-alignment assertion in one validate-findings test. Now deterministic regardless of ambient color; unrelated to notify but was surfacing as a spurious failure.

### Validation

Full suite `2934 passed, 1 skipped`; `trunk check --no-fix` clean.

PR #283, opened 2026-07-24
fix(validate): accept merged bmad-review, drop unreleased verification-gap requirement (#260)

### LINKED ISSUE #260: validate requires bmad-review-verification-gap, which no BMAD-METHOD release ships; error misdiagnoses cause
## Summary
`bmad-loop validate` hard-fails unless the skill ID `bmad-review-verification-gap` is present in the project's skill tree, but **no BMAD-METHOD release currently ships that skill**, and the failure message points users at the wrong remedy.

## Environment
- bmad-loop 0.9.0
- BMAD-METHOD v6.10.0 (latest release)

## What happens
On a stock latest-release BMAD-METHOD install (`bmm` present), `validate` reports:
```
FAIL: .claude/skills/bmad-review-verification-gap not found — install the BMad Method (bmm) module (the orchestrator drives this upstream skill directly)
```
But `bmm` **is** installed. The real cause: `bmad-review-verification-gap` is not in v6.10.0. It was added post-6.10.0 (BMAD-METHOD PR #2550), then consolidated into a merged `bmad-review` skill (as the `verification-gap` lens) on `main`. On current `main`, `bmad-review-verification-gap` exists only as a thin v6-shim **forwarder** to `bmad-review`. None of this is in a tagged release yet, so a normal install cannot satisfy the check.

## Impact
- The error misdiagnoses the problem (says bmm is missing when it isn't), so users cannot act on it.
- There is no supported path to obtain the required skill from a release. The only workaround is hand-copying the unreleased merged `bmad-review` + forwarders from BMAD-METHOD `main` into the skill tree — which also desyncs BMAD's `_bmad/_config/files-manifest.csv` / `skill-manifest.csv` (stale checksums + manifest-untracked skills).

## Suggested fix
- Make the message actionable: name the exact missing skill ID and the minimum BMAD-METHOD version/layout that provides it.
- Accept **either** a standalone `bmad-review-verification-gap` **or** the merged `bmad-review` + verification-gap forwarder, so both pre- and post-consolidation installs validate.
- Document/pin the minimum BMAD-METHOD (`bmm`) version bmad-loop 0.9.0 supports, and coordinate with BMAD-METHOD to cut a release including the consolidation.
- Optionally, have `bmad-loop init` vendor/verify the required upstream review skills at a known-good ref to avoid manual copying + manifest drift.

pbean:
Closes #260.

## The bug

`bmad-loop validate` — and the shared `run`/`resume`/`sweep` preflight — hard-failed unless `bmad-review-verification-gap` was present in every active CLI skill tree, and told the user to *"install the BMad Method (bmm) module"* on projects where bmm **was** installed. On a stock latest-release install the loop could not start at all.

## Upstream topology (verified against `bmad-code-org/BMAD-METHOD`, 2026-07-23)

- **v6.10.0 (latest release)** ships `bmad-dev-auto` (with its `customize.toml`), `bmad-review-adversarial-general`, and `bmad-review-edge-case-hunter`. It does **not** ship `bmad-review-verification-gap`, and its `step-04-review.md` invokes only those two hunters by name — never verification-gap.
- **`main` (unreleased)** ships the merged `bmad-review` skill, which provides every review lens itself; the three hunter IDs are thin forwarders to it, and step-04 is layer-driven (`{workflow.review_layers}`), naming no hunter.

So the required skill was unsatisfiable in **every** real topology, and its absence causes no runtime failure on the only release that satisfies the loop's other requirements.

## The fix

- `DEV_BASE_SKILLS` (the preflight catalog) drops `bmad-review-verification-gap`. It stays in `BASE_SKILLS` so a hand-installed forwarder is still carried into a worktree copy-if-present — it is simply never validated.
- New `MERGED_REVIEW_SKILL` / `_REVIEW_LAYER_SKILLS`: a tree carrying `bmad-review/SKILL.md` satisfies the review layers on its own, so the standalone hunters are skipped there. Computed **per tree**, since a project can have a post-consolidation `.claude` tree and a pre-merge `.agents` one side by side.
- `bmad-dev-auto` stays hard-required — the merged reviewer never substitutes for the dev primitive.
- Both `skills.base-missing` messages are rewritten to be actionable and to stop claiming bmm is uninstalled: they name the remedy (install **or update** bmm, BMAD-METHOD >= 6.10.0) and mention the consolidated `bmad-review` alternative.

Deliberately unchanged: check ids (`skills.base-missing` / `skills.base-incomplete`), `"problem"` severity, `_require_base_skills`' abort-on-any-problem, `checks.py`, and `worktree_flow.py`. No init-time vendoring of upstream skills.

## Behavior, on real skill trees

| Topology | Before | After |
| --- | --- | --- |
| v6.10.0 shape (dev-auto + 2 hunters, no verification-gap, no bmad-review) | 1 problem — unsatisfiable | **0 problems** |
| merged (dev-auto + `bmad-review`, no standalone hunters) | 3 problems | **0 problems** |
| broken install (missing `bmad-review-edge-case-hunter`, no `bmad-review`) | reported, misdiagnosed | 1 problem, actionable message |
| nothing installed | 4 problems | 3 problems |

## Docs

`docs/setup-guide.md` now states the **minimum BMAD-METHOD: v6.10.0** for sprint mode and explains the merged-`bmad-review` path; `README.md` and `docs/FEATURES.md` no longer claim `validate` preflights "all four upstream skills" or list verification-gap as a prerequisite. CHANGELOG entry under `[Unreleased] → Fixed`.

## Tests

`tests/test_install.py`: existing absent/detail tests updated for the 3-skill catalog and the new wording, and the verification-gap sub-case inverted (removing it from a complete tree must now pass). Three new tests: merged `bmad-review` satisfies the review layers (and still doesn't substitute for `bmad-dev-auto`), the v6.10.0 shape validates with neither escape hatch present, and a genuinely missing hunter is still reported with a message that does not falsely claim bmm is uninstalled.

Full suite: 2931 passed, 7 skipped. `trunk check --no-fix`: clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation no longer requires the separate verification-gap review skill and properly derives required review layers from the installed configuration.
  * Added clearer results for missing, unresolved, unreadable, or empty review-layer setups (blocking problems vs non-blocking warnings).
  * Worktree provisioning now includes the resolved review layers and respects project customization seeding behavior.
* **Documentation**
  * Updated requirements and setup guides to reflect BMAD-METHOD ≥ 6.10.0 and the merged vs standalone review-layer topologies.
  * Refreshed terminology and preflight behavior descriptions around “review layers”.
* **Tests**
  * Expanded coverage for merged-review behavior, layer resolution/overrides, and warning-vs-fail CLI outcomes.


pbean:
Correction to the topology summary above: **post-consolidation `bmad-dev-auto` defines four review layers, not three.** Pushed `9a7fd1c` to state this precisely in the docs and the `install.py` comment.

From `customize.toml` `[[workflow.review_layers]]` on BMAD-METHOD main:

| layer | invokes |
| --- | --- |
| `blind-hunter` | `bmad-review`, `adversarial` lens |
| `edge-case-hunter` | `bmad-review`, `edge-case-hunter` lens |
| `verification-gap` | `bmad-review`, `verification-gap` lens |
| `intent-alignment` | nothing — a self-contained inline prompt, no skill |

This does not change the fix: three of the four layers resolve to the single merged `bmad-review` skill and the fourth needs no skill, so accepting `bmad-review` still covers the whole review step. On v6.10.0 there is no `review_layers` section at all — `step-04-review.md` hardcodes exactly two skill invocations (`bmad-review-adversarial-general`, `bmad-review-edge-case-hunter`), which is what the preflight requires there.

Re-verified the core premise across tags: no tag from v6.3.0 through v6.10.0 ships a standalone `bmad-review-verification-gap/SKILL.md` (v6.9.0 and v6.10.0 both ship only the other two). On main it exists solely as a `src/core-skills/v6-shims/` forwarder to `bmad-review`.

**Known gap, pre-existing and out of scope here:** a project whose `customize.toml` is post-consolidation (layers referencing `bmad-review`) but whose installed skills are the pre-consolidation standalone hunters passes this preflight while three of its four layers would fail at run time, because `validate` checks skill directories and not whether the configured layers resolve. The old verification-gap requirement did not catch this either. Worth a follow-up: preflight the skills named by `review_layers` instead of a fixed catalog.

pbean:
Pushed `d1f39f2`, which fixes the layer-resolution gap in this PR rather than deferring it.

**The preflight no longer holds projects to a fixed catalog of review skills.** Which reviewers the review step invokes is a property of the installed `bmad-dev-auto`, not of this package, so it is now read from the skill itself, per tree:

1. `customize.toml` `[[workflow.review_layers]]` when present — each layer's `instruction` names the skill it hands off to. A layer disabled with an empty `instruction` requires nothing, and `_bmad/custom/bmad-dev-auto.toml` / `.user.toml` overrides are merged first (arrays of tables merge by `id`, per customize.toml's own header).
2. otherwise the reviewers `step-04-review.md` names inline — the v6.10.0 shape.
3. if neither can be read (malformed TOML, unknown layout), fall back to the previous static requirement, which both mainstream topologies satisfy.

Missing derived skills are reported as a new `skills.review-layer-missing` problem (registered in `checks.VALIDATE_CHECKS`) carrying `{tree, skill, layers, source}`, so `validate --json` consumers can see exactly which layer is unresolvable.

**This also corrects something the first commit got wrong.** Dropping `bmad-review-verification-gap` from the static catalog silently removed protection for projects whose layers genuinely do invoke it. Derivation restores it: verified against a real install whose four layers name the three standalone hunters — removing `bmad-review-verification-gap` there is now reported again, while a stock v6.10.0 install (which never names it) stays green. Requiring it of everyone was the bug; requiring it of projects that configure it is correct.

Verified against the real upstream files, not fixtures — `git show`n straight out of `bmad-code-org/BMAD-METHOD`:

| topology (real files) | result |
| --- | --- |
| v6.10.0 skill + the two hunters it names | 0 problems |
| v6.10.0 skill + only the merged `bmad-review` | 2 problems (its step-04 names the hunters — the static bypass would have wrongly passed this) |
| main skill + `bmad-review` | 0 problems |
| main skill + the three standalone hunters | 1 problem: `bmad-review` unresolvable, layers `blind-hunter, edge-case-hunter, verification-gap` |

Seven new tests cover the layer-driven, pre-consolidation, disabled-layer, project-override, and unreadable-config paths. Full suite 2938 passed / 7 skipped; `trunk check --no-fix` clean.

pbean:

The central fix is directionally correct for both official BMAD topologies: the v6.10.0 two-hunter layout and current `main`'s merged `bmad-review` layout. However, I recommend addressing the findings below before merge because validation can still disagree with what an isolated run actually executes.

### High severity

1. **Dynamically validated custom reviewers are not copied into isolated worktrees** (`src/bmad_loop/install.py:107-118`, `src/bmad_loop/worktree_flow.py:199-211`).

   Validation can accept a configured skill such as `bmad-review-company`, but `provision_worktree` copies only the fixed `BASE_SKILLS` catalog. If that custom skill is gitignored/untracked, preflight passes in the main checkout and the isolated review fails because the skill is absent.

2. **Personal review overrides are validated but not used in isolated worktrees** (`src/bmad_loop/install.py:129-190`, `src/bmad_loop/worktree_flow.py:127-170`).

   `_bmad/custom/bmad-dev-auto.user.toml` participates in preflight resolution, but it is intentionally gitignored and is not seeded into the worktree. Validation can therefore approve a disabled or replaced layer while the isolated runtime executes the default layer and requires a different skill set.

3. **Override merging does not match BMAD's resolver** (`src/bmad_loop/install.py:173-189`).

   BMAD uses keyed replacement only when every item in the combined array shares `id` or `code`; otherwise it appends the whole override array. The implementation here replaces matching IDs individually even when the override also contains an item without an ID. In that supported mixed-array case, validation can omit a default reviewer that the real runtime resolver retains.

4. **Disabling every review layer passes preflight but guarantees a blocked run** (`src/bmad_loop/install.py:208-224`, `348-394`).

   With every resolved layer's `instruction` empty, `_configured_review_skills` returns an empty requirement set and validation succeeds. `bmad-dev-auto` subsequently halts with `no active review layers`.

### Medium severity

5. **Valid TOML with the wrong structural shape crashes preflight** (`src/bmad_loop/install.py:157-159`).

   For example, syntactically valid `workflow = "wrong shape"` causes `_layers_of` to call `.get` on a string and raise `AttributeError`. This affects `validate`, `run`, `resume`, and `sweep` instead of falling back or returning an actionable finding.

6. **Skill detection silently depends on one exact prose phrase** (`src/bmad_loop/install.py:121-128`, `216-224`).

   Only `Invoke the \`bmad-...\` skill` is recognized. A valid custom instruction such as `Use the \`bmad-review-company\` skill` executes as an agent instruction but creates no preflight dependency. A structured dependency field would be preferable; short-term, an unrecognized backticked `bmad-*` reference should produce an unresolved-layer finding rather than silently pass.

7. **The new topology/override behavior lacks an end-to-end verification matrix** (`tests/test_install.py:845-977`).

   Missing cases include:

   - a newly appended override layer;
   - standalone `bmad-review-verification-gap` required when explicitly named;
   - pre- and post-consolidation skill trees in the same call;
   - malformed override files, not only malformed defaults;
   - team-versus-user override precedence;
   - dynamic-reviewer and personal-override parity under worktree isolation;
   - every layer disabled.

## Recommended remediation plan

1. Introduce one resolver result per skill tree containing active layer IDs, required skill IDs, source, and resolution errors. Match BMAD's structural merge rules exactly.
2. Use that same resolved result for both preflight and worktree provisioning.
3. Seed personal customization files—or a materialized effective configuration—into isolated worktrees.
4. Copy every dynamically required skill into its corresponding worktree, rather than only the fixed `BASE_SKILLS` set.
5. Emit explicit validation findings for malformed configuration, zero active layers, and unresolvable skill references.
6. Add the topology/override/worktree test matrix above.
7. Re-run the full suite, `trunk check --no-fix`, and Pyright.

## Verification performed

Reviewed head `d1f39f2c0f3d82d36e1a797333d4721841bf00ec`.

- Relevant test modules: **328 passed**.
- Full suite: **2,938 passed, 7 skipped**.
- `trunk check --no-fix`: clean.
- Current GitHub checks are green across Python 3.11-3.14, Windows, lint, and Pyright.

The green suite confirms the two official happy-path topologies, but it does not exercise the runtime-parity and customization cases above.

pbean:
## Review remediation — verdicts and fixes (`a2af045`)

Every finding was validated against the code and against the real upstream sources (`bmad-code-org/BMAD-METHOD` `main` @ `bb45db4a`, plus tag `v6.10.0`) rather than against fixtures. **All 7 confirmed**, two with corrections; **2 more surfaced** that the review did not raise. All fixed in this PR.

### Verdicts

| # | Finding | Verdict | Fix |
| --- | --- | --- | --- |
| 1 | Dynamic reviewers not copied into worktrees | Confirmed — gap introduced by `d1f39f2` | `provision_worktree` copies the union of `BASE_SKILLS` and the skills the tree's own layers name |
| 2 | Personal overrides validated but not seeded | Confirmed, **broader than reported** | `_bmad/custom/` seeded into each worktree |
| 3 | Override merge diverges from BMAD's resolver | Confirmed, **understated — 3 divergences** | Ported the real algorithm |
| 4 | All layers disabled passes preflight | Confirmed verbatim upstream | New `skills.review-layers-empty` problem |
| 5 | Wrong-shape TOML crashes preflight | Confirmed, reproduced | `_layers_of` guards each hop |
| 6 | Detection depends on one prose phrase | Confirmed — **remedy changed** | Warning, not problem |
| 7 | No topology/override/worktree test matrix | Confirmed | 21 new tests |

### Corrections

**F3 was understated.** The resolver is `src/scripts/resolve_customization.py`. `_detect_keyed_merge_field` runs on the **combined** array, requires **every** item to carry the same key, and checks **`code` before `id`**; anything else plain-concatenates. Keying on `id` alone diverges three ways, not one:

1. mixed/key-less arrays — upstream appends the whole override, we replaced by id → **dropped a base layer the run keeps** (the reported case, a false PASS);
2. all-`code` arrays — upstream replaces by `code`, we appended both → **required a skill the run never invokes** (a false FAIL, the opposite direction);
3. items carrying both keys — upstream keys on `code`, we keyed on `id`.

The merge is now a line-for-line port, cited by path in the docstring.

**F2 is broader, and its stated mechanism is only half of it.** The upstream installer writes `_bmad/custom/.gitignore` = `*.user.toml` (`tools/installer/core/installer.js:687`), so on a stock install only the *personal* override is untracked — the team `bmad-dev-auto.toml` is tracked and present in a worktree. But projects routinely gitignore `_bmad/` wholesale (**this repo does** — `.gitignore:12`), in which case a worktree has no `_bmad` at all and resolves base layers only, ignoring **both** override layers `validate` just read. Seeding the directory covers both.

**F6's proposed remedy would have re-created the bug this PR fixes.** Turning an unrecognized backticked `bmad-*` reference into a *problem* means a prose mention, or an override handing off to something that is not a skill, blocks every run — #260's exact failure mode. It is now a **warning**. (The concern itself is real: upstream already writes ``use the `bmad-code-review` skill`` elsewhere, which the narrow regex misses.)

**Malformed *override* handling also diverged**, which the findings implied but did not state. `_read_toml` returning `None` made derivation bail to the static catalog; upstream warns and treats the file as empty, still resolving every other layer. We now match upstream and report `skills.customize-unreadable`.

### Two findings not in the review

**A1 — `when`-gated layers were hard-required.** `step-04-review.md` skips *"every layer whose `when` condition (if present) does not hold"*. The preflight ignored `when` and hard-required the layer's skill, so an override that gates a layer produced a false FAIL. The condition is model-evaluated against the diff and undecidable statically → advisory.

**A2 — severity was not honoured, which blocked the whole remedy.** `cli.py:504` aborted the run when `missing_base_skills` returned **anything at all**. Any new warning would have hardened into a hard run abort, so this had to land first. `_require_base_skills` now partitions by severity (warnings print, only problems abort), and `cmd_validate` gates its ok line on problems rather than emptiness.

### On the recommended remediation plan

Steps 1–6 are implemented. One deliberate deviation: step 5 asked for explicit findings on unresolvable skill references, which is step 6's remedy — those emit as warnings, per above.

Deliberately **out of scope**: seeding all of `_bmad/` (the resolver script, `bmm/config.yaml`). That is the separate concern of the still-open #155. `_bmad/custom/` alone is sufficient for layer-resolution parity, because the skill's documented fallback when `resolve_customization.py` is unavailable reads those same three TOMLs by hand (`bmad-dev-auto/SKILL.md:81-89`).

### Verification

- Full suite **2959 passed, 7 skipped** (was 2938 — +21 tests).
- `trunk check --no-fix` clean; `uvx pyright@1.1.411` — 0 errors, 0 warnings.
- New tests were checked to actually discriminate the old behavior, not merely pass. One draft case (mixed `code`/`id` arrays) resolved identically under both implementations and was replaced with the case that does discriminate: an override carrying one id-less item, where the old id-keyed merge silently dropped a base layer the run still executes.

Issue #259, opened 2026-07-22
bmad-loop-setup docs invoke PEP 723 scripts with python3; merge-config.py fails without uv run


jmsundar:
## Summary
`bmad-loop-setup` documents its helper scripts as `python3 ./scripts/<name>.py …`, but `merge-config.py` is a PEP 723 script with an inline `pyyaml` dependency. Under bare `python3` it exits with an error; it only works via `uv run`.

## Environment
- bmad-loop 0.9.0, macOS, uv-managed Python

## Repro
```
$ python3 ./scripts/merge-config.py --help
Error: pyyaml is required (PEP 723 dependency)

$ uv run ./scripts/merge-config.py --help
usage: merge-config.py ...   # works — uv provisions the inline deps
```

`merge-config.py` header:
```
# /// script
# requires-python = ">=3.9"
# dependencies = ["pyyaml"]
# ///
```

`merge-help-csv.py` has no third-party deps, so it runs fine under `python3` — which masks the inconsistency across the two documented commands.

## Suggested fix
Document/use `uv run` for the PEP 723 scripts (BMAD's own `resolve_config.py` header already notes BMAD is "standardizing on `uv run`"), or drop the `pyyaml` dependency in favor of the stdlib.

Issue #64, opened 2026-07-05
bmad-loop-setup deletes core BMAD config/manifest (_bmad/core/config.yaml, _bmad/_config/) in multi-module projects


mswanson:
## Summary

In a multi-module BMAD-METHOD project, running `/bmad-loop-setup` **deletes core BMAD's live, manifest-tracked configuration** — `_bmad/core/config.yaml` and the entire `_bmad/_config/` manifest directory (`manifest.yaml`, `files-manifest.csv`, `agent-manifest.csv`, `skill-manifest.csv`, `bmad-help.csv`). This directly contradicts the README's claim that *"Your standard BMAD install is never modified,"* and it leaves `npx bmad-method install` unable to compute safe upgrades (the file-hash manifest it depends on is gone).

## Environment

- **bmad-loop:** v0.8.0 (installed via `uv tool install`)
- **BMAD-METHOD core:** 6.10.0, with modules `core`, `bmm`, `tea`, `cis` all installed (i.e. a normal multi-module project, not loop-only)
- **IDE:** claude-code

## Steps to reproduce

1. Have a standard multi-module BMAD 6.10 project (`_bmad/_config/manifest.yaml` lists core/bmm/tea/cis; each module has its own `_bmad/<module>/config.yaml`, and `_bmad/core/config.yaml` is tracked with a content hash in `_bmad/_config/files-manifest.csv`).
2. Install the bmad-loop module and run `/bmad-loop-setup`.
3. The skill runs `merge-config.py` (consolidates per-module config into a central `_bmad/config.yaml`) and `cleanup-legacy.py --module-code bmad-loop --also-remove _config --skills-dir .claude/skills`.

## Expected

bmad-loop's setup touches only bmad-loop's own artifacts. Core BMAD's `_bmad/core/config.yaml` and `_bmad/_config/` manifest remain intact.

## Actual

`cleanup-legacy.py` reports `directories_removed: [bmad-loop, core, _config]`, and `merge-config.py` deletes `_bmad/core/config.yaml`. Core BMAD's manifest is destroyed.

## Root cause (confirmed in source)

Both scripts in `src/bmad_loop/data/skills/bmad-loop-setup/scripts/` hardcode `core` as a disposable "legacy installer" directory.

**`merge-config.py` → `cleanup_legacy_configs()`:**
```python
def cleanup_legacy_configs(legacy_dir, module_code, verbose=False):
    """Delete legacy config.yaml files for this module and core only."""
[...]

PR #285, opened 2026-07-24
fix(setup): register into the BMAD v6.10 TOML config layout (#258)

### LINKED ISSUE #258: bmad-loop-setup writes legacy config.yaml/module-help.csv; ignored on BMAD v6.10.0 TOML config layout
## Summary
The `bmad-loop-setup` skill's config/help registration targets the **legacy per-file YAML layout**, but current BMAD-METHOD (v6.10.0) uses a **consolidated TOML layout** that its own resolver reads exclusively. The scripts write files the runtime never reads.

## Environment
- bmad-loop 0.9.0
- BMAD-METHOD v6.10.0 (latest release), `bmm` + `core` modules
- macOS, Python via uv

## Details
`bmad-loop-setup` documents and runs:
- `merge-config.py` → writes `_bmad/config.yaml` and `_bmad/config.user.yaml`
- `merge-help-csv.py` → writes a root `_bmad/module-help.csv`

But on v6.10.0 the consolidated config is TOML, resolved by `_bmad/scripts/resolve_config.py`, which uses `tomllib` and reads **only** these four layers:
`_bmad/config.toml`, `_bmad/config.user.toml`, `_bmad/custom/config.toml`, `_bmad/custom/config.user.toml`. It never reads `config.yaml`. Help is consolidated in `_bmad/_config/bmad-help.csv` (plus per-module `_bmad/<module>/module-help.csv`), not a root `_bmad/module-help.csv`.

Net effect: the merge scripts produce dead / split-brain files. On a stock install the BMAD installer has already registered the module (manifest + per-module config + help CSV), and `bmad-loop validate` reports `BMAD config OK` sourced from the TOML layout — so the merge scripts add nothing but confusion. The `bmad-loop` module's `module.yaml` also has no prompt variables, so there is nothing to collect.

## Suggested fix
- Detect the consolidated TOML layout; when present, no-op config/help registration (the installer already did it) or write TOML.
- Update the skill docs (they reference `config.yaml` throughout) to the TOML layout.
- Consider skipping the merge entirely for modules with no prompt variables.

pbean:
## Problem

BMAD-METHOD **v6.10 consolidated onto TOML**. Its resolver (`_bmad/scripts/resolve_config.py`)
reads exactly four files — `config.toml`, `config.user.toml`, `custom/config.toml`,
`custom/config.user.toml` — and **never** `config.yaml`. `/bmad-help` reads exactly one
catalog, `_bmad/_config/bmad-help.csv`, the installer's assembled concatenation of the
per-module `_bmad/<mod>/module-help.csv` files.

`/bmad-loop-setup`'s two merge scripts still wrote the **pre-6.10** root `_bmad/config.yaml`
and `_bmad/module-help.csv`. Neither is read by anything on v6.10, and neither is even tracked
in `_bmad/_config/files-manifest.csv`. Net effect: on a v6.10 project the module was
**registered nowhere** and its skills never appeared in `/bmad-help`.

Two claims in the issue were checked and **corrected** (details in the
[validation comment](https://github.com/bmad-code-org/bmad-loop/issues/258#issuecomment-5066651324)):

- *"The installer already registered the module, so setup can no-op."* Not on a
  `uv tool install` + `bmad-loop init` project — `manifest.yaml` lists only `core` + `bmm`,
  there are 0 `bmad-loop` rows anywhere, and `init` never touches `_bmad/`. A no-op would
  leave the module **permanently invisible**, so the fix has to *register into the new layout*.
- *"`validate` reports BMAD config OK from the TOML layout."* No — `cli.py` → `bmadconfig.py`
  reads `_bmad/bmm/config.yaml`, the **per-module** YAML, which v6.10.1 still generates and
  manifest-tracks. The runtime is unaffected by this issue (that migration is #154).

## Design — detect the layout, register into what the project actually reads

Layout is **TOML iff `<bmad-dir>/config.toml` exists** (pure `Path.exists()`, no parsing, no
imports). Both scripts gained a required `--bmad-dir` and carry their **own** copy of
`detect_layout` / `detect_orphans` — they ship standalone into user projects, so a shared
import module would break them.

**TOML branch (v6.10+)** writes three things and nothing else:

| File | How |
| --- | --- |
| `_bmad/custom/config.toml` | `[modules.bmad-loop]` — surgical text edit of that table's span only |
| `_bmad/bmad-loop/module-help.csv` | full overwrite (the whole file is ours) |
| `_bmad/_config/bmad-help.csv` | anti-zombie merge, catalog's own header preserved |

**Legacy YAML branch (pre-6.10)** is byte-for-byte today's behavior, still driven by
`--config-path` / `--user-config-path` / `--target`. Each branch fails loudly with the
layout-specific argument it is missing.

### Why no core values on the custom layer

`custom/config.toml` is the layer BMAD documents as *"never touched by the installer"* — which
is exactly why it is the right home for our module table, and exactly why **`user_name`,
`output_folder` and friends must not go there**. The custom layer *wins* over the installer
layer, so a value pinned there would silently override **every future installer answer**. Core
config stays installer-owned; `core_updated` is always `false` and `user_keys` always empty on
this branch, by design. Since `assets/module.yaml` declares zero prompt variables, there is
nothing to collect on this layout at all.

### Writing TOML without a writer

There is no stdlib TOML *writer*, and `custom/config.toml` is a human-authored, comment-bearing
file — round-tripping it through a parser + emitter would destroy it. So the table is
hand-emitted and spliced in via `upsert_module_table()`; everything outside the replaced span
comes back byte-identical, and the result is validated by re-parsing with `tomllib` (lazy
import; `toml_validated: false` + a stderr note when unavailable). **No new dependency was
added** — `tomli-w` was deliberately avoided because it would deepen #259.

### Orphans: report, never delete

Inert pre-6.10 leftovers directly under `_bmad/` (`config.yaml`, `config.user.yaml`,
`module-help.csv`) are reported as `orphans_detected` — `[{"path", "has_module_entries"}]` —
and **never deleted**, per the #64 never-delete-live-config doctrine. `legacy_configs_deleted`
/ `legacy_csvs_deleted` stay `[]`. The skill now tells the user these are safe to remove by
hand and that bmad-loop will not do it for them.

### Caveat: the installer regenerates the catalog

`_config/bmad-help.csv` is generated, not hash-tracked in `files-manifest.csv` — merging into
it desyncs no manifest, but a **BMAD installer re-run regenerates it and drops our rows**.
Expected, and now documented in the skill's Confirm step: re-run `/bmad-loop-setup` to restore
them.

## Commits

- `e03d579` — `merge-config.py`: `--bmad-dir`, `detect_layout`, `emit_module_table`,
  `upsert_module_table`, `detect_orphans`, TOML branch (+8 tests)
- `606f76a` — `merge-help-csv.py`: own layout/orphan copies, per-module CSV + catalog merge,
  stale `HEADER` constant fix (`after`/`before` → `preceded-by`/`followed-by`) (+8 tests)
- `fff4c6b` — docs: `SKILL.md` two-layout rewrite, `docs/setup-guide.md` (what setup does +
  uninstall), `README.md`, `data/skills/README.md`, CHANGELOG

`SKILL.md`'s two script command lines were stale after the first two commits (neither passed
the now-required `--bmad-dir`); the docs commit closes that seam. The commands now carry the
legacy-only arguments too, so **one command line works on both layouts** — they are simply
ignored on the TOML branch.

One drive-by content fix: `assets/module-help.csv`'s `outputs` cell for `bmad-loop-setup` read
`config.yaml entries`, which is wrong on v6.10. It is now the layout-neutral
`module config + help entries`. No test asserted the old string.

## End-to-end proof

A throwaway v6.10 fixture (`config.toml` + `config.user.toml` + a comment-bearing
`custom/config.toml` carrying an unrelated `[agents.bmad-agent-pm]` table + a realistic
`_config/bmad-help.csv`), both scripts run against it **twice**, then BMAD's own **unmodified**
`resolve_config.py` used as the oracle.

**(a) BMAD's resolver reads our registration back:**

```console
$ uv run --no-project --python 3.11 <fixture>/_bmad/scripts/resolve_config.py \
    --project-root <fixture> --key modules.bmad-loop
{
  "modules.bmad-loop": {
    "name": "BMAD Loop Skills",
    "description": "Automation-mode skills driven by the bmad-loop orchestrator: interactive escalation resolution (bmad-loop-resolve) and deferred-work sweep triage (bmad-loop-sweep) — the inner dev primitive (which self-reviews and commits) is the upstream bmad-dev-auto skill",
    "version": "0.9.0",
    "default_selected": false
  }
}
```

The pre-existing comments and the unrelated `[agents.bmad-agent-pm]` table survived verbatim:

```toml
# Team / enterprise overrides for _bmad/config.toml.
# Committed to the repo — applies to every developer on the project.

# my own overrides — keep me
[agents.bmad-agent-pm]
description = "Prefers short, bulleted PRDs."

[modules.bmad-loop]
name = "BMAD Loop Skills"
description = "..."
version = "0.9.0"
default_selected = false
```

**(b) Exactly one block of our rows in the catalog after two runs** (2 pre-existing rows kept):

```console
module,skill,display-name,menu-code,description,action,args,phase,preceded-by,followed-by,...
BMad Method,_meta,,,,,,,,,false,https://docs.bmad-method.org/llms.txt,
Core,bmad-help,Help,H,Show the BMad help menu.,menu,,anytime,,,false,,
BMAD Loop Skills,bmad-loop-setup,Setup Loop Module,SA,Install or update BMAD Loop module c...
BMAD Loop Skills,bmad-loop-sweep,Sweep Triage,ST,Read-only triage of the deferred-work led...

--- 'BMAD Loop Skills' row count: 2 (source has 2 rows)
```

Run 1 reported `table_replaced: false` / `rows_removed: 0, rows_added: 2, total_rows: 4`;
run 2 reported `table_replaced: true` / `rows_removed: 2, rows_added: 2, total_rows: 4`.
`toml_validated: true` on both.

**(c) No legacy files created:**

```console
  ok: _bmad/config.yaml absent
  ok: _bmad/config.user.yaml absent
  ok: _bmad/module-help.csv absent
```

**(d) Orphans reported, never deleted** — a second fixture pre-seeded with pre-6.10 leftovers:

```json
{
  "layout": "toml",
  "orphans_detected": [
    { "path": ".../_bmad/config.yaml",      "has_module_entries": true  },
    { "path": ".../_bmad/config.user.yaml", "has_module_entries": false },
    { "path": ".../_bmad/module-help.csv",  "has_module_entries": true  }
  ],
  "legacy_configs_deleted": [],
  "core_updated": false
}
```

All three files survived byte-identical (`config.yaml` still `bauto:|  version: 0.4.0|bmad-loop:|  version: 0.8.0|`).

## Verification

```
uv run pytest -q     2944 passed, 7 skipped
trunk check          No issues
```

16 new tests in `tests/test_setup_cleanup_scripts.py` behind a `_v610_bmad` fixture cover both
scripts' TOML branches: custom-table write, comment/unrelated-table preservation (append **and**
replace-with-a-neighbour-below), no YAML written, no core values leaked, upsert idempotence
including a version bump, per-module CSV, catalog merge + idempotence, catalog created when
absent, root CSV left alone, orphans reported-not-deleted, and both legacy YAML branches
unchanged plus their missing-argument errors.

Note: `test_merge_help_toml_strips_pre_rename_rows` asserts the *opposite* of what its name
suggests, on purpose — the anti-zombie filter keys on **source** module codes
(`BMAD Loop Skills`), so leftover pre-rename `BMAD Automator Skills` rows **survive** the merge.
That is now documented as a manual-delete step in the skill's bauto-migration section, extended
to cover `_bmad/_config/bmad-help.csv`.

## Out of scope (both remain open)

- **#259** — the scripts' `python3` invocation prefix vs `uv run` for PEP 723 headers. The
  prefix is left exactly as-is here; bundling it was explicitly declined.
- **#154** — the runtime's `_bmad/bmm/config.yaml` read in `bmadconfig.py`. That per-module
  YAML is still live and installer-written at v6.10.1, so the orchestrator runtime is
  unaffected by this PR. Untouched.

Closes #258

## Summary by CodeRabbit

* **New Features**
  * Added support for BMAD v6.10+ TOML-based configuration during setup.
  * Module registration now updates custom configuration and shared help catalogs safely.
  * Added layout-aware migration, cleanup guidance, orphan reporting, and refusal safeguards.

* **Bug Fixes**
  * Improved duplicate prevention, schema handling, symlink support, and protection against data loss during concurrent or failed writes.

* **Documentation**
  * Updated the README, setup guide, skill instructions, component descriptions, and changelog for TOML and legacy layouts.

* **Tests**
  * Expanded regression coverage for TOML workflows, migration, validation, failure handling, and safe file updates.
  * Added a CI gate validating resolver behavior.


pbean:

**Recommendation: request changes; do not merge as-is.** The v6.10 registration design is sound, but the TOML editor can corrupt a valid human-authored config and still report success.

Triage: **2 high**, **3 medium**. The focused and full test suites are green, so these are untested behavioral gaps rather than ordinary test failures.

### Findings

1. **High — valid TOML can be corrupted by the regex upsert** (`merge-config.py:450-500`).

   Reproduced with a valid inline-commented header (`[modules.bmad-loop] # comment`), a quoted equivalent (`["modules"."bmad-loop"]`), and a multiline string containing a line beginning with `[`. The first two are missed as existing tables, causing a duplicate semantic table to be appended. The multiline case is mistaken for a new table boundary and truncated. Both produce invalid TOML.

2. **High — invalid TOML is written and reported as success** (`merge-config.py:634-661`, `SKILL.md:233`).

   The script overwrites `custom/config.toml` before validation. A parse failure only sets `toml_validated: false`; it still returns `status: success` and exits 0. The setup skill also explains `false` as only meaning `tomllib` was unavailable, even though parse errors take the same path. The documented invocation additionally parses unused legacy YAML after writing TOML, so malformed legacy YAML can leave another partial setup.

3. **Medium — the byte/comment preservation guarantee is false** (`merge-config.py:464-500`, `SKILL.md:20`).

   Comments inside the existing module table and footer comments after the final table are deleted during replacement. CRLF input processed on POSIX is also normalized to LF by text I/O, rewriting unrelated bytes. This contradicts the stated preservation contract for the human-authored custom layer.

4. **Medium — catalog rows assume source and destination headers have identical ordering** (`merge-help-csv.py:331-348`).

   The catalog header is preserved, but source rows are appended positionally without checking or remapping columns. Current v6.10 headers match; a reordered or extended BMAD schema would silently shift help fields into the wrong columns.

5. **Medium — the two help outputs are not committed safely** (`merge-help-csv.py:324-348`).

   The per-module CSV is overwritten before the shared catalog is fully read and staged, and both files are truncated directly. A catalog failure or interruption can leave split registration or damage the shared help catalog.

### Remediation plan

1. Replace the regex-only TOML boundary detection with a TOML-aware scanner, or fail closed whenever semantic parsing finds the module but the editor cannot safely locate its source span.
2. Build and validate the complete candidate before touching disk. Invalid candidates must exit nonzero and leave the original unchanged.
3. Preserve comments, newline convention, permissions, and unrelated bytes; write through a sibling temporary file followed by atomic replacement.
4. Validate CSV headers and row widths, require one source module, and map source rows by column name into the catalog schema. Stage both outputs before replacing either.
5. Add regression tests for quoted/commented headers, multiline values, comments at EOF, CRLF, invalid-candidate rollback, malformed legacy YAML, reordered CSV headers, and partial-write failures. Include an end-to-end assertion using BMADs actual resolver rather than only `tomllib.loads()`.

### Verification

- Focused setup suite: **26 passed**
- Full local suite: **2,944 passed, 7 skipped**
- GitHub CI: all Python 3.11–3.14, Windows, lint, typecheck, and version checks pass
- `git diff --check`: clean

The documented catalog-regeneration caveat, manual pre-rename cleanup, and issues #154/#259 were treated as accepted or out of scope rather than review findings.

pbean:
## Review findings addressed — `9eb5a7f` (code + tests), `6578023` (docs)

All five findings reproduced against the branch and are fixed. One correction to the triage: **finding 1+2 is more severe than "reports success"** — it can leave the project *worse* than not running setup at all, silently.

`resolve_config.py` loads the custom layer as **optional**, so a `TOMLDecodeError` there is swallowed as a warning and merged as `{}`:

```console
warning: failed to parse .../custom/config.toml: Unterminated string (at end of document)
{ "core": { "user_name": "BMad" } }     # our table gone — AND the user's own overrides gone
resolver exit = 0
```

So a corrupt write doesn't just fail to register the module; it silently disables every hand-written override in that file, with a zero exit code from setup *and* from BMAD. That is what drove the fix toward "never write something we haven't verified" rather than "validate harder after writing".

### Findings → resolution

**1. Regex upsert corrupts valid TOML** — replaced with `scan_table_headers()`, a real scanner tracking single/multi-line strings, comments and array/inline-table depth, normalising bare and quoted key paths. All reported cases confirmed and fixed, plus three more found while building fixtures: `'''` literal strings, `#` inside a string, and a multi-line array whose elements start a line with `[`.

**2. Invalid TOML written and reported as success** — inverted to **verify-before-write**. `verify_candidate()` checks the *result*, not the scanner: the candidate must parse and must differ from the original in nothing but `modules.<code>`. Duplicates raise in `tomllib`, truncation fails the parse, a clobbered neighbour fails the diff — so a scanner bug I haven't thought of refuses the run instead of corrupting a config. A refusal writes nothing, leaves the file byte-identical and exits non-zero. An existing file that is already unparseable is refused outright: bmad-loop will not edit a config it cannot reason about.

That needs `tomllib`, so the TOML branch now requires **Python 3.11+** — no real loss, since BMAD's own `resolve_config.py` exits without it, so a v6.10 project cannot function on 3.10 anyway. The legacy YAML branch still runs on 3.9/3.10 and the PEP 723 header is unchanged.

Also fixed: the TOML branch no longer *parses* legacy YAML under `--legacy-dir`. It discarded those values anyway and read them **after** the write, so a malformed `_bmad/core/config.yaml` raised a traceback and reported failure for a registration that had already succeeded.

**3. Preservation guarantee false** — newline convention and file mode are now preserved (CRLF stays CRLF; verified on raw bytes), and the trailing blank/comment back-off applies at EOF too, so a footer comment on the last table survives. In-table comments genuinely do not survive — that table is rewritten wholesale, which *is* the anti-zombie guarantee — so the docs now state the contract that holds instead of over-promising.

**4. Positional CSV append** — `align_rows()` re-keys rows into the target's own column order **by name**, in both branches, with row widths normalised. Columns the target cannot hold are reported in a new `columns_dropped` field and warned on stderr. Identical headers stay the fast path, so nothing changes on a current BMAD tree. The source must now carry exactly one module code on the TOML branch, since `_bmad/<code>/module-help.csv` is by definition one module's file.

**5. Unsafe commit of the two help outputs** — both are rendered in full before either is written, and every write in both scripts goes through a temp file + one atomic rename (mirroring `platform_util.atomic_replace`; duplicated rather than imported, since these ship standalone). `_config/bmad-help.csv` can no longer be left truncated. An `OSError` now surfaces through the JSON error contract instead of a traceback.

### On "an end-to-end assertion using BMAD's actual resolver"

`_bmad/` is gitignored, so a checkout never carries `resolve_config.py` and CI cannot depend on it. Two tests instead: `_resolve_layers()` replicates the documented four-layer merge as the always-on CI assertion, and `test_registration_resolves_through_bmads_own_resolver` runs the **real** resolver when one is available (`BMAD_RESOLVE_CONFIG`, or the project's own `_bmad/`), skipping otherwise. Verified passing against a real v6.10.1 `resolve_config.py`.

### Verification

- **21 new tests**, each confirmed to **fail against the previous code** (18 red on the old scripts) — the corruption cases fail there with real `TOMLDecodeError`s.
- Full suite: **2965 passed, 8 skipped** (was 2944/7; the extra skip is the opt-in real-resolver test).
- `trunk check`: no issues.
- End-to-end on a fresh v6.10 fixture — CRLF custom layer, comments, an unrelated `[agents.bmad-agent-pm]` carrying a multi-line string containing a fake `[modules.bmad-loop]` line, footer comment at EOF — both scripts run **twice**, then BMAD's real resolver:

```json
{
  "modules.bmad-loop": { "name": "BMAD Loop Skills", "version": "0.9.0", "default_selected": false },
  "agents.bmad-agent-pm": { "description": "[modules.bmad-loop]\nnot a real table\n" }
}
```

File still CRLF, footer comment intact, every original line byte-identical, 2 catalog rows after two runs, no legacy files created.

pbean:
## Comprehensive follow-up review

**Recommendation: request changes before merge.** The core issue #258 design is implemented correctly, and no high-severity corruption path remains, but this review found **six behavioral/documentation defects and two verification gaps**. All eight are medium severity.

### Findings

1. **Anti-zombie CSV merging still assumes `module` is column zero.**

   `align_rows()` remaps source rows by header name, but `extract_module_codes()` / `filter_rows()` still read `row[0]`. With a legitimate catalog header such as `skill,module,description`, setup never removes its existing rows. Reproduction: running the merge twice produced **4 `BMAD Loop Skills` rows instead of 2**, with `rows_removed: 0` on the second run.

   **Locations:** `merge-help-csv.py:139-178`, `:461-470`; the reordered-header test at `tests/test_setup_cleanup_scripts.py:1062-1092` keeps `module` at index zero and misses this case.

2. **Concurrent edits can be silently lost.**

   Config and catalog updates perform read → derive → verify → replace without a lock or compare-before-replace guard. Two setup processes—or setup racing an editor/installer—can both validate against state S and then overwrite each other. The fixed sibling temp names (`config.toml.bmad-loop-tmp`, `bmad-help.csv.bmad-loop-tmp`) also collide between concurrent processes.

   **Locations:** `merge-config.py:802-824`, `:929-952`; `merge-help-csv.py:243-285`, `:451-479`.

3. **Atomic replacement destroys symlink-backed configuration.**

   `path.is_file()` / `path.stat()` follow a symlink, but `os.replace(tmp, path)` replaces the symlink directory entry with a regular file. Reproduced result: the local path ceased to be a symlink and contained the new config, while the intended shared target remained unchanged. This silently disconnects centrally managed configuration.

   **Location:** `merge-config.py:802-824` (and the duplicated writer in `merge-help-csv.py`).

4. **An unrelated valid TOML `nan` value makes verification fail.**

   `verify_candidate()` parses the before/after documents independently and compares them with ordinary Python equality. Since `nan != nan`, a valid unrelated value such as `[tool.other]\nvalue = nan` produces the false error “The edit would have changed more than the [modules.bmad-loop] table.” No file is damaged, but setup cannot run on that valid config.

   **Location:** `merge-config.py:739-770`.

5. **Valid alternative TOML representations cannot be upgraded.**

   Existing registrations expressed as dotted keys or inline tables—for example `modules.bmad-loop.name = "old"` or `[modules]\nbmad-loop = { name = "old" }`—are valid TOML but have no explicit `[modules.bmad-loop]` header for the scanner to replace. The script appends that header, `tomllib` detects a duplicate declaration, and setup refuses the edit. Nested module subtables have a related failure mode. The refusal is safe, but the generic “please report this” error gives the user no conversion path.

   **Locations:** `merge-config.py:670-770`.

6. **The documented rollback guarantee is false.**

   `SKILL.md:106` says a non-zero exit always means configuration was left unchanged. In the TOML help branch, however, the per-module CSV is committed before the catalog write. Reproduction: making `_bmad/_config` unwritable caused exit code 2 after `_bmad/bmad-loop/module-help.csv` had been created, while the catalog remained unchanged. The legacy config branch similarly commits `config.yaml` before a possibly failing `config.user.yaml` write, and the skill launches the two scripts in parallel.

   Per-file atomicity is working; transaction-wide rollback is not implemented and should not be promised.

   **Locations:** `SKILL.md:93-106`; `merge-help-csv.py:473-479`; `merge-config.py:1012-1022`.

7. **The config writer's failure path is untested.**

   Tests cover successful mode preservation and pre-write TOML validation failures, but do not force `_atomic_replace()` in `merge-config.py` to fail after its temp file has been written. The catalog test exercises only the separate copy in `merge-help-csv.py`.

   **Locations:** `merge-config.py:802-824`; `tests/test_setup_cleanup_scripts.py:985-1038`.

8. **The real BMAD resolver test skips in required CI.**

   `test_registration_resolves_through_bmads_own_resolver` skips unless `BMAD_RESOLVE_CONFIG` or a local gitignored `_bmad/scripts/resolve_config.py` exists. Normal CI supplies neither, so it runs only the partial local merge oracle. The real installed v6.10 resolver passed during this review, but that integration protection is absent from future required CI runs.

   **Locations:** `tests/test_setup_cleanup_scripts.py:1174-1254`; `.github/workflows/ci.yml`.

### Remediation plan

1. Make source-code extraction and stale-row filtering header-aware; require exactly one `module` column in both schemas. Add a two-run idempotence test with `module` away from column zero, plus missing/duplicate-column cases.
2. Use unique sibling temp files and protect shared read-modify-write operations. At minimum, abort if the on-disk bytes changed before replacement; ideally hold a cross-platform sibling lock across read → verify → replace.
3. Treat symlinks explicitly: either update the resolved target atomically or fail closed with an actionable message instead of silently replacing the link.
4. Replace ordinary parsed-structure equality with recursive NaN-aware equality.
5. Detect a semantically existing module registration with no editable explicit header and either support that representation or return an actionable normalization procedure.
6. Correct the skill/setup-guide guarantee to promise **per-file atomicity**: an earlier output may already have committed, no output is left truncated, and rerunning safely completes the idempotent merge.
7. Add failure-injection coverage for `merge-config.py`'s atomic writer, asserting target byte identity and temp cleanup after a failed replace.
8. Make one required CI job run a pinned real BMAD v6.10 resolver instead of silently skipping the integration test.

### Verification performed

- Reviewed exact range `d4d1764c1e9573145f889610d57978cf8b56a680...f83facbe0e0d1fc99d7124dbb5e43a5fa5a7879f`; head remained stable.
- Full local suite: **2,996 passed, 8 skipped**.
- Focused setup suite: **47 passed, 1 skipped**.
- Real installed BMAD resolver integration: **passed**.
- `trunk check` and `git diff --check`: clean.
- GitHub Linux/Windows tests, lint, typecheck, and version checks: all green.

The current BMAD v6.10 happy path works and the issue #258 acceptance design is present; the findings above are gaps hidden by that green path.

pbean:
## Follow-up review addressed — `3f8ed48` (code), `3884acd` (tests + CI), `40b70d0` (docs)

All eight findings reproduced against `f83facb` and are fixed. **No false positives in the review**, and one addition: the false rollback guarantee was in **two** places, not one — `docs/setup-guide.md:74-75` carried the same sentence as `SKILL.md:106`.

Two corrections to the triage's severity ordering:

**Finding 1 is worse than "setup never removes its existing rows" — it grows the catalog without bound.** The reproduction ran three times, not two:

```console
run1: removed=0 total=3   run2: removed=0 total=5   run3: removed=0 total=7
```

2 rows appended every run, forever, and `rows_removed: 0` each time. That is the failure mode worth naming: `align_rows` re-keys by name, so a filter reading `row[0]` matches *nothing* on a reordered catalog — and a filter that removes nothing is indistinguishable from a first run. It reports success, every time, while `/bmad-help` accumulates duplicate entries. This is the only finding that misbehaves on a tree that never errors.

**Finding 4 is not "setup cannot run on that valid config" in the abstract — `inf` works and `nan` does not**, which is what makes it worth fixing rather than documenting. `[tool.other]\nvalue = inf` passes; swap in `nan` and setup refuses with "the edit would have changed more than the [modules.bmad-loop] table" about a file it never touched outside that table.

### Findings → resolution

| # | Finding | Resolution |
| --- | --- | --- |
| 1 | Anti-zombie merge assumes `module` is column 0 | `module_column()` locates it **by name** for the source, the catalog and the legacy `--target`. Zero or multiple `module` columns are refused, not silently duplicated. |
| 2 | Concurrent edits silently lost; temp names collide | Temp names are per-process/per-call, and `atomic_write_text` re-reads the target immediately before the rename, raising `ConcurrentEdit` if it changed since the merge was derived from it. |
| 3 | Atomic replace destroys symlink-backed config | `resolve_for_write()` follows the link and writes **through** to the real target. A dangling link fails closed. |
| 4 | A valid unrelated `nan` fails verification | `_toml_equal()` recurses with NaN equal to NaN, plus a bool-vs-int guard so `1` never equals `true`. |
| 5 | Alternative TOML representations give no conversion path | `diagnose_unsupported_shape()` names the shape and states the manual fix. |
| 6 | The documented rollback guarantee is false | Both docs now promise **per-file atomicity**, and say "safe to repeat" instead of "nothing was written". |
| 7 | The config writer's failure path is untested | Both windows covered in **both** scripts' copies of the writer. |
| 8 | The real resolver test skips in required CI | New `bmad-resolver` job runs a pinned real v6.10 resolver, and cannot silently degrade to a skip. |

### On finding 5 — why the shapes are still refused

The remediation plan offered "either support that representation or return an actionable normalization procedure". I took the second, deliberately: rewriting a dotted key or an inline table *is* the parse-and-re-emit this whole design exists to avoid, and for the nested case, deleting a hand-written `[modules.bmad-loop.extra]` that may hold the user's own settings is not ours to do. All four shapes (I added **array of tables**, which otherwise landed in the same dead-end message) still refuse with the file byte-identical — the safety net was already working. What changed is that the message names the shape and gives the fix instead of asking the user to report a bug they can solve in ten seconds.

### On finding 2 — scope

Unique temp names + a compare-before-replace guard, not a cross-platform lockfile. The concrete defect is the shared temp path: two runs opened the same `<name>.bmad-loop-tmp`, so one's `finally` unlinked the other's file, or the loser's still-open handle wrote into the inode the winner had just renamed onto the target. That class is gone entirely. The remaining read→derive→verify→replace window is now *detected* rather than held closed — a genuine race aborts with a non-zero exit and an intact file instead of silently discarding the other writer. Duplicating stdlib-only locking into two standalone scripts to close a window that only opens when two humans run setup simultaneously did not seem like the right trade; happy to add it if you disagree.

### On finding 8 — the pin

Verified before wiring it up: the raw file at the pinned tag is **byte-identical** to the `resolve_config.py` installed by BMAD 6.10.1-next.20, so the pin is representative and not a museum piece.

```
https://raw.githubusercontent.com/bmad-code-org/BMAD-METHOD/v6.10.0/src/scripts/resolve_config.py
```

The job fetches it with `curl -f` (an error page cannot masquerade as a resolver) and sets `BMAD_REQUIRE_REAL_RESOLVER=1`, which turns the test's `skip` into a `fail`. Both directions verified locally:

```console
$ BMAD_RESOLVE_CONFIG=<pinned> BMAD_REQUIRE_REAL_RESOLVER=1 pytest -q tests/test_setup_cleanup_scripts.py
65 passed                                        # 0 skipped — it actually runs now

$ BMAD_RESOLVE_CONFIG=/nonexistent BMAD_REQUIRE_REAL_RESOLVER=1 pytest -q -k bmads_own_resolver
Failed: BMAD_REQUIRE_REAL_RESOLVER is set but no resolver exists at /nonexistent/resolve_config.py.
```

Bumping the tag is how we opt in to a newer resolver and find out whether it still reads our registration. It must not float to a branch.

### Verification

- **17 new tests. 14 confirmed red against `f83facb`** — the catalog idempotence tests fail there with 6 and 4 rows where 2 are correct, the `nan` test with a non-zero exit, the symlink test with *"the link was replaced by a regular file"*.
- The other **3 pass against the old code on purpose**: they are finding 7, a *verification gap*, not a defect. The writer worked; nothing exercised it. Both windows are now covered in both scripts — failing at the `open()` (read-only parent) and failing at the rename after the temp file is written and fsynced — each asserting byte-identical target and no temp litter.
- Full suite: **3013 passed, 8 skipped** (was 2996/8).
- `trunk check`: no issues. `git diff --check`: clean.

End-to-end on a fresh v6.10 fixture carrying every hostile condition at once — `custom/config.toml` a **symlink** to a shared file outside `_bmad/`, CRLF, a footer comment at EOF, an unrelated `[agents.bmad-agent-pm]` whose multi-line string contains a fake `[modules.bmad-loop]` line, a valid `nan`, and a catalog with `module` at **column 1** — both scripts run twice, then BMAD's real pinned resolver:

```json
{
  "modules.bmad-loop": { "name": "BMAD Loop Skills", "version": "0.9.0", "default_selected": false },
  "agents.bmad-agent-pm": { "description": "[modules.bmad-loop]\nnot a real table\n" }
}
```

Resolver exit 0, no `failed to parse` warning. Still a symlink pointing at the shared file, still mode `0640`, still CRLF, footer and in-table comments intact, every original line byte-identical, `nan` still there, exactly one *real* `[modules.bmad-loop]` table (counting the literal finds two — the fixture's fake is inside a string), 2 catalog rows after two runs with `rows_removed: 2` on the second, our rows keyed into the right columns, no legacy files created, `core_updated: false`.

One note on that last point, since it bit me while writing the harness: my first end-to-end assertion counted the literal `[modules.bmad-loop]` and failed at 2 — exactly the mistake `test_merge_config_toml_edit_never_corrupts_a_valid_layer` already warns about in its own comment. The scanner was right; the assertion was wrong.

pbean:
## CodeRabbit's three nitpicks — one taken, two declined with reasons

**Taken — per-module help CSV missing from the component table** (`data/skills/README.md:18`) → `fe2b95f`. Correct: setup writes `_bmad/bmad-loop/module-help.csv` as well as the assembled `_config/bmad-help.csv` it is concatenated into, and listing only the catalog made the registration contract look incomplete.

**Declined — "Either tell" → "Either signal"** (`SKILL.md:45`). `tell` is the established voice here, not a typo: `merge-config.py:418` and `merge-help-csv.py:428` both open `detect_layout`'s docstring with *"The tell is a single file: \_bmad/config.toml"*, and `SKILL.md:48` already says *"step 2's tell"*. Changing this one occurrence would make the wording less consistent, not more.

**Declined here, worth a follow-up — error JSON on stdout vs stderr** (`merge-help-csv.py:534`). Real inconsistency, but a **deliberate and pre-existing** one, not something this round introduced: `merge-help-csv.py` matches its own `reject_unresolved_paths` and `cleanup-legacy.py` (stdout), while `merge-config.py` matches its `_fail` (stderr). It was left alone when the two scripts were split precisely to avoid an unrequested behavior change.

Unifying it is a contract change, not a cleanup: it moves where the skill reads a failure from, so it touches `SKILL.md`'s "Write Files" step and every test asserting `json.loads(proc.stdout)["error"]`. That does not belong in a review round scoped to eight specific findings, and picking the winner is a coin flip — an LLM-driven skill arguably wants the error document on **stdout** where it already reads the result, which would mean changing `merge-config.py` instead. Happy to file it as its own issue and settle the direction there.

CI: **11/11 green**, including the new `bmad-resolver` job.

pbean:
Closing unmerged. The implementation grew far out of proportion to the problem — a surgical TOML-scanner, verify-before-write, symlink/concurrency-safe replace machinery and an extra CI oracle job for what is fundamentally 'write the module registration where v6.10 actually reads it'. Not the right direction; #258 stays open for a simpler fix.

Issue #288, opened 2026-07-24
psmux backend: per-server bare-id targets and the pre-#483 rc-0 no-op ceiling


pbean:
Umbrella issue for the psmux target-resolution family. The five children are separate defects on
separate call paths, but four of them are the same root cause — psmux reporting execution rather
than effect — and the fifth is the audit that found them.

## Root cause: ids are per-server, and psmux runs one server per session

On tmux a bare `@N` (window) or `%N` (pane) id is **server-global** and unambiguous. On psmux it
is only meaningful on the server that minted it, and the CLI resolves it against the *caller's
current-session* server. In a normal TUI-launched run there are at least two servers — the
`bmad-loop-ctl` session's and the agent session's — with colliding ids (`@1` exists on both). So
a bare-id `-t` from inside the ctl pane silently binds to the wrong pane and exits 0.

This was already root-caused once, for `switch-client`, in #221 / psmux/psmux#483 and fixed in
#225 by composing a session-qualified target in the backend that knows its own target grammar.
**The fix was applied to one call path, not to the class.** #254 is the same bug on `pipe_pane`.

## Compounding factor: the pre-#483 rc-0 no-op

psmux builds without the #483 fix exit **0** as a no-op when `switch-client` gets an unresolvable
target. Every fallback in this codebase that keys off a nonzero exit code — the POSIX trailer in
`adapters/tmux_base.py` (`switch-client -t "$ret" || switch-client -l`), the pwsh override in
`adapters/psmux_backend.py`, and `BaseTmuxBackend.switch_client` — is therefore *structurally
unreachable* on those builds. The knowledge currently lives only in a CHANGELOG entry.

psmux/psmux#483 is CLOSED, fixed on `main` (d56d777 + 6c76ff9), but **v3.3.7 (2026-07-20) is
still the newest tag** and carries the no-op. Two children are gated on the next release.

## Children

| | | |
|---|---|---|
| #254 | `pipe_pane` targets a bare `@N` → run logs empty on TUI-launched runs | **actionable now** |
| #227 | `return_attached_client` reports success even when `switch_client` fails | **actionable now** (backend-agnostic — affects tmux too) |
| #228 | version-gate the rc-0 no-op + document the fallback ceiling | comment half actionable; version gate needs the tag |
| #222 | live re-verification of the switch-client paths on the next psmux release | **blocked on the next psmux tag** |
[...]

Issue #217, opened 2026-07-20
psmux pipe_pane sink captures nothing on 3.3.7: -EncodedCommand transport is stripped upstream, and the CopyTo sink can't flush incrementally


dracic:
## Summary

On psmux 3.3.7 — the first release that passes the backend's version gate — `PsmuxMultiplexer.pipe_pane` produces a **0-byte run log, always**. Pane output does reach a pipe child's stdin on 3.3.7 (upstream #440 fix is real), but our sink never captures it, for two stacked reasons:

1. **The sink can never start (upstream psmux/psmux#482):** psmux's `pipe-pane` strips every dash-flag token from the piped command before spawning it. Our sink is `pwsh -NoProfile -EncodedCommand <b64>`, which arrives as `pwsh <b64>` and dies instantly. Quoting does not protect the tokens.
2. **Even a running sink would show nothing (our bug, survives any upstream fix):** the raw-copy sink (`[System.Console]::OpenStandardInput().CopyTo($fileStream)`) parks bytes in the `FileStream`'s internal 4 KB buffer, and pipe EOF is not reliably delivered on psmux — so a typical run log stays empty until pane death, or forever. Live-tail consumers (activity-based stall detection reads this log) would see nothing either way.

Both were found and diagnosed during the #185 live smoke (full evidence and the argv-dump matrix there and in psmux/psmux#482).

## Reproduce

Windows 11, psmux 3.3.7 (05cc5d4), `pwsh` 7 on PATH. Drive the real adapter:

```python
b = PsmuxMultiplexer()
b.new_session("repro", Path.cwd())
wid = b.new_window("repro", "w", Path.cwd(), {},
    'pwsh -NoProfile -Command "Start-Sleep 2; Write-Output hello; Start-Sleep 60"')
b.pipe_pane(wid, log)   # exit 0, no warning
# wait any amount of time: log never exists / stays 0 bytes
```

`capture-pane -p -t <wid>` shows the pane output fine, and no sink process ever appears in the process list.

## Expected

The run log receives the pane's raw output stream byte-exact and incrementally (it is live-tailed for activity detection, so "flushed only at exit" is not enough).

## Verified fix (ran end-to-end on 3.3.7)

Write the sink source to a `.ps1` next to the log file and invoke it **positionally** — `pwsh "<log>.sink.ps1"` — with a chunked read/write loop that flushes per chunk:

```powershell
$in = [System.Console]::OpenStandardInput()
[...]

PR #290, opened 2026-07-24
fix(adapters): session-qualify psmux window ids

### LINKED ISSUE #254: psmux: pipe_pane targets a bare @N window id — per-server ambiguity binds the sink to the wrong server's pane (empty run logs)
## Summary

On the psmux backend, per-task pane logs stay empty even though the sink process spawns and the pipe delivers bytes — because `pipe_pane` targets the **bare `@N` window id** returned by `new_window`, and psmux window ids are **per-server** (one server per session). When the engine runs inside a `bmad-loop-ctl` pane (the normal TUI-launched case), its `pipe-pane -t @N` resolves against the **control session's server**, not the agent session's server — so the sink gets attached to whatever window happens to hold id `@N` there (typically the engine's own window), and the agent pane is never piped at all.

This is the same per-server bare-id ambiguity already root-caused for `switch-client` in #221 (psmux/psmux#483: "a bare pane or window id with no session prefix is only resolved within the current session"). The pipe-pane call path was missed.

## Environment

- psmux 3.3.7 (stock crates release 05cc5d4, includes the #440 pipe-pane tee rewrite)
- bmad-loop 0.9.0, Windows 11, engine launched from the TUI (runs inside a `bmad-loop-ctl` window)

## Evidence (live run forensics)

- Sink (`<log>.sink.ps1`, positional invocation from #217) spawned at T+2s and held the log open for 16 minutes, capturing **0 bytes**, while the agent CLI in the real dev pane repainted continuously (`capture-pane` showed the full TUI throughout).
- The only bytes ever captured — 16 bytes, `ESC[?9001l ESC[?1004l` — arrived at the exact moment the engine's own control-session pane was torn down (its ConPTY reset emission), after which the sink hit EOF and exited. That is only possible if the pipe was bound to the engine's own window, not the dev window.
- Both servers in a normal run have colliding ids (e.g. `@1` exists on the ctl server and on the agent-session server), so the misbind is silent: `pipe-pane` exits 0.
- Bonus failure mode, reproduced from inside a pane: `pipe-pane -t @1 -o 'pwsh "<sink>"'` can also fail outright with `psmux: no server running on session '-o'` (the bare-id target derails argument parsing and `-o` is taken as a session name). The adapter's best-effort `except` turns that into a stderr warning that is lost with the parked pane, so the run log is just silently empty either way.
- Raw-mode gating was ruled out as the cause: on the same binary, piped panes capture fine during `?9001h` (win32-input-mode), `?1049h` (alt screen), and 57 KB output bursts.

## Root cause

`GenericCliAdapter.start_session` passes `new_window`'s return value (`#{window_id}`, a bare `@N`) straight to `mux.pipe_pane`. On tmux that id is server-global and unambiguous; on psmux it is only meaningful on the server that minted it, and the CLI resolves it against the caller's current-session server (or worse, see the `-o` parse derail above).

## Proposed fix

Session-qualify the pipe-pane target in the psmux backend, mirroring the #221 resolution for the parked-trailer return target: pipe by `session:window` (verified working on 3.3.7) or `=session:@N`, composed by the backend that knows its own target grammar. Options:

1. `PsmuxBackend.pipe_pane` composes the qualified target itself (it already knows the session via the adapter's `target()` seam) — smallest diff, tmux path untouched.
2. `new_window` on psmux returns a session-qualified id so every downstream `-t` consumer is safe — broader, worth considering since any other bare-id consumer has the same latent bug (an audit of `@N`/`%N` uses on the psmux path is cheap while at it).

## Impact

- Per-task run logs are empty on psmux for TUI-launched runs (engine inside the ctl session), which also blanks the TUI Log tab and starves the pane-log activity signal used by stall detection.
- Completion detection (Stop hook / result artifacts) is unaffected.

Related: #185 (3.3.7 live smoke), #217 (positional sink — still valid, the sink itself is fine), #221 (session-qualified switch-client target), psmux/psmux#483 (bare-id model boundary), psmux/psmux#482

dracic:
## Summary

psmux mints window ids per server (one server per session), so the bare `@N` returned by `new_window` routed by the caller's `$TMUX` when replayed as a `-t` target — from a `bmad-loop-ctl` pane that is the ctl server, not the agent's. The log sink then bound to the engine's own window (empty run logs, dead activity signal), stop/stall nudges were typed into the engine's own pane, and session teardown harvested and killed the engine's window instead of the agent's.

`PsmuxMultiplexer` now returns `session:@N` from `new_window` and emits the identical form from `list_window_ids`. The symmetry is load-bearing: the engine's liveness probe is a membership check (`native_id in list_window_ids(session)`), so qualifying only one side would read every window as instantly dead. With both sides qualified, every target the engine replays — `pipe_pane`, `send_text`, `kill_window`, `window_pane_pids` — routes to the owning server with zero changes to the generic adapter. The tmux path is untouched (its ids are server-global), and the id degrades to bare `@N` when the session name contains `:` (the #221 rule).

This is option 2 from the issue, with the qualification done by post-processing the base methods' return values so the argv construction (and its byte-identity tests) stays untouched.

## How the target grammar was verified (not assumed)

The #483 target-forwarding fixes on psmux main are **not** in the 3.3.7 release, so every form used here was verified against stock 3.3.7 independently:

1. **Source audit at the release tag.** psmux `v3.3.7` (tag == installed binary `05cc5d4`): a `session:`-prefixed target picks the server by the session's `.port` file and forwards the full target out-of-band to that server, where `@N` resolves by window id. Resolution is uniform across `send-keys`, `kill-window`, `capture-pane`, `pipe-pane`, and `list-windows` (a single `parse_target` + temp-focus path); the #483 breakage was specific to `switch-client`'s dedicated handler and does not affect these verbs. Bare targets route by the caller's `$TMUX`, else most-recent-session — the exact misroute this PR fixes.
2. **CLI probes with deliberately colliding ids.** Two live sessions were brought up so both held a window `@2`, then every verb the adapter replays was driven with `sess:@2`: `send-keys` landed only in the minting session's pane, `capture-pane` read the right pane, `pipe-pane -o` with the positional sidecar sink (#217) captured the marker bytes, `list-panes -F '#{pane_pid}'` returned the correct distinct pids, and `kill-window` killed the minting session's `@2` while the other server's `@2` survived. `=sess:@N` also parses (the `=` is stripped), but plain `session:@N` is the only form probed on every verb, so it is the form the backend emits.
3. **End-to-end smoke through the backend itself.** The same colliding-id scenario driven through `PsmuxMultiplexer`: `new_window` returned the qualified id, `window_alive` stayed true, `send_text` reached the right pane, the pipe log grew with the marker, and `kill_window` took down only the minting session's window.

`native_id` lives only in the in-memory `SessionHandle` (never serialized), so no migration is needed for in-flight runs.

## Verification

- New unit tests: qualified mint + qualified list, colon-session degrade on both methods, falsy-id passthrough, `window_alive(new_window(...))` symmetry round-trip, and a transport-failure test pinning that `list_window_ids` still raises rather than reading as "session crashed".
- Full suite on Windows (`PYTHONUTF8=1`, `-n auto`): 2929 passed; remaining failures are the known local skill-sync baseline plus two that require GNU `sleep`/`true` on PATH (pass with Git `usr/bin`, as on CI runners). Full suite on WSL: 2959 passed; `test_opencode_live` failures reproduce identically on the merge base (environmental).
- `ruff check`, `ruff format --check`, `prettier --check CHANGELOG.md`, `pyright@1.1.411` (basic): all clean.

## Out of scope

The TUI-side surfaces (`new_parked_window` return, `list_windows` field ids, ctl-window prune targets) still hand around bare ids — the same ambiguity class, but a different process context. Per the issue's implementation notes this is deliberately excluded here — now filed as #291, together with the bare-id audit #288 asked for. The audit's live probes found the TUI half is **not** a straight port of this recipe: on stock 3.3.7 `select-window` rejects the `sess:@N` form (rc 1) and the window-option verbs silently ignore the window slot (options land session-scoped), so the parked/select paths need name/index targeting — details and the per-verb acceptance matrix are in the #291 comment.

Part of the #288 umbrella (this PR is its #254 child, taken as the class-level shape #288 recommends — qualify at the minting seam so every downstream consumer inherits it).

Closes #254

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed psmux window targeting by session-qualifying window IDs (e.g., `session:`@N``) so replayed operations route to the correct server.
  * Improved window lifecycle/liveness checks and ensured listing timeouts raise proper errors.
  * Preserved behavior when session names contain `:` by falling back to unqualified `@N`.
* **Documentation**
  * Updated the changelog and adapter authoring guide to clarify the opaque window-id/replay contract.
* **Tests**
  * Added tests covering qualified IDs, `pipe-pane` targeting, fallback behavior, and timeout error handling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,19 @@ story <id>`, the same annotation a sweep bundle writes. Both sprint and stories 
 
 ### Fixed
 
+- **psmux window ids are now session-qualified (#254).** psmux mints window ids per server (one
+  server per session), so the bare `@N` the backend returned from `new_window` routed by the
+  caller's `$TMUX` when replayed as a `-t` target — from a `bmad-loop-ctl` pane that is the ctl
+  server, not the agent's. The log sink then bound to the engine's own window (empty run logs,
+  dead activity signal), stop/stall nudges were typed into the engine's own pane, and session
+  teardown harvested and killed the engine's window instead of the agent's. `new_window` and
+  `list_window_ids` now emit the `session:@N` form symmetrically (endorsed in psmux/psmux#483 as
+  the permanent model boundary), so every target the engine replays — `pipe_pane`, `send_text`,
+  `kill_window`, `window_pane_pids` — routes to the owning server unambiguously, and the
+  `window_alive` membership check keeps matching the minted form. Degrades to the bare id when
+  the session name contains `:` (the #221 rule); the tmux backend is untouched (its ids are
+  server-global).
+
 - **A session's read-back could adopt another story's spec (#261).** The generic dev/review
   read-back located "the artifact this session produced" by scanning the implementation-artifacts
   dir for the most-recently-modified qualifying `*.md`, with nothing tying the match to the story
--- a/docs/adapter-authoring-guide.md
+++ b/docs/adapter-authoring-guide.md
@@ -108,9 +108,11 @@ module-level `parse_target()` — or the backend's own **native id** (whatever y
 `target()`. (Two values are composed by the backend itself rather than by
 `target()`, precisely so the seam grammar never has to carry a pane or window
 id: the parked-window return target — `current_return_target`, above — and the
-native window id, which psmux qualifies to `session:@N` at both minting seams
-because its ids are per-server. Both are replayed opaquely; neither is parsed by
-core.) tmux consumes the token natively (it coincides with tmux exact-match
+native window id, which psmux qualifies to `session:@N` across the
+`new_window`/`list_window_ids` pair because its ids are per-server (falling back
+to the bare id where that grammar would not survive — the backend owns those
+conditions, and applies them to both halves of the pair so the symmetry rule
+still holds). Both are replayed opaquely; neither is parsed by core.) tmux consumes the token natively (it coincides with tmux exact-match
 syntax), so `BaseTmuxBackend` passes it straight through. A native-id backend
 calls `parse_target()` first — `None` means "already a native id, use as-is",
 otherwise resolve `(session, window)` yourself; the herdr adapter's
--- a/docs/adapter-authoring-guide.md
+++ b/docs/adapter-authoring-guide.md
@@ -82,7 +82,10 @@ seams of a full OS port are in
   (run a command, then _park_ on a keypress so the exit status stays inspectable,
   then return any attached client to its origin — the POSIX `sh -c` recipe is
   composed from the base's overridable shell-dialect hooks, so a non-POSIX
-  backend swaps the dialect fragments, not the method body), `list_window_ids`, `list_windows` (selected fields per window),
+  backend swaps the dialect fragments, not the method body), `list_window_ids`
+  (which MUST emit the same id form your `new_window` returns — `window_alive` is
+  a membership test over it, so qualifying one side and not the other reads every
+  live window as dead), `list_windows` (selected fields per window),
   `window_alive`, `kill_window`, `select_window`, `set_window_option`,
   `unset_window_option`, `show_window_option`, `pipe_pane` (tee a pane to a log),
   `send_text`.
@@ -102,11 +105,13 @@ families: the **seam-canonical target token** `=session[:window]` — formatted 
 the concrete `TerminalMultiplexer.target(session, window=None)`, decoded by the
 module-level `parse_target()` — or the backend's own **native id** (whatever your
 `new_window` returned). Core never hand-assembles the grammar; it calls
-`target()`. (The parked-window return target is the one value composed by the
-backend itself — `current_return_target`, above — precisely so this grammar
-never has to carry a pane id.) tmux consumes the token natively (it coincides
-with tmux exact-match syntax), so `BaseTmuxBackend` passes it straight
-through. A native-id backend
+`target()`. (Two values are composed by the backend itself rather than by
+`target()`, precisely so the seam grammar never has to carry a pane or window
+id: the parked-window return target — `current_return_target`, above — and the
+native window id, which psmux qualifies to `session:@N` at both minting seams
+because its ids are per-server. Both are replayed opaquely; neither is parsed by
+core.) tmux consumes the token natively (it coincides with tmux exact-match
+syntax), so `BaseTmuxBackend` passes it straight through. A native-id backend
 calls `parse_target()` first — `None` means "already a native id, use as-is",
 otherwise resolve `(session, window)` yourself; the herdr adapter's
 `_parse_target`
--- a/src/bmad_loop/adapters/multiplexer.py
+++ b/src/bmad_loop/adapters/multiplexer.py
@@ -129,6 +129,15 @@ class TerminalMultiplexer(ABC):
         """Create a window running ``command`` (with ``env`` layered on) in
         ``session``, rooted at ``cwd``. Returns the backend-native window id.
 
+        That id is **opaque to core**: it is replayed verbatim as the ``-t``
+        target of :meth:`pipe_pane`, :meth:`send_text`, :meth:`kill_window` and
+        :meth:`window_pane_pids`, and membership-tested against
+        :meth:`list_window_ids` — never parsed, and never re-composed through
+        :meth:`target`. So a backend MAY return an already-qualified target
+        instead of a bare id (psmux returns ``session:@N``), provided
+        :meth:`list_window_ids` emits the identical form — see its symmetry
+        rule.
+
         ``command`` is a POSIX shlex-joined argv string, not a shell line:
         shell-operator behavior (``&&``, ``|``, ...) is backend-defined —
         one backend may hand the string to a shell, another may shlex
@@ -146,6 +155,15 @@ class TerminalMultiplexer(ABC):
     @abstractmethod
     def list_window_ids(self, session: str) -> list[str]:
         """Native ids of every window in ``session`` (empty if it is gone).
+
+        SYMMETRY RULE: these must be the *same id form* :meth:`new_window` and
+        :meth:`new_parked_window` return, because :meth:`window_alive` is a
+        membership test over this list. A backend that qualifies one side and
+        not the other reads every live window as instantly dead. The form
+        itself is the backend's own — psmux emits ``session:@N`` because its
+        ids are minted per server (one server per session), so a bare ``@N``
+        replayed as a ``-t`` target routes by the *caller's* server instead of
+        the owning one.
 
         Raises :class:`MultiplexerError` if the transport itself fails (timeout /
         missing binary): an empty list means "no windows" and must not be
--- a/src/bmad_loop/adapters/multiplexer.py
+++ b/src/bmad_loop/adapters/multiplexer.py
@@ -156,14 +156,19 @@ class TerminalMultiplexer(ABC):
     def list_window_ids(self, session: str) -> list[str]:
         """Native ids of every window in ``session`` (empty if it is gone).
 
-        SYMMETRY RULE: these must be the *same id form* :meth:`new_window` and
-        :meth:`new_parked_window` return, because :meth:`window_alive` is a
-        membership test over this list. A backend that qualifies one side and
-        not the other reads every live window as instantly dead. The form
-        itself is the backend's own — psmux emits ``session:@N`` because its
-        ids are minted per server (one server per session), so a bare ``@N``
-        replayed as a ``-t`` target routes by the *caller's* server instead of
-        the owning one.
+        SYMMETRY RULE: these must be the *same id form* :meth:`new_window`
+        returns, because :meth:`window_alive` is a membership test over this
+        list. A backend that qualifies one side and not the other reads every
+        live window as instantly dead. The form itself is the backend's own —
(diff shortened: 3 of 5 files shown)


pbean:
Reviewed and validated — **the approach is right and stays inside #254's scope**; I pushed one follow-up commit (95c667f) for a contract gap rather than asking for changes.

## Approach: endorsed, and correctly scoped

This is option 2 verbatim from #254 and the #288 umbrella ("have `new_window` return a session-qualified id so *every* downstream `-t` consumer is safe … the class-level fix"). It honors #254's trap #1 (both minting and listing qualified, so `window_alive`'s membership check keeps matching), honors the scope decision (TUI-side surfaces excluded, filed as #291), keeps composition backend-owned per the #221/#225 precedent, and leaves the tmux leaf structurally untouched. `native_id` is never serialized, so no in-flight migration. The `:`-degrade is unreachable in practice (`session_name()` is `bmad-loop-<ts>-<hex>`) — fine as defensive parity with the #221 rule.

## The 3.3.7 grammar claim, independently corroborated

Since the #483 fixes are not in the release, I audited psmux source **at tag `v3.3.7`** rather than take the probes on faith:

- `src/cli.rs:523 parse_target` — strips a leading `=`, splits at the first `:`, and an `@N` in the window slot sets `window=N, window_is_id=true`. So `s:@2` → session `"s"`, window id 2.
- `src/main.rs:281-315` — `has_explicit_session` ⇒ `PSMUX_TARGET_SESSION` is set from the session part and the server is chosen by its `.port` file (the full target rides in `PSMUX_TARGET_FULL`). A **bare** `@N` gives `session=None` and falls through to `resolve_routing_target($TMUX, …)` → caller's server, else most-recent-session. Exactly the misroute, exactly fixed.
- `src/server/connection.rs:772-905` — target resolution is **one shared pre-dispatch path** (`FocusWindowByIdTemp` when `window_is_id`) applied to every verb that is not `select-window`/`select-pane` or a move/swap/join. `pipe-pane`, `send-keys`, `kill-window`, `list-panes`, `capture-pane` all inherit it. The uniformity claim holds — and it also explains the `select-window` rc-1 you found for #291, since that one takes the separate focus-cmd branch.

I could not re-run the byte-level probes (psmux is Windows/ConPTY, not installed on my box); the routing *mechanism* is confirmed from source at the tag, the per-verb live results rest on your report plus CI.

## What I pushed (95c667f)

The change establishes a **cross-backend invariant and records it nowhere a backend author reads** — that was the one real finding:

1. `multiplexer.py` — the symmetry rule as a `SYMMETRY RULE:` block on `list_window_ids`, plus an "opaque to core" sentence on `new_window` (replayed verbatim as `-t`, membership-tested, never parsed, never re-composed through `target()`). Your #254 notes called this trap #1; the ABC was silent on it, so the herdr adapter author had no way to know.
2. `docs/adapter-authoring-guide.md` — the guide said the parked-window return target is *"the one value composed by the backend itself"*. psmux now composes two; corrected, and the symmetry rule added to the Windows bullet.
3. `psmux_backend.py` — two comments: the TUI-side mint/list surfaces deliberately stay bare (#291), and why `_qualified_window_id` hand-builds `session:@N` instead of calling `self.target()` (`target()` must stay a by-name token, `@N` is an id — do not unify the two grammars).
4. `tests/test_psmux_backend.py` — `test_qualified_id_reaches_the_pipe_pane_target`. All six of your tests assert on return values; #254 is about the `-t` argv. This pins that a minted id reaches `pipe-pane -t s:@2` verbatim through the one verb this backend reimplements.

Verified by ablation (each mutation applied, the named test confirmed failing, then reverted): deleting the `new_window` override fails the mint test **and** the new argv test; deleting the `list_window_ids` override (= qualifying only one side) fails the list test and the symmetry test; dropping the `":" in session` guard fails the degrade test; dropping the falsy guard fails the sentinel test. Full suite 3060 passed (the 2 failures are pre-existing local skill-sync drift in untracked install copies); `trunk check` clean.

## One process note

The `CodeRabbit` check reads **pass**, but its comment says *"Review limit reached … we couldn't start this review."* That is a rate-limit no-op, not a review — we've been burned by that fake-green before. Triggering a real one below; merging once it comes back.

@coderabbitai review

pbean:
@coderabbitai Partly taken, in `5675eed` — and checking your comment surfaced a real inaccuracy of mine that you were pointing at without naming.

**Taken:** `docs/adapter-authoring-guide.md` said psmux qualifies "at both minting seams". That names the wrong pair — `new_parked_window` *is* a minting seam and is deliberately **not** qualified (#291), while `list_window_ids` does not mint at all. Corrected to name `new_window`/`list_window_ids` explicitly, and the fallback is now noted there: bare-id fallback conditions are backend-owned and applied to **both halves of the pair**, so the symmetry rule holds under them too.

**Not taken:** duplicating psmux's colon condition into the two `multiplexer.py` docstrings. Three reasons:

1. **The invariant is unconditional; only the form is conditional.** The rule the ABC states is "`list_window_ids` must emit the same form `new_window` returns". The degrade does not weaken it — both sides degrade identically, which is exactly what `test_qualification_degrades_to_bare_on_colon_session` pins. Writing "qualifies *only when* the session name has no colon" into the symmetry rule invites the opposite misreading: that the *symmetry* is conditional. It never is.
2. **Nothing actionable changes for the ABC's reader.** That reader is writing a new backend. They need "pick a form, emit it from both seams". psmux's colon policy is one backend's internal choice; knowing it does not change a line they write. The condition already lives where it is actionable — `PsmuxMultiplexer._qualified_window_id`, with the #221 reference and the reasoning.
3. **Restating it in three places is the failure mode we deliberately avoid here.** A condition with three copies drifts; the seam names the rule, the backend owns the conditions.

The `session:@N` mentions in the ABC are parenthetical illustrations of "a backend MAY return an already-qualified target", and the surrounding sentence already says "the form itself is the backend's own" — so it does not claim the form is unconditional.

pbean:
@greptileai review

pbean:
Greptile's P1 — **doc point taken in `643434e`; severity downgraded to a documentation defect.** It caught a real self-contradiction I introduced, and pointed at it more precisely than I'd noticed.

## The finding is right, and it's mine

`95c667f` wrote the SYMMETRY RULE as binding `new_window` **and `new_parked_window`** — while this same PR ships a psmux backend that deliberately does not qualify parked ids. The ABC stated an invariant its own backend violates.

Worse, `5675eed` already made *exactly this correction* — to the other document. Its message: "`new_parked_window` IS a minting seam and is deliberately NOT qualified (#291)". I fixed `docs/adapter-authoring-guide.md` and left the ABC docstring, which is the copy a backend author actually reads. Greptile found the half I missed.

## But the runtime bug is unreachable, not latent-and-lurking

Traced end to end rather than assumed:

- `new_parked_window` has **one** caller in `src/`: `tui/launch.py:296` (`start_detached`). Its id goes to `set_window_option`; three callers discard it, one (`start_resolve_detached`) hands it to the TUI. It is never stored on a dataclass.
- Both `window_alive` consumers are fed `new_window`-minted ids only — `generic.py:987` (`handle.native_id`, set at `generic.py:499`) and `probe.py:505` (id from `probe.py:497`).
- `tests/test_multiplexer.py:86-87` and `:91-92` are structural guards: the adapter's `StubMux` raises `AssertionError` for **both** `new_parked_window` and `window_alive`. The engine path cannot mint a parked id, or membership-test one, without those firing.

So the comment's own hedge is the accurate reading — "if `window_alive` is never called with a parked-window id, the bug is latent". It is never called, and the separation is pinned rather than incidental.

## And the implied remedy would be worse than the defect

Overriding `new_parked_window` to satisfy the rule as written is **verified broken** on stock 3.3.7, per this PR's own #291 audit: `select-window` rejects the `sess:@N` form (rc 1), and the window-option verbs silently ignore the window slot (options land session-scoped). `set_window_option` is precisely what `start_detached` does with that id, on every detached launch. The docstring as written pointed a future backend author at a change that breaks the TUI — which is the strongest argument for fixing the text rather than the code.

## What changed

`643434e`, docs-only, one file. The rule now binds the pair core actually membership-tests (`new_window` / `list_window_ids`), plus a short clause naming `new_parked_window` as *deliberately* outside it and why — so the omission reads as a decision, not an oversight. Single copy: the guide already scopes it correctly, and `psmux_backend.py`'s module docstring already records the exemption on the backend side. All three now agree.

No behavior change; #291 keeps the TUI-side surfaces. For the record, the surfaces this finding gestures at are already enumerated there in full — `set_window_option(PROJECT_OPTION)` at `launch.py:310`, `select_window` at `:72`, `set_window_option(RETURN_OPTION)` at `:105`, and the independent `list_windows` → `win_id == current` → `kill_window` prune path — including the `current_window_id()` like-with-like hazard. No new issue needed.

`pytest tests/test_multiplexer.py tests/test_psmux_backend.py` 62 passed; `trunk check --no-fix` clean.

---

**Process note, again:** the `CodeRabbit` check reads **pass** but is still the rate-limit no-op — `gh pr checks 290` shows `pass 0 "Review rate limited"`. CodeRabbit's last *actual* review was of `95c667f`, so `5675eed` and `643434e` have never been reviewed by it. Not merging on that green.

@coderabbitai review

Issue #287, opened 2026-07-24
Adopt the BMAD v6.10 consolidated TOML config layout end-to-end


pbean:
Umbrella issue. No new requirements of its own — it exists so the three children read as one
adoption rather than three unrelated defects.

## The shift

BMAD-METHOD v6.10.0 resolves module config from a **consolidated four-layer TOML** stack, read by
`_bmad/scripts/resolve_config.py` (`tomllib`) in this precedence order:

```
_bmad/config.toml → _bmad/config.user.toml → _bmad/custom/config.toml → _bmad/custom/config.user.toml
```

with `[modules.bmm]` beating `[core]` on key collision. Help is consolidated at
`_bmad/_config/bmad-help.csv` plus per-module `_bmad/<module>/module-help.csv`.

The legacy per-module `_bmad/bmm/config.yaml` and a root `_bmad/module-help.csv` are no longer
read by the runtime.

## Where bmad-loop still assumes the legacy layout

bmad-loop is on the wrong side of this on **both** the read and the write path:

- **Read** — `bmadconfig.load_paths` (`src/bmad_loop/bmadconfig.py:70-72`) resolves
  `implementation_artifacts` / `planning_artifacts` from `_bmad/bmm/config.yaml` only. → #154
- **Write** — the `bmad-loop-setup` skill's `merge-config.py` / `merge-help-csv.py` write
  `_bmad/config.yaml` and a root `_bmad/module-help.csv`: files the runtime never reads. → #258
- **Docs** — the same skill's docs invoke its PEP 723 scripts with bare `python3`, which fails on
  the inline-dependency scripts. → #259

The installer regenerates the per-module YAML from the TOML on each install/update, so the two
surfaces normally agree — which is exactly what makes the divergence quiet. A hand-added
`custom/config*.toml` override desyncs them until the next install, and from that moment the
skill writes specs and the deferred-work ledger where the orchestrator is not looking.

## Children

- #154 — `bmadconfig`: read the four-layer TOML as primary, legacy YAML as fallback
- #258 — `bmad-loop-setup` writes legacy `config.yaml` / `module-help.csv`
- #259 — `bmad-loop-setup` docs invoke PEP 723 scripts with `python3`

Each is independently ship-able; #259 in particular is a small standalone docs fix and should not
wait on the other two.

pbean:
[...]

PR #297, opened 2026-07-25
fix(setup): stop registering BMAD config; the installer owns it (#258, #259)

### LINKED ISSUE #258: bmad-loop-setup writes legacy config.yaml/module-help.csv; ignored on BMAD v6.10.0 TOML config layout

### LINKED ISSUE #259: bmad-loop-setup docs invoke PEP 723 scripts with python3; merge-config.py fails without uv run

pbean:
Closes #258. Closes #259. Part of #287.

## The problem

`bmad-loop-setup` wrote `_bmad/config.yaml`, `_bmad/config.user.yaml` and a root
`_bmad/module-help.csv` — the pre-v6.10 layout. BMAD's resolver never reads any of them:
`resolve_config.py` merges four TOML layers, and `/bmad-help` reads a catalog assembled from
per-module `_bmad/<module>/module-help.csv`. The scripts produced dead files.

## Why this is subtraction, not a TOML port

An earlier attempt (#285, closed unmerged) tried to port the writer to TOML. Reading upstream
shows why that direction was wrong:

- **bmad-loop is already an installer-installed module** — registered in BMAD-METHOD's
  `bmad-modules.yaml` with `marketplace-plugin: true`, shipped as the v6.10.0 headline feature.
- **The installer does the registration**, on every install *and* quick-update: stages
  `_bmad/bmad-loop/` with all three skill dirs, copies `assets/module-help.csv` to that module
  root, writes the per-module `config.yaml`, adds rows to `manifest.yaml` / `skill-manifest.csv` /
  `files-manifest.csv`, and rebuilds `_config/bmad-help.csv` by scanning `_bmad/*/`.
- **There was never a `[modules.bmad-loop]` table for us to write.** `writeCentralConfig` skips
  modules with no prompt answers, and our `module.yaml` declares none. `config.toml` is also
  regenerated wholesale each install — anything written there from outside is discarded.
- **Upstream is retiring this pattern rather than porting it** — bmad-builder#96
  (*"the setup skill creation will be removed soon"*) and BMAD-METHOD#2591.

## What changed

**The skill.** `merge-config.py`, `merge-help-csv.py` and `cleanup-legacy.py` are removed.
Setup writes exactly one file under `_bmad/` — the per-module `module-help.csv`, byte-for-byte
what the installer places there, surviving its reruns and idempotent. `SKILL.md` keeps only what
the installer structurally cannot do: install or upgrade the orchestrator tool, run
`bmad-loop init`, preflight with `validate`. It reads `user_name` / `communication_language`
through BMAD's own `resolve_config.py` instead of collecting and storing them.

**`assets/` is deliberately untouched.** `plugin-resolver.js` strategy 2 requires
`assets/module.yaml`, `assets/module-help.csv` *and* a directory name ending in `-setup`, or
`official-modules.js` throws and the module stops installing. Nothing reads those files at
runtime, so only a test keeps them honest — that test is new here.

**#259 closes with them.** No inline-dependency script is invoked with bare `python3` any more,
because none ships. The one remaining script call is the resolver, via `uv run`.

**The bmad-auto rename compatibility is gone** — dead since 0.8.0, with no installs left in the
wild. `init` no longer strips `bmad_auto`-marked hooks, deletes `bmad-auto-*` skill dirs, carries
`.automator/policy.toml` forward, or prints the leftover-state note. That also retires
`cleanup-legacy.py`'s last legitimate caller: its other invocation targeted `_bmad/bmad-loop/`,
which is now the installer's staged skill payload *and* its copy source for `.claude/skills/`.

## Tests

`test_setup_cleanup_scripts.py` guarded "setup must never delete live BMAD config" (#64/#73);
with no deleting code shipping, that regression is impossible. It is replaced by
`test_setup_skill_contract.py`, which pins the installer's resolver contract and guards against
prose drifting back toward the legacy paths. The strip-hook tests' one unique invariant — a
foreign `Stop` hook coexisting with our relay — was already covered by
`test_merge_hooks_preserves_existing`, verified before deleting.

## Verification

End-to-end against a real v6.10.1 `_bmad/` tree, running the documented steps twice:

- 44 pre-existing files byte-identical; `config.yaml`, `config.user.yaml` and the root
  `module-help.csv` unchanged and uncreated
- exactly one file created, matching `assets/module-help.csv` byte-for-byte
- **`resolve_config.py` used as the oracle**: output byte-identical before and after, rc 0

Double-`init` hook regression (the riskiest edit, since removing the strip meant collapsing the
write conditionals): registers on the first run, reports already-registered on the second, and a
pre-existing foreign `Stop` hook plus `permissions` survive both.

```
pytest -q                  2960 passed, 1 skipped
trunk check --no-fix       No issues
seed_skills.py --check     every skill fork matches canonical
```

## Not in scope

#154 (the runtime still resolves artifact paths from `_bmad/bmm/config.yaml`) stays open under
#287.

## Summary by CodeRabbit

* **New Features**
  * Streamlined project setup now installs or upgrades the orchestrator, initializes project wiring, and validates the configuration.
  * Setup refreshes module help information while leaving broader BMAD configuration management to the BMAD installer.

* **Bug Fixes**
  * Existing project policy settings are preserved during upgrades.
  * Removed outdated migration behavior and legacy cleanup guidance.

* **Documentation**
  * Updated setup, upgrade, and uninstall instructions to reflect the simplified workflow.


pbean:
@coderabbit review

Issue #194, opened 2026-07-19
Dev session lost to an API/transport failure is charged as a story timeout, exhausting the attempt budget


tjoignant:
## Environment

- bmad-loop 0.8.1 (uv tool install, commit `5caac1c`), macOS 24.6.0, tmux
- adapter: `claude`, `scm.isolation = "worktree"`, `branch_per = "story"`
- `limits.session_timeout_min = 90`, `max_dev_attempts = 2`

## The bug

A dev session that dies because the agent CLI lost its API connection is indistinguishable
from one that genuinely ran out of wall clock. Both surface as `status="timeout"`, and both
consume a dev attempt. Two consecutive transport failures therefore exhaust
`max_dev_attempts` and defer a story on which **no dev work was ever attempted**.

The CLI process stays alive after the API error — it just stops making progress — so the
liveness probe never sees a dead window and never returns `crashed`. The session sits idle
until `spec.timeout_s` stamps it `timeout`.

`env_fault` already exists as the concept that would cover this, but it cannot reach this
path:

- `engine.py:1628` — `outcome` is only computed `if result.status == "completed"`.
- `engine.py:1671` — the journal therefore records
  `env_fault=bool(outcome is not None and outcome.env_fault)` → always `False` for a timeout.
- `escalation.py:71-75` — `decide_dev` charges an attempt for any non-completed status,
  with no env-fault branch at all:

  ```python
  if result.status != "completed":
      reason = f"dev session {result.status}"
      if budget_left:
          return Decision(Action.RETRY, reason)
      return Decision(exhausted, _exhaust_reason(task, reason))
  ```

Nothing in `adapters/generic.py`, `adapters/base.py`, or `engine.py` matches `API Error`,
`Unable to connect`, or `ConnectionRefused` — the transport failure is never classified.
Verified against current `main`, not only 0.8.1.

## Evidence

Run `20260718-191419-44e7`, story `3-7-ingestion-eval-harness-and-no-retention-lifecycle`,
both attempts lost to transport failures:

**dev-1** — 75 minutes of clean, productive work (implementation → verification →
four parallel review layers → triage; zero tool errors), then in the transcript:

```
01:37:03  API Error: Unable to connect to API:
          Self-signed certificate detected. Check your proxy or corporate SSL certificates
```

Idle for the remaining 13 minutes. Journal: `status="timeout"`, `tokens=12558555`,
[...]

PR #301, opened 2026-07-25
test(win32): use a portable no-op verify command (#292)

### LINKED ISSUE #292: Windows: env-fault engine test replays check.sh through cmd — pops the 'How do you want to open this file?' dialog on every local full-suite run
## Summary

`tests/test_engine.py::test_fix_phase_session_env_fault_escalates` builds a POSIX verify script (`check.sh` with a `#!/bin/sh` shebang) and puts `"<path>/check.sh"` into `VerifyPolicy(commands=...)`. The engine runs verify commands through the host shell — `cmd /c` on Windows — and cmd hands a document path to ShellExecute. On a machine with no `.sh` association (stock Windows), that pops the interactive "How do you want to open this file?" picker in the middle of the test run.

Its three sibling tests in the same file (`test_review_verify_env_fault_escalates_instead_of_fix_session`, `test_skip_review_env_fault_escalates_not_defers`, `test_fix_phase_env_fault_escalates_instead_of_looping`) all carry `@pytest.mark.skipif(sys.platform == "win32", ...)`; this one does not.

## Why it goes unnoticed

The test still *passes* on Windows — the fault it asserts comes from the injected `SessionResult(env_fault=True)`, not from the script's exit code — so CI stays green and nothing flags it. The only symptom is a GUI dialog stealing focus on every local full-suite run (observed on Windows 11 during the #254 verification runs; confirmed root cause via `cmd /c assoc .sh` → "File association not found").

## Suggested fix

The test's subject (#194: a fix session that lost its API connection escalates instead of burning dev budget) is not POSIX-specific, so a skip would lose real Windows coverage. Prefer a per-OS script, mirroring the platform split `tests/conftest.py` already uses for verify commands:

```python
if sys.platform == "win32":
    script = project.project / "check.cmd"
    script.write_text("@exit 0\r\n", encoding="utf-8")
else:
    script = project.project / "check.sh"
    script.write_text("#!/bin/sh\nexit 0\n", encoding="utf-8")
    script.chmod(0o755)
```

Refs: #194

dracic:
Closes #292

> **Note:** the approach changed during review (see [the review comment](https://github.com/bmad-code-org/bmad-loop/pull/301#issuecomment-5081837870)); this description was updated by a maintainer to match what ships. The original diagnosis, the Windows reproduction, and the `write_script_launcher` fix are @dracic's.

## Problem

`test_fix_phase_session_env_fault_escalates` put a POSIX `check.sh` into a `VerifyPolicy` command tuple. Verify commands run through the host shell (`run_verify_commands` is `shell=True`, `verify.py:1379`), so on Windows `cmd` hands that path to ShellExecute — two failure modes, neither of which the test can see:

- **no `.sh` association** (stock Windows): the interactive "How do you want to open this file?" picker pops mid-run and steals focus.
- **`.sh` associated** (e.g. Git for Windows): the associated app launches and the command returns success without the script ever running — a silent false pass.

Either way the test passes, because the fault it asserts comes from an injected `SessionResult(env_fault=True)`, so CI stayed green and only local full-suite runs saw the dialog.

## Fix

Two commits, both test-only.

**1. `tests/test_engine.py`** — drop the script entirely and use conftest's `_OK` as the second verify command.

The script was inert: written once as `exit 0`, never mutated, referenced by no assertion. `run_verify_commands` does not short-circuit, so its whole contribution was rc 0, twice. It is copy-paste residue from the sibling `test_fix_phase_env_fault_escalates_instead_of_looping` (`tests/test_engine.py:5314`), which `chmod(0o644)`s it mid-run to produce rc=126 — there it is the point. Here the only verify signal is the marker command, and the escalation is the session's.

`tests/conftest.py:79-82` is the sanctioned spot for this — the block headed *"host-shell verify/lifecycle stub commands (single platform-detection spot)"*, whose comment reads *"These build them per-OS in one place instead of each test file re-deriving the win32 branch."* It exports `_OK = "exit 0"`, already imported by `tests/test_engine_worktree.py:14`.

```python
-    script = project.project / "check.sh"
-    script.write_text("#!/bin/sh\nexit 0\n", encoding="utf-8")
-    script.chmod(0o755)
...
-        verify=VerifyPolicy(commands=(_file_exists_cmd(marker), f'"{script}"')),
+        verify=VerifyPolicy(commands=(_file_exists_cmd(marker), _OK)),
```

Net +5/−5, no `sys.platform` branch in the test. A win32 skip was also rejected: the subject (#194 — a fix session that lost its API connection escalates instead of burning dev budget) is not POSIX-specific, unlike the three sibling tests, which skip on win32 because they depend on `sh` rc 126/127.

**2. `tests/conftest.py`** — `write_script_launcher`'s win32 launcher wrote an explicit `\r\n`. `Path.write_text` uses `newline=None` and already translates to CRLF on Windows, so the explicit `\r` landed on disk as `\r\r\n` (reproduced). Harmless for a one-line launcher, and unrelated to #292 once the test stops hand-rolling a launcher — kept as an acknowledged drive-by because the defect is real and the fix is correct.

## Verification

- Windows 11 (original per-OS revision): `pytest tests/test_engine.py` → 194 passed, 3 skipped, no dialog; `pytest tests/test_engine_plugin.py tests/test_opencode_http.py` (the `write_script_launcher` callers) → 112 passed, 4 skipped
- Linux: `pytest tests/test_engine.py` → 198 passed; `-k env_fault` → 6 passed, 0 skipped (unchanged from baseline)
- Linux: full suite → 2960 passed, 1 skipped; `trunk check` clean
- CI: all 11 checks green, including both Windows jobs

Two ablations, since nothing in the suite catches this class on its own — the test's assertions never depended on the script:

- `env_fault=True` → `False` fails the test (the escalation path is still load-bearing)
- `_OK` → `exit 3` fails the test (the second command is genuinely executed, so `_OK` occupies the same slot the script did)

What `_OK` buys over a per-OS branch is that the failure mode becomes structurally impossible rather than handled: no script file exists for any shell to mis-handle.

## Scope

The three other inline `.sh` verify sites (`_self_disarming_script` and `test_fix_phase_env_fault_escalates_instead_of_looping`) are already `@pytest.mark.skipif(sys.platform == "win32")`, so this was the only unguarded one. `test_install.py`'s POSIX-script sites are likewise win32-skipped, and the git-hook sites run under git's own `sh`.

No CHANGELOG entry: test-only.

## Follow-up: #302

One finding from this review is deliberately out of scope and filed as its own issue: `ENV_FAULT_RCS = frozenset({126, 127})` never matches on Windows — `cmd` reports a missing or non-executable verify tool as 9009/1, so the charged-attempt behavior #126 exists to prevent still happens on win32, and the three skips above hide it. Reclassifying rc 1 carries a real false-positive risk (under `cmd` it is also the ordinary "tests failed" code), so it needs its own decision rather than riding a test-only fix.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -64,7 +64,9 @@ def write_script_launcher(directory: Path, name: str, body: str) -> Path:
     sidecar.write_text(body, encoding="utf-8")
     if sys.platform == "win32":
         launcher = directory / f"{name}.cmd"
-        launcher.write_text(f'@"{sys.executable}" "{sidecar}" %*\r\n', encoding="utf-8")
+        # `\n`, not `\r\n`: write_text translates it to the CRLF cmd wants, so an
+        # explicit `\r` would land on disk doubled (`\r\r\n`).
+        launcher.write_text(f'@"{sys.executable}" "{sidecar}" %*\n', encoding="utf-8")
     else:
         launcher = directory / name
         launcher.write_text(
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -11,6 +11,7 @@ from pathlib import Path
 
 import pytest
 from conftest import (
+    _OK,
     _file_exists_cmd,
     _spec_baseline,
     committing_crash_state,
@@ -4739,9 +4740,6 @@ def test_fix_phase_session_env_fault_escalates(project):
     burning the remaining dev budget on repair sessions that never ran."""
     write_sprint(project, {"1-1-a": "ready-for-dev"})
     marker = project.project / "fixed.marker"
-    script = project.project / "check.sh"
-    script.write_text("#!/bin/sh\nexit 0\n", encoding="utf-8")
-    script.chmod(0o755)
 
     def dev_with_marker(spec):
         marker.write_text("ok\n")
@@ -4757,7 +4755,10 @@ def test_fix_phase_session_env_fault_escalates(project):
     policy = Policy(
         gates=GatesPolicy(mode="none"),
         notify=QUIET,
-        verify=VerifyPolicy(commands=(_file_exists_cmd(marker), f'"{script}"')),
+        # `_OK`, not a script file (#292): verify commands run through the host shell,
+        # and cmd hands a `.sh` path to ShellExecute instead of running it. Only the
+        # marker command carries signal here — the escalation is the session's.
+        verify=VerifyPolicy(commands=(_file_exists_cmd(marker), _OK)),
         limits=LimitsPolicy(max_dev_attempts=3),  # budget left -> must not be spent
     )
     engine, adapter = make_engine(


pbean:
Diagnosis is right and the Windows verification is appreciated — I reproduced the reasoning end to end: `run_verify_commands` is `shell=True` (`verify.py:1379`), so `cmd` really does hand `"…\check.sh"` to ShellExecute, and the test can't see it because the fault it asserts comes from the injected `SessionResult(env_fault=True)`. Both of your side observations check out too: `Path.write_text` uses `newline=None`, so an explicit `\r\n` genuinely lands as `\r\r\n`, and `exit /b` is the correct form in a batch file. Your scope claim also holds — I swept `tests/` and the other POSIX-script sites are either `skipif(win32)` or run under git's own sh.

I pushed an amendment that changes the approach, so flagging it here rather than merging silently.

**The script in that test is inert.** It's written once as `exit 0`, never mutated, and no assertion references it. `run_verify_commands` doesn't short-circuit, so its entire contribution is rc 0, twice. It's copy-paste residue from the sibling `test_fix_phase_env_fault_escalates_instead_of_looping` (`tests/test_engine.py:5314`), where the script *is* `chmod(0o644)`-ed mid-run to produce rc=126 — there it's the whole point. Here the only verify signal is the marker command, and the escalation is the session's.

**And `tests/conftest.py:79-82` is the sanctioned spot for this.** That block is headed *"host-shell verify/lifecycle stub commands (single platform-detection spot)"* and its comment says: *"These build them per-OS in one place instead of each test file re-deriving the win32 branch."* It exports `_OK = "exit 0"`, already imported by `tests/test_engine_worktree.py:14`.

So the amendment drops the file instead of branching for it:

```python
-    script = project.project / "check.sh"
-    script.write_text("#!/bin/sh\nexit 0\n", encoding="utf-8")
-    script.chmod(0o755)
...
-        verify=VerifyPolicy(commands=(_file_exists_cmd(marker), f'"{script}"')),
+        verify=VerifyPolicy(commands=(_file_exists_cmd(marker), _OK)),
```

Net +5/−5, no `sys.platform` branch, and the failure mode becomes structurally impossible rather than handled per-OS — there's no script file for any shell to mis-handle.

**Your conftest commit is kept as-is**, authorship intact. Its stated rationale (not symptom-patching the shared helper) no longer applies once the test stops hand-rolling a launcher, but the `\r\r\n` defect is real and the fix is correct — dropping a correct one-line fix for scope purity would be the worse trade. It's now an acknowledged drive-by.

Verification on this end:
- `pytest tests/test_engine.py` → 198 passed; `-k env_fault` → 6 passed, 0 skipped (unchanged from baseline)
- full suite → 2960 passed, 1 skipped; `trunk check` clean
- two ablations, since nothing in the suite can catch a regression here on its own: flipping `env_fault=True` → `False` fails the test (it's still load-bearing), and swapping `_OK` → `exit 3` also fails it (the second command is genuinely executed, so `_OK` occupies the same slot the script did)

One thing left to you: the PR description now documents the superseded per-OS approach — feel free to update it, I'd rather not edit your body. #302 stays open and untouched; the reasoning there for not riding it on a test-only fix is right.

PR #307, opened 2026-07-25
feat(opencode): live attached-UI viewer in the dashboard (show_attached_ui)

Wrenbjor:
## What

`[adapter] show_attached_ui = true` gives the opencode-http adapter a live viewer: a detached tmux window running `opencode attach`, rendered in the dashboard Log tab — so an OpenCode run is watchable like the tmux adapters' sessions.

## Why

opencode-http drives sessions over HTTP with no pane to observe; the Log tab could only show the raw server log, leaving runs effectively headless.

## How

- New `viewer_launch` module: parked, authenticated (password via tmux session environment, never argv), verified-live viewer window; failures land in `viewer-error.log` + a `viewer-create-failed` journal event, and `viewer.json` never records a dead target
- New mux seam ops (`capture_pane -t`, `set_session_environment`, `resize_window`, `window_exists`); parked windows fixed for dash (`read -r _park`)
- Dashboard renders the capture through pyte (LNM) each tick, keeps the viewer window resized to the displayed pane, and `attach`/`a` prefer it

## Testing

Unit + real-tmux integration tests (2990 passed, isolated tmux server); verified end-to-end in the live TUI against a real run (`opencode serve` + vllm).

## Summary by CodeRabbit

* **New Features**
  * Added an optional attached UI viewer in a detached terminal window.
  * Added configuration support for enabling the viewer, disabled by default.
  * Attach now prefers the viewer when available.
  * Viewer windows can be reused across retries and resumed sessions.

* **Bug Fixes**
  * Improved viewer rendering, terminal alignment, resizing, pane targeting, and lifecycle cleanup.
  * Added graceful fallback to headless operation when viewer startup is unavailable or fails.
  * Secured viewer connections with per-session authentication.

* **Documentation**
  * Updated configuration examples, key-binding guidance, and the unreleased changelog.
  * Removed legacy `bmad-auto` to `bmad-loop` migration compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,30 @@ breaking changes may land in a minor release.
 
 ### Added
 
+- **`adapter.show_attached_ui` viewer capability.** When
+  `policy.toml` sets `show_attached_ui = true` in `[adapter]`, the opencode-http
+  adapter creates a detached tmux viewer window alongside the headless `opencode
+  serve` process. The viewer runs `opencode attach <server-url> --dir <cwd>
+  --session <session-id>` so the operator can watch the live session in the
+  dashboard Log tab, attach from the TUI (`a` / `Ctrl-A`), or use `bmad-loop
+  attach`. Key behaviors:
+  - HTTP/SSE remains authoritative; the viewer is a read-only observation path.
+  - The dashboard Log tab renders the viewer pane through the same pyte-based
+    terminal-emulator rendering (no headless server log).
+  - `bmad-loop attach` and the TUI `a` action prefer the viewer window when one
+    exists.
+  - Teardown order: stop viewer → abort server session → kill `opencode serve`.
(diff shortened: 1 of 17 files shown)


pbean:
I do appreciate the contribution but at this time I'm going to have to decline the PR as is. We recently merged PR #279 as our first iteration of logging in the run and in the TUI. We may modify and edit it some more as testing continues and encourage contributions against that. I do also want to recommend bringing discussions in regards to PR's this large into the discord first before you spend any time working on it to prevent duplication of work or perhaps design problems we may run into working in a vacuum. We always want to keep the PR scoped to one thing, the blast radius of your actual PR is quite large and adjust things not actually relevant to it. We also built the opencode adapter as an http adapter solution so It doesn't have the user require tmux in its usage so it would be an unexpected requirement if we add it in now as well. Nor is the log a live viewer, its actually a pipe of the output written and emulated in pyte, it just looks live :) I do look forward to reviewing any of your other contributions in the future.

Issue #310, opened 2026-07-26
psmux: per-window user options are session-scoped - the ctl-window project guard cannot discriminate and the parked-return mechanism is inert


dracic:
## Summary

psmux has exactly **one user-option scope per server**. Every `set-option -w` write lands in that one map and every `-w` read of an `@`-prefixed name returns `''`, regardless of target. Two bmad-loop mechanisms ride per-window user options and are therefore broken on psmux **today**, independently of any window-id work:

- **`PROJECT_OPTION` cross-project prune scoping does not discriminate.** `_ctl_window_candidates` reads `runs.PROJECT_OPTION` as a `list-windows` field to leave another project's ctl window alone (`tui/launch.py:214`, tag written at `:310`). On psmux the stored value is session-scoped and `#{@bmad_project}` expands to the same value for *every* row, so the guard cannot tell one project's window from another's on the shared `bmad-loop-ctl` session.
- **`RETURN_OPTION` — the parked-window return mechanism — is fully inert.** The trailer reads it with `show-options -wqv` (`adapters/psmux_backend.py:107`, base at `adapters/tmux_base.py:231`) and `return_attached_client` does the same at `tui/launch.py:146`. That read always yields `''` on psmux, so neither the `detach` nor the `switch-client` branch ever fires: an attached client is never returned to its origin.

This was split out of #291 by scope review — #291 is about *id routing*, this is about needing a different *state channel*. Filing separately rather than growing that diff.

## Verified in the psmux source at tag `v3.3.7`

`v3.3.7` is the tag matching the installed build (`05cc5d4`). Read via `git show v3.3.7:<path>`, since a working checkout a few commits past the tag touches some of these files.

- `src/server/options.rs:4` — `is_window_option` is a hard `matches!` allowlist of exactly **14 builtin names** (`automatic-rename`, `monitor-activity`, `remain-on-exit`, the `window-status-*` family, `main-pane-*`, `window-size`). No `@`-prefixed name can ever be in it.
- `src/server/options.rs:155` — `get_window_option_value_for` returns `String::new()` **immediately** when `!is_window_option(name)`, before ever consulting the user-options map. This is the whole read path for `show-options -wqv`, so a `@`-prefixed `-w` read is empty in every case: targeted or untargeted, in-pane or out.
- `src/types.rs` — the `Window` struct has **no options storage of any kind** (`root`, `active_path`, `name`, `id`, flags, layout, panes — no options map).
[...]

Issue #302, opened 2026-07-25
Windows: verify env-fault classification never fires — cmd reports missing/non-executable tools as 9009/1, not 126/127


dracic:
## Summary

`ENV_FAULT_RCS = frozenset({126, 127})` (`src/bmad_loop/verify.py:1376`) encodes the **sh** launcher convention: 126 = found but not executable, 127 = not found. `run_verify_commands` runs commands with `shell=True`, so on Windows the shell is `cmd`, which never uses those codes — a missing executable is 9009 ("is not recognized as an internal or external command") and a non-executable one typically surfaces as 1.

The consequence is that on Windows the env-fault arm at `verify.py:1410` is unreachable in practice. A broken verify environment is classified an ordinary fixable failure, so the run dispatches repair sessions and burns dev attempts against a tree no session can fix — precisely the failure #126 was filed for and #130 fixed on POSIX.

The comment at `verify.py:1374-1375` states the gap ("Windows cmd signals these as 9009/1 instead, so it keeps the charged behavior") but nothing tracks closing it.

## Evidence

The three engine tests that pin this behavior are all win32-skipped for exactly this reason:

- `test_review_verify_env_fault_escalates_instead_of_fix_session`
- `test_skip_review_env_fault_escalates_not_defers`
- `test_fix_phase_env_fault_escalates_instead_of_looping`

each carrying `@pytest.mark.skipif(sys.platform == "win32", reason="sh env-fault exit codes (cmd reports 9009)")`. So the Windows behavior is neither correct nor exercised.

## Why this is not a trivial patch

Adding 9009 is safe and unambiguous. Adding **1** is not — under `cmd`, rc 1 is also the ordinary "the tests failed" code, which is exactly the fixable failure that *should* route to a repair session. Classifying it an env fault would pause runs on every genuine test failure on Windows.

So the fix needs a decision on how far to go, roughly:

1. **9009 only** — narrow, no false positives, catches the common "tool not on PATH" case, leaves non-executable files charged.
2. **9009 + a stderr probe** — treat rc 1 as an env fault only when the output tail matches cmd's not-recognized/access-denied phrasing. Catches more, but pattern-matches localized shell text.
3. **Bypass the shell for env detection** — resolve the command's leading token before running it and classify unresolvable ones without relying on the shell's exit code at all.

[...]

PR #312, opened 2026-07-26
fix(adapters): session-qualify psmux TUI-side window ids

### LINKED ISSUE #291: psmux: TUI-side window ids are still bare — parked mint, list_windows fields, and prune kill targets replay @N cross-server
## Summary

#254 (fixed by #290) session-qualified the engine-side window-id seam: `PsmuxMultiplexer.new_window` and `list_window_ids` now emit `session:@N`. The TUI-side surfaces were deliberately excluded from that diff (per the issue's implementation notes) and still hand around bare `@N` ids on psmux. Both #290 review passes independently confirmed the remaining surfaces, so this files the promised follow-up audit.

## Affected surfaces

- **`new_parked_window`** (inherited from the base backend) still returns a bare `#{window_id}`. Consumers replay it as a `-t` target: `set_window_option(win_id, PROJECT_OPTION, ...)` and window selection in `tui/launch.py`, plus attach targeting in the TUI app.
- **`list_windows` field rows**: `window_id` columns stay bare, and `prune_ctl_windows` passes those ids straight to `kill_window` — a destructive verb, against the shared cross-project ctl session.
- **`select_ctl_window_id` / ctl window selection** takes bare `@N` values from callers.

## Why this is the same bug class as #254

psmux runs one server per session; a bare `-t @N` routes by the caller's `$TMUX`, else the most-recent-session fallback. The TUI process frequently runs *outside* any psmux pane (user launches from a plain shell), so its bare-id verbs land on whatever server the fallback picks. The #254 forensics also showed warm-recycled panes carrying stale env, making the fallback nondeterministic across sessions — in a controlled two-session test a bare `kill-window -t '@2'` issued in session A killed session B's `@2` with rc 0. The prune path uses exactly this verb shape.

## Notes for the fix

- Mirror the #290 approach: qualify at the psmux backend's minting/listing seams, not per caller. `new_parked_window` should return the same `session:@N` form as `new_window`; `list_windows` should qualify `window_id` fields identically.
- Symmetry hazards to watch: `current_window_id()` comparisons in the ctl-candidate scan (`win_id == current`) must compare like with like, and any caller that composes `mux.target(CTL_SESSION, window)` from an already-qualified id would double-qualify.
- The `session:@N` target grammar is already verified on stock psmux 3.3.7 for `kill-window`, `send-keys`, `pipe-pane`, `list-panes`, and `capture-pane` (see #290's verification notes); only the TUI-side consumers and their id round-trips need auditing here.

Refs: #254, #290, #221, psmux/psmux#483

dracic:
Closes #291

## What

#254/#290 covered the engine seam but left the launcher's surfaces bare — and the TUI usually runs *outside* any psmux pane, where a bare `@N` replayed as a `-t` target resolves through psmux's most-recent-session fallback. The ctl-window prune could `kill-window` another server's identically-numbered window with rc 0, and the detached-resolve select/attach could land on the wrong server.

- `new_parked_window`, the `window_id` columns of `list_windows`, and `current_window_id` now emit `session:@N` (via the existing `_qualified_window_id`, sharing its degrade rules — including a new guard so an empty session name degrades on the compose side exactly like the parse side).
- `current_window_id` resolves session and id from **one** `#{session_name}:#{window_id}` expansion — a split probe can resolve the id while the session fails, which would break the prune's self-exclusion and kill the window it runs in.
- `select_window` resolves a qualified id to `session:<index>` just-in-time: psmux validates a scoped target's window part CLI-side against index/name only (never `window_id`), while the server resolves ids fine (`FocusWindowById`). On a resolve miss/failure it still sends the original target (best-effort contract) but now warns on stderr — psmux's `can't find window` exit 1 lands in a discarded pipe, so the miss was previously untraceable.
- tmux untouched: its ids are server-global (divergence pinned by test).

Every psmux behavior relied on here was read from the psmux source at tag `v3.3.7`, the tag matching the installed build.

## Follow-ups (split out, not fixed here)

- #310 — per-window user options are session-scoped on psmux, so the `PROJECT_OPTION` prune guard cannot discriminate between projects and the `RETURN_OPTION` parked-return mechanism is inert. Split out of #291 by scope review: this PR fixes *id routing* only; that needs a different *state channel*. The window-option verbs are deliberately untouched here — psmux drops the window slot of a `-w` target either way, so the qualified ids this PR introduces neither help nor harm those surfaces.
- A garbled `current_window_id` probe degrades to `None`, which disables the prune's self-exclusion; pre-existing on both backends and out of scope here — will be filed separately.

## Verification

- `pytest tests/test_psmux_backend.py tests/test_tui_launch.py tests/test_multiplexer.py tests/test_portability_guard.py -q` → 134 passed (16 new tests, incl. ablation-checked degrade/passthrough gates)
- `trunk check` clean, `pyright@1.1.411` clean on the touched module

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved psmux window targeting by consistently scoping TUI window IDs to their sessions.
  * Prevented selection/cleanup from targeting identically numbered windows belonging to different sessions.
  * Added safer handling when session-qualified IDs can’t be represented or resolved.
* **Documentation**
  * Clarified adapter and TUI expectations for backend-specific window ID formatting and matching behavior.
* **Tests**
  * Added coverage for session-qualified ID behavior across parked-window creation, listing, probing, and selection, plus tmux compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,6 +125,16 @@ story <id>`, the same annotation a sweep bundle writes. Both sprint and stories 
 
 ### Fixed
 
+- **Session-qualify the psmux TUI-side window ids (#291).** #254 covered the engine seam but left
+  the launcher's surfaces bare, and that process usually runs _outside_ any pane — where a bare
+  `@N` resolves through psmux's most-recent-session fallback, not the session that minted it. The
+  ctl-window prune replayed such ids as `kill-window` targets and could close another server's
+  identically-numbered window with rc 0. `new_parked_window`, the `window_id` columns of
+  `list_windows`, and `current_window_id` now emit `session:@N`; the prune compares the last two,
+  so they qualify together or a prune kills the window it runs in. `select_window` resolves the id
+  to an index first — psmux validates a scoped target's window part CLI-side against window
+  index/name only. tmux is untouched (its ids are server-global).
+
 - **Verify environment faults are classified per shell, so Windows stops burning dev attempts on a
   broken tree (#302).** `ENV_FAULT_RCS = {126, 127}` is `sh`'s launcher convention, but verify
   commands run through the host shell — and `cmd` has no equivalent: a missing tool exits `1`,
--- a/docs/adapter-authoring-guide.md
+++ b/docs/adapter-authoring-guide.md
@@ -108,11 +108,14 @@ module-level `parse_target()` — or the backend's own **native id** (whatever y
 `target()`. (Two values are composed by the backend itself rather than by
 `target()`, precisely so the seam grammar never has to carry a pane or window
 id: the parked-window return target — `current_return_target`, above — and the
-native window id, which psmux qualifies to `session:@N` across the
-`new_window`/`list_window_ids` pair because its ids are per-server (falling back
-to the bare id where that grammar would not survive — the backend owns those
-conditions, and applies them to both halves of the pair so the symmetry rule
-still holds). Both are replayed opaquely; neither is parsed by core.) tmux consumes the token natively (it coincides with tmux exact-match
+native window id, which psmux qualifies to `session:@N` because its ids are
+per-server (falling back to the bare id where that grammar would not survive —
+the backend owns those conditions, and applies them uniformly, so the
+`new_window`/`list_window_ids` symmetry rule holds under the fallback too).
+Both are replayed opaquely; neither is parsed by core. psmux applies the same
+qualification to `new_parked_window`, the `window_id` columns of `list_windows`
+and `current_window_id`; the latter two must agree, since the ctl-window prune
+compares them to skip its own window.) tmux consumes the token natively (it coincides with tmux exact-match
 syntax), so `BaseTmuxBackend` passes it straight through. A native-id backend
 calls `parse_target()` first — `None` means "already a native id, use as-is",
 otherwise resolve `(session, window)` yourself; the herdr adapter's
--- a/src/bmad_loop/adapters/multiplexer.py
+++ b/src/bmad_loop/adapters/multiplexer.py
@@ -164,11 +164,10 @@ class TerminalMultiplexer(ABC):
         server per session), so a bare ``@N`` replayed as a ``-t`` target
         routes by the *caller's* server instead of the owning one.
 
-        :meth:`new_parked_window` is deliberately *outside* the rule: nothing
-        membership-tests a parked id — it is only replayed as a ``-t`` target,
-        by the TUI — so a backend MAY mint it in a form this list never
-        carries. psmux keeps it bare, because the qualified grammar is not
-        accepted by the verbs the parked path uses (#291).
+        :meth:`new_parked_window` is *outside* the rule — nothing
+        membership-tests a parked id, it is only replayed as a ``-t`` target by
+        the TUI — so a backend MAY mint it in a form this list never carries
+        (psmux happens to qualify it too, #291).
 
         Raises :class:`MultiplexerError` if the transport itself fails (timeout /
         missing binary): an empty list means "no windows" and must not be
@@ -180,7 +179,10 @@ class TerminalMultiplexer(ABC):
         """One tuple per window in ``session``, each holding the requested
         backend fields in order. Best-effort: returns ``[]`` on a transport
         failure (unlike :meth:`list_window_ids`, this is metadata, not a liveness
-        probe, so a sentinel is safe)."""
+        probe, so a sentinel is safe).
+
+        A ``window_id`` column carries the same id form :meth:`current_window_id`
+        returns; core compares the two directly."""
 
     @abstractmethod
     def window_alive(self, session: str, window_id: str) -> bool:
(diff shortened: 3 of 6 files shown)


pbean:
Reviewed and validated. **Approach is right, scope is clean, and the tmux path is provably untouched.** Rebased onto `main` (CHANGELOG only) and pushed one review commit; details below.

## Independently re-verified against the psmux source at `v3.3.7`

I didn't take the source claims on faith — the `select_window` override is the one piece of new machinery here, so its premise had to hold:

- **`cli_window_exists` (`src/main.rs:93-118`)** compares the target's window part only against `#{window_index}` and `#{window_name}`. Never `#{window_id}`. The premise holds.
- **It has exactly one call site** — the `select-window` handler (`src/main.rs:1752`). That is the load-bearing fact, and it settles the design two ways at once: `select_window` really is the *only* verb needing the index translation, and **`kill-window` has no CLI-side validation at all**, so the prune — the destructive path this PR exists to fix — takes the qualified id straight to the server. No prune regression.
- **`parse_target` strips exactly one `=`** (`src/cli.rs:527`), so the `==session` failure the `=`-strip comment describes is real, not theoretical.
- **Session→server routing is an exact-string port-file lookup** (`extract_session_from_target`, `src/cli.rs:634`) — so recomposing `session:<index>` *without* the `=` introduces no prefix-match hazard. Worth stating explicitly since dropping an exact-match marker normally would.

**I also treated the #310 deferral as a regression risk rather than a note**, because `current_window_id` now feeds qualified ids to the option verbs via `return_attached_client`:

- `set-option` server-side (`src/server/connection.rs:2088`) filters the `-t` value out of its positional args entirely — target wholly ignored.
- `show-options` *does* read `-t`, but only through `parse_target(...).window`, which yields `Some(2)` for **both** `@2` and `session:@2` (`window_is_id` is discarded, `src/cli.rs:602`).

So both option verbs behave identically before and after. Qualifying is strictly better — right server, same pre-existing scope bug. The split to #310 is correct, and its "do this after #291" ordering is right.

## Linux/macOS: zero impact

`multiplexer.py`'s diff is **docstring-only**, `tui/launch.py`'s three hunks are docstrings, and `TmuxMultiplexer`/`PsmuxMultiplexer` are siblings under `BaseTmuxBackend` with every override on the psmux leaf.

`test_tmux_backend_keeps_bare_tui_ids` is a *placement* assertion, so I ran the **inverse** ablation it needs — re-implementing the qualification in the shared base — and confirmed it goes red. It isn't passing merely because nothing qualifies anywhere.

## Scope

Exactly #291's three named surfaces plus the `current_window_id` symmetry hazard #291's own implementation notes flagged. The double-qualification hazard those notes warned about does not materialize: `select_ctl_window`, `attach_plan`, `kill_ctl_window`, and `action_attach` all compose `ctl_target()` from window **names**, never from an id.

The one reach past the TUI surfaces is the `not session` guard in `_qualified_window_id`, which also covers the engine seam. That's justified beyond the symmetry argument given for it: composing `":@N"` yields a target `parse_target` resolves to a session literally named `default`, so the guard prevents an actively wrong target, not just an asymmetric one.

## Verification

- Full suite **3111 passed, 30 skipped** post-rebase. The 30 is `main`'s own baseline (`main` alone: 3091/30) — this PR adds 20 passing tests and **zero** new skips.
- `pyright@1.1.411` clean; `trunk check` clean.
- **Ablation: 9/9 held.** I ablated each new gate individually — empty-session degrade, `list_windows` column filter, `select_window` resolve gate, `=` strip, send-on-resolve-miss, the one-probe `current_window_id`, the malformed-id check, `_window_index` session scoping — and confirmed each guarding test turns red, plus the inverse placement ablation above. The negative assertions here are trustworthy.

## One finding (addressed in `d323fd6`, comment-only)

The resolve-miss warning is **inert at its only caller, and the comment claimed otherwise.**

`select_window` prints the miss to `sys.stderr`, justified as "or that is untraceable". But the only production path reaching the qualified-id branch is `app.py:494` `_launch_resolve` → `select_ctl_window_id` — the sole caller of that function (`select_ctl_window` passes a *name* token and returns before the warning). That runs inside the Textual app, and `tui/app.py:1114-1117` documents that Textual captures `sys.stderr` for the app's whole run — `run_tui` pre-trips the forced-backend warning before `App.run()` for exactly this reason. So the miss is untraceable either way, and the cited `pipe_pane` precedent (`tmux_base.py:322`) is an **engine**-path warning where stderr is a real terminal — not comparable.

Not a correctness bug: the degrade is safe (futile target sent, base swallows the rc 1, the window simply isn't focused). The defect was a diagnostic that cannot fire carrying a rationale that would mislead the next reader.

Fixed by correcting the comment rather than the plumbing — the print is worth keeping for the CLI-side callers `tui/launch.py` stays textual-free to allow, and making the miss TUI-visible would mean giving a best-effort `-> None` focus seam a return value. Wrong altitude for a rare cosmetic miss.

## Rebase note

`CHANGELOG.md` was the only conflict (#302's entry from #308 landed in the same `### Fixed` slot); resolved keep-both, #291 above #302 per the newest-first order this PR already used. `+363/-28` unchanged across the rebase.

Minor: the description says 16 new tests; it's 20.

PR #314, opened 2026-07-26
fix(tests): remove hardcoded GNU coreutils dependencies on Windows (#293)

### LINKED ISSUE #293: Windows: two tests hard-require GNU sleep/true on PATH — green on CI runners, red on a stock local shell
## Summary

Two tests depend on GNU coreutils being resolvable on Windows PATH:

- `tests/test_runs.py::test_stop_run_signals_live_process` — spawns its victim with `subprocess.Popen(["sleep", "30"])`. No `sleep.exe` on PATH → `FileNotFoundError` before the test's subject (SIGTERM delivery to a live engine pid) even starts.
- `tests/test_verify.py::test_verify_review_happy_and_commands` — uses `VerifyPolicy(commands=("true",))` (and `("true", "false")`) as the happy/fail commands. cmd has no `true`/`false` builtins, so the happy path exits 9009 and the assertion fails.

GitHub's `windows-latest` runners ship `C:\Program Files\Git\usr\bin` on PATH, so CI is green. A stock local shell (only `Git\cmd` on PATH — the default Git for Windows install) fails both. Verified during the #254 full-suite runs: both fail on a clean PATH and both pass once `Git\usr\bin` is prepended.

## Suggested fix

Remove the coreutils dependency instead of documenting it:

- The sleeper: `subprocess.Popen([sys.executable, "-c", "import time; time.sleep(30)"])` — same lifetime semantics, zero PATH dependency, and the pid is a real process the SIGTERM path can kill on any OS.
- The verify commands: `tests/conftest.py` already centralizes cross-shell verify verbs (`_OK = "exit 0"` honored by both `cmd /c` and `sh -c`); use that pattern for the happy command and an `exit 1` equivalent for the failing one.

Refs: #254

dracic:
## Summary

Two tests depended on GNU coreutils being resolvable on Windows PATH:
- `tests/test_runs.py::test_stop_run_signals_live_process` spawned `subprocess.Popen(["sleep", "30"])`.
- `tests/test_cleanup.py::_dead_pid()` spawned `subprocess.Popen(["true"])`.

On stock Windows environments without `Git\usr\bin` on PATH, these failed with `FileNotFoundError` or exit code 9009 (`'true' is not recognized as an internal or external command`).

## Changes

- Updated `test_stop_run_signals_live_process` to use `sys.executable` with `import time; time.sleep(30)` and added process cleanup in a `finally` block.
- Updated `_dead_pid` in `tests/test_cleanup.py` to spawn `sys.executable` with `""`.

The `tests/test_verify.py` half of the original diff (`true`/`false` verify commands) landed on main via #308, so this branch was rebased onto it and now carries only the two `Popen` call sites.

Fixes #293

## Summary by CodeRabbit

* **Tests**
  * Improved process-related test reliability by using the active Python interpreter instead of platform-specific external commands.
  * Added cleanup safeguards to ensure test processes are terminated and reaped properly.

### DIFF
--- a/tests/test_cleanup.py
+++ b/tests/test_cleanup.py
@@ -3,6 +3,7 @@ artifact trim, and the `clean` CLI command."""
 
 import argparse
 import subprocess
+import sys
 
 from conftest import install_bmad_config, machine_json
 
@@ -22,7 +23,7 @@ def _state_run(project, run_id, **kw):
 
 
 def _dead_pid() -> int:
-    p = subprocess.Popen(["true"])
+    p = subprocess.Popen([sys.executable, "-c", ""])
     p.wait()
     return p.pid
 
--- a/tests/test_cleanup.py
+++ b/tests/test_cleanup.py
@@ -2,8 +2,6 @@
 artifact trim, and the `clean` CLI command."""
 
 import argparse
-import subprocess
-import sys
 
 from conftest import install_bmad_config, machine_json
 
@@ -20,12 +18,6 @@ def _state_run(project, run_id, **kw):
         RunState(run_id=run_id, project=str(project), started_at="2026-06-11T10:00:00", **kw),
     )
     return run_dir
-
-
-def _dead_pid() -> int:
-    p = subprocess.Popen([sys.executable, "-c", ""])
-    p.wait()
-    return p.pid
 
 
 # --------------------------------------------------------------- predicates
--- a/tests/test_runs.py
+++ b/tests/test_runs.py
@@ -405,12 +405,17 @@ def test_stop_run_dead_pid_falls_back(tmp_path, monkeypatch):
 def test_stop_run_signals_live_process(tmp_path, monkeypatch):
     monkeypatch.setattr(runs, "kill_session", lambda _rid: None)
     run_dir = _make_state_run(tmp_path, "r1")
-    proc = subprocess.Popen(["sleep", "30"])
-    (run_dir / "engine.pid").write_text(str(proc.pid))
-    assert runs.stop_run(run_dir) is True
-    # the process received SIGTERM and is gone
-    assert proc.poll() is not None or proc.wait(timeout=5) is not None
-    assert load_state(run_dir).stopped is True
+    proc = subprocess.Popen([sys.executable, "-c", "import time; time.sleep(30)"])
+    try:
+        (run_dir / "engine.pid").write_text(str(proc.pid))
+        assert runs.stop_run(run_dir) is True
+        # the process received SIGTERM and is gone
+        assert proc.poll() is not None or proc.wait(timeout=5) is not None
+        assert load_state(run_dir).stopped is True
+    finally:
+        if proc.poll() is None:
+            proc.kill()
+            proc.wait()
 
 
 def test_stop_run_respects_engine_written_stopped(tmp_path, monkeypatch):
--- a/tests/test_runs.py
+++ b/tests/test_runs.py
@@ -415,7 +415,7 @@ def test_stop_run_signals_live_process(tmp_path, monkeypatch):
     finally:
         if proc.poll() is None:
             proc.kill()
-            proc.wait()
+            proc.wait(timeout=10)
 
 
 def test_stop_run_respects_engine_written_stopped(tmp_path, monkeypatch):


pbean:
Reviewed — approach is right, and I pushed one follow-up commit (6454a6a).

**The rebase landed well.** You beat me to it; your `bb7bfef` resolved the `tests/test_verify.py` conflict correctly — the file is byte-identical to main and #308's `_FAIL` constant survived. Worth flagging for anyone hitting this shape later: that conflict was the one real hazard here. This branch forked at `5432c81`, before `ed77c6b` ("classify verify env faults per shell", #302) landed the same fix with a named `_FAIL` in `conftest.py` plus the comment explaining why POSIX `false` isn't cmd's failure verb. Resolving it the reflex way — keeping the branch's inline `"exit 1"` — would have quietly reverted that. It didn't, so no action needed; just noting why that hunk is gone from your commit now.

**What I changed:** `tests/test_cleanup.py::_dead_pid` is dead code — it has never had a call site. `62818ed` created that file by copying the helper out of `test_runs.py`, `Popen(["true"])` body and all, but never wrote a call; the occurrence count has been 1 (the bare `def`) from that commit through HEAD, with no cross-file import.

That means it was never part of #293's failure set — an uncalled helper can't fail to resolve `true` on any PATH. It's also why #31 fixed only the live copy when it added the native-Windows lane, and why three weeks of that lane never flagged this one. Porting it would re-apply #31's fix to code that never runs and re-sync a duplicate of `test_runs.py:42`, recreating exactly the duplication that let it drift. So I deleted it and its now-unused `import subprocess` instead — the coreutils dependency goes away either way, and the PR body's claim that this spot failed with `FileNotFoundError`/9009 doesn't hold for it.

Also bounded the `proc.wait()` in your new `finally` to `timeout=10`, matching every other wait in that file and in `test_process_host.py`. The `finally` itself is a good addition — before it, a failed assertion leaked the child for 30s.

The `test_runs.py` sleeper fix is the real one and is exactly right: `sys.executable` is the pattern already established at `test_runs.py:42`, and it's more faithful than `sleep` since a real engine pid *is* a Python process.

**One thing left, and CI can't do it.** All checks green is not evidence for this issue — GitHub's `windows-latest` ships `C:\Program Files\Git\usr\bin` on PATH, which is precisely why #293 was invisible to the Windows lane to begin with. Could you post a run of

```
uv run pytest tests/test_runs.py::test_stop_run_signals_live_process -q
```

from a shell **without** `Git\usr\bin` on PATH, including the duration? Two things I'd like confirmed: that it passes, and how long it takes. `WindowsProcessHost.terminate` is `taskkill /PID` without `/F` and swallows failure (`check=False`); a console child with no message loop may well refuse it, in which case `stop_run` burns the full `_STOP_WAIT_S = 10.0` grace window before reaching `taskkill /F /T`. Should still pass — `sleep.exe` had the same shape, so it's not a regression — but worth knowing before we call the Windows story closed.

Green here: full suite `3111 passed, 30 skipped`, `trunk check` clean.

pbean:
**Verified on native Windows, stock PATH.** @pbean ran this on a Windows VM (PowerShell, CPython 3.13 via uv, no `Git\usr\bin` on PATH) as a before/after ablation — the evidence CI structurally can't produce, since `windows-latest` carries `Git\usr\bin`.

**Before — `main`, reproduces #293:**

```
> uv run pytest tests/test_runs.py::test_stop_run_signals_live_process -q
tests\test_runs.py:408:
    proc = subprocess.Popen(["sleep", "30"])
E   FileNotFoundError: [WinError 2] The system cannot find the file specified
1 failed in 0.55s
```

`CreateProcess` failing to resolve `sleep` is also a stronger proof of the stock-PATH precondition than a `where.exe` check.

**After — this branch (`6454a6a`):**

```
> uv run pytest tests/test_runs.py::test_stop_run_signals_live_process -q --durations=0
1 passed in 11.19s
   10.26s call     tests/test_runs.py::test_stop_run_signals_live_process
```

**Full #293 surface, same shell:** `302 passed, 1 skipped in 354.72s`
(`tests/test_runs.py tests/test_cleanup.py tests/test_verify.py`).

So #293 is genuinely closed. 👍

---

**One confirmed side-observation, for a follow-up — not this PR.**

That `10.26s` is not noise; it's `_STOP_WAIT_S = 10.0` being burned in full. It means `host.is_alive(pid)` stayed true for the entire grace window after `terminate()`, which empirically confirms that `WindowsProcessHost.terminate` — `taskkill /PID` without `/F`, whose failure is swallowed by `check=False` — **cannot close a console child with no message loop**. Every stop on native Windows therefore waits 10s and then force-kills via `taskkill /F /T`.

Not a regression (`sleep.exe` had the identical shape, so this cost predates the PR) and functionally correct, since `stop_run`'s fallback still marks `stopped` and `kill_session` still backstops the agent window. But it does mean the path `stop_run`'s own docstring says it *prefers* — "the engine's own SIGTERM handler so the engine stays the single writer of `stopped`" — never actually runs on native Windows; every stop takes the force-kill branch. Worth an issue against the native-Windows tier rather than widening this PR.

pbean:
Filed the follow-up as #319 (`bug`, `area:engine`, `needs-design`) — native-Windows `stop_run` always force-killing after the full 10s wait, with the 10.26s measurement as evidence. Out of scope here; this PR stays the two-file test fix.

Issue #271, opened 2026-07-23
[BUG] Review session timeout always RETRY until max_review_cycles; no salvage/converge exit


Haven2026:
### Description

`decide_review_session` treats **any** non-`completed` review session status the same way: if `review_cycle < max_review_cycles`, return **RETRY**; else DEFER/PAUSE.

There is **no** branch for:

- `timeout` / `timeout_fired` vs crash vs user-stop,
- “prior review cycle already left the story `status: done` + verify-green — converge and refile residual follow-up instead of burning another full review,”
- “do not spend `max_followup_reviews` after a pure timeout with no new review artifact.”

Result under unattended multi-story runs: a **90m session timeout** on review-1 (or a heavy codex second-opinion) becomes **review-retry → review-2 → review-3…** until the outer cycle budget dies, even when the product tree from dev is already good enough to commit/salvage.

### Steps to reproduce

1. Policy: `review.enabled = true`, `review.trigger = "recommended"` (or always), `limits.session_timeout_min = 90`, `limits.max_review_cycles = 3`, `limits.max_followup_reviews = 1`.
2. Adapter review on a slower/heavier model (field: codex second opinion) so a single review pass can hit session timeout.
3. Dev pass finishes with a green-enough tree and often already `followup_review_recommended` from inline review.
4. Watch journal: `review session` non-completed → `review-retry` → next cycle → repeat until cycles exhausted or human stops the run.

### Expected behavior

At least one of (policy-selectable is fine):

1. **Timeout exit class**: after N timeouts (default 1) on review, if the last known good state is `spec status: done` (and optional verify commands green), **converge**: commit / refile lingering follow-up to deferred-work, **do not** open another full review pass solely because the session timed out.
2. **No follow-up grant after timeout**: a timeout that produced **no** successful review result_json should not consume or re-arm `followup_review_recommended` damping in a way that schedules another independent pass with empty product delta.
3. Journal a distinct event e.g. `review-timeout-salvage` vs bare `review-retry` so external supervisors can act without guessing.

### Actual behavior

From installed 0.8.1 `escalation.decide_review_session`:

```python
if result.status != "completed":
    reason = f"review session {result.status}"
    if budget_left:
        return Decision(Action.RETRY, reason)
    return Decision(_exhausted_action(task), _exhaust_reason(task, reason))
```

[...]

PR #339, opened 2026-07-27
feat(review): escalate deliberate sprint-status regressions at review-verify

### LINKED ISSUE #334: Escalate deliberate sprint-status regressions at review-verify instead of burning review cycles
When a review session deliberately reverts a story's sprint-status because the work is not actually done, the engine has no way to hear it. It re-runs the review until `max_review_cycles` is exhausted, then defers the story and rolls the work back. The disagreement is structural — no number of retries can resolve it — so every cycle is wasted, and the outcome is the worst available one.

## The livelock chain (validated at `3c68153`)

1. **The orchestrator flips sprint-status to `done` at dev time**, not the dev session — `engine.py:2072-2080` (`_dev_review_enabled` is hardwired `False` for bmad-dev-auto → `target="done"`). `engine.py:2080` is the *only* `sprint_advance` call site in the codebase. The dev skill sets spec frontmatter `status: done` unconditionally (`step-04-review.md:93`) with no acceptance-criteria gate anywhere.
2. The review session reads the board, sees work that cannot be done, and reverts the status to `in-progress` — the correct call.
3. `verify_review` literally checks `status != "done"` for both spec and sprint → `VerifyOutcome.retry(fixable=False)` — `verify.py:1565-1581`.
4. `engine.py:1479-1492`: no escalate, no fix, bare `continue`.
5. The review loop never re-advances sprint-status, so the condition is unrecoverable. Bounded only by `max_review_cycles` (reporter's run: 3 wasted passes).
6. `engine.py:1551` → defer + rollback. The review was right and the run treats it as a failure.

`review-verify-failed` is a journal event kind (`engine.py:1473-1478`); nothing routes on it.

## Why it is deterministically detectable

Three facts make "sprint-status regressed below `done`" a reliable signal of a *deliberate session write*, distinguishable from "never got there":

- `engine.py:2080` is the sole `sprint_advance` site, and it runs at dev time.
- `verify_dev` already asserted the story reached `done` (`verify.py:1194-1199`).
- `sprintstatus.advance()` never regresses.

So at review-verify, `STATUS_ORDER.index(sprint) < STATUS_ORDER.index("done")` plus the durable knowledge that it got to `done` ⇒ a session wrote it backwards on purpose. No snapshot needed; the invariant is structural and survives crash-resume for free.

There is no structured review→engine verdict channel today: the review session is a bmad-dev-auto re-invocation with no `result.json`, and the spec template's `warnings: []` frontmatter (`spec-template.md:8`) is declared but never read.

## Fix outline

- Detection lands in **`verify_review`** (`verify.py:1565-1581`), not the engine loop: new kw-only `sprint_reached_done: bool`, passed by `Engine._verify_review` (`engine.py:2378-2379`) as `not self._dev_review_enabled()`. All four `_verify_review` call sites inherit routing through the existing not-retryable→escalate path; only the rescue site (`engine.py:1528-1531`, bare `.ok`) needs an edit — capture the outcome, escalate on contradiction instead of misleading-deferring.
- Keys on **sprint status only** (orchestrator-owned truth), never spec status (skill-owned, legitimately cycles). Guard: the fresh status must be in `STATUS_ORDER` with index < `index("done")`; `None`/unknown token → conservative retry. Spec `status: blocked` remains the sanctioned reviewer channel and does not overlap (it pauses before verify).
- New knob `review.on_status_contradiction = "escalate"` (default) `| "retry"` (legacy opt-out), following the `review.on_timeout` template from #271 verbatim: `policy.py` constant ~`:31`, `ReviewPolicy` field ~`:206`, parse/validation ~`:781-791`, `POLICY_TEMPLATE` ~`:1050`, `core.toml` `[review]` block after `:54`, `OPTIONS_ENUM` in `tests/test_settings_schema.py:75-85`. Default `escalate` is justified: the released behaviour is a defect — it burns cycles and then *rolls the work back*.
- `VerifyOutcome.contradiction: bool = False` field + `escalate()` kwarg (`env_fault` precedent, `model.py:474-499`). Journal keeps the `review-verify-failed` kind and adds a `contradiction=` field at both emit sites (`:1473-1478`, `:1667-1672`); the TUI colorizer is substring-based, no change.
- Escalation reason text names both sides of the contradiction and the resolution options; it travels via `paused_reason` → `resolve.build_context` (`resolve.py:110-121`).
- **Prompt hardening is explicitly out.** Prohibiting the revert in the review prompt would make a *correct* reviewer comply, and the story would silently commit without sign-off — the contradiction becomes invisible, which is strictly worse.
- Re-drives rearm budgets; an unresolved contradiction re-escalates with the same reason (each pauses, so no ping-pong). No `escalation.py` / `resolve.py` / stories / sweep / TUI changes.

## Known limitation (documented, out of scope)

A review that regresses **both** spec and sprint status never reaches `_verify_review` (gated at `engine.py:1462`) and still lands in the non-terminal-status exhaustion-defer path.

## Doc fix to fold in here

`gates.on_escalation` is a **dead reserved field** — declared at `policy.py:64`, parsed at `:670`, with zero readers; pause is hardcoded. The umbrella reporter had set `on_escalation = pause` and it was inert. Document it as reserved (or read it) as part of this change.

## Tests

`test_policy.py` (default / parse / invalid / template); `test_verify.py` four cases (escalates — this one is the ablation detector; retry-mode legacy; no-launch-flag; missing/unknown token); `test_engine.py` (sign-off regression escalates within 2 scripted sessions — with the gate ablated the engine requests a nonexistent 3rd session and fails hard; retry-mode 3-cycle legacy defer; non-regression verify failure still retries; rescue-site contradiction escalates rather than defers, with `max_followup_reviews=5` so damping never trips; keep the `:5140` fixable-path test green). Run both ablations before trusting the suite (repo rule).

Part of the 0.9.0 field-report umbrella (linked parent).

pbean:
Closes #334

## The livelock

The orchestrator — not the dev session — advances `sprint-status` to `done` at dev time (`engine.py:2080`, its sole `sprint_advance` site; `verify_dev` then asserts the write landed, and `advance` never regresses). A review session that judges the story unfinished and writes the board back to an earlier stage is doing the right thing, but nothing in the review loop re-advances it. So `verify_review`'s `sprint != "done"` check returned a non-fixable retry, the loop `continue`d, and every remaining cycle re-read the identical failure — until `max_review_cycles` ran out and the story deferred with its work rolled back.

Reported by Wojtek Diset against 0.9.0: three review passes burned on a structural disagreement, then a rollback onto an `attempt-preserve` ref.

## The fix

`verify_review` gains a kw-only `sprint_reached_done: bool = False`. Only `Engine._verify_review` passes it (`not self._dev_review_enabled()` — exactly the case where `_post_dev_state_sync` targeted `done`); the sweep and stories gates are separate functions and are untouched. When the flag is set, the knob is `escalate`, and the fresh sprint status is in `STATUS_ORDER` at an index below `done`, the gate returns `VerifyOutcome.escalate(..., contradiction=True)` — a reason naming both sides, both ways out, and the opt-out knob. That flows into the *existing* `not outcome.retryable → self._escalate` routing at both call sites, so the run pauses and `paused_reason` carries the text to `bmad-loop resolve`.

One site needed an edit: the budget-exhaustion rescue (`engine.py:1528`) tested `self._verify_review(task).ok` inline. When every cycle recommends its own follow-up the in-loop verify never runs, so the rescue is the *first* read of the board — a contradiction there would have fallen through to the `did not converge` defer that rolls the work back naming neither side. It now captures the outcome and escalates on `contradiction`.

**New knob** `[review] on_status_contradiction = "escalate" (default) | "retry"`, following the #271 `on_timeout` template (constant, `ReviewPolicy` field, parse + validation, `POLICY_TEMPLATE`, `core.toml`, `OPTIONS_ENUM`). Default `escalate` because the released behavior is the defect — it burns cycles and *then* discards the work; `retry` reproduces it verbatim for anyone who wants it back.

## Deliberate scope boundaries

- **Sprint-status only, never spec status.** The spec's frontmatter legitimately cycles (`in-review` / `in-progress`) while a review patches; the board is orchestrator-owned truth. `status: blocked` remains the sanctioned reviewer hand-back channel and pauses before this gate — no overlap.
- **No review-prompt change.** Prohibiting the revert would make a *correct* reviewer comply, and the story would commit without sign-off. The contradiction would become invisible rather than resolved — strictly worse than the livelock.
- **Conservative on every uncertainty.** No launch flag, a missing board entry (`None`), or a token outside `STATUS_ORDER` (hand-edited or future board) all fall back to the legacy retry. A wrong escalation halts an otherwise healthy run.
- No changes to `escalation.py`, `resolve.py`, `sweep.py`, `stories_engine.py`, or the TUI. The journal keeps the `review-verify-failed` kind and only gains a `contradiction=` field, so the TUI's substring colorizer is unaffected.

## Known limitation (documented, out of scope)

A review that regresses **both** the spec status and the sprint board never reaches `_verify_review` — the loop's `status == "done"` gate at `engine.py:1462` short-circuits first — so it still lands in the non-terminal-status exhaustion defer.

## Tests

`test_policy.py` (default / parse / invalid / template), `test_verify.py` ×4 (escalates · `retry`-mode legacy · no launch flag · missing-or-unknown token), `test_engine.py` ×4 (sign-off regression escalates in 2 scripted sessions · `retry`-mode 3-cycle legacy defer + rollback · a non-regression verify failure keeps its fixable-retry routing · rescue-site contradiction with `max_followup_reviews=5` escalates instead of deferring).

Three ablations run before trusting them (repo rule):

| Ablation | Expected to fail | Result |
|---|---|---|
| Delete the escalate branch | the 3 escalate tests | ✅ — and the engine one fails exactly as designed: `ScriptExhausted: no scripted result for session 1-1-a-review-2` (the ablated engine asks for the review cycle the fix prevents) |
| Ignore the knob (always escalate) | the 2 `retry`-mode tests | ✅ |
| Drop the launch-flag + `STATUS_ORDER` guards | the 3 must-not-escalate tests | ✅ |

Full suite `3363 passed, 24 skipped`; `trunk check` clean; `pyright@1.1.411` 0 errors.


pbean:
Addressed the Greptile P2: the rescue-gate escalation now emits `review-verify-failed` with `contradiction=True` like the two in-loop gates (`ad1b628`). The claim checked out — that path reached `_escalate` directly, so a consumer keying on the new flag saw only `story-escalated`, which defeats the point of adding the field. The non-contradiction arm keeps its existing silence: its outcome is described by the `did not converge` defer reason, and emitting there would change unrelated behavior.

Ablated: dropping the new append fails the rescue-gate test's journal assertion.

Issue #343, opened 2026-07-28
_run_git translates only TimeoutExpired, so a spawn-level OSError bypasses every except GitError guard


pbean:
`verify._run_git` is the sole git spawn point, and it translates exactly one failure into `GitError`:

```python
    try:
        return subprocess.run(cmd, capture_output=True, text=True, timeout=_git_timeout_s, env=...)
    except subprocess.TimeoutExpired as exc:
        raise GitError(f"git {cmd[3]} timed out after {_git_timeout_s}s in {repo}") from exc
```

An `OSError` — EMFILE, ENOMEM, ENOENT on the git binary — is not translated. It escapes `subprocess.run` untyped, sails past every `except verify.GitError` guard in the codebase, and reaches the engine's catch-all (`engine.py:441`), which converts it to `state.crashed` + a `crash.txt`.

`OSError` also reaches these guards without any spawn at all: `verify.snapshot_worktree` opens a `tempfile.TemporaryDirectory()` (`verify.py:461`) before its first git child, so an ENOSPC/EMFILE there raises straight out of a function whose contract says `GitError`.

This is already known and defended case by case: `verify.is_ancestor` and `verify.worktree_prune` catch `(GitError, OSError)`, with comments acknowledging the gap. But the defense is per-site and by reviewer memory.

## Scale

```
$ grep -rnE "^\s*except [(]?[a-zA-Z_.]*GitError" src/bmad_loop/ | wc -l
27
$ grep -rnE "^\s*except \([a-zA-Z_.]*GitError, OSError)" src/bmad_loop/ | wc -l
8
```

19 of 27 guards are `OSError`-blind (was 22 of 26 before #341). Each is a latent crash under memory or fd pressure — and pressure is precisely when the recovery paths that own most of these guards are running.

#341 closed `recovery_flow.py` completely; it is now uniformly `OSError`-safe. The remaining 19 sit in `workspace.py` (7), `worktree_flow.py` (4), `runs.py` (2), `engine.py` (2), and one each in `verify.py`, `tui/app.py`, `decisions.py`, `cli.py`.

## The worked example: `recovery_flow.rollback_or_pause` — now fixed in #341

#341 widened three guards on this path, then found four more frames *upstream* of all of them, on the way into the same method. **All four are fixed in #341** and are recorded here only as the evidence for what the repo-wide default costs:

| Site | Guard before | Failure before |
| --- | --- | --- |
[...]

PR #353, opened 2026-07-28
fix(recovery): probe for a free attempt-preserve-dirty ref name

### LINKED ISSUE #349: A re-armed escalation overwrites the previous attempt's attempt-preserve-dirty snapshot
`recovery_flow.py:430` names the dirty-tree rollback snapshot
`refs/attempt-preserve-dirty/{slug}-{baseline[:8]}-{task.attempt}`, on the premise stated in the
comment directly above it:

> `task.attempt` only ever increments (never resets), so it uniquely discriminates each retry's
> recovery ref.

`runs.rearm_escalation` resets `task.attempt` to 0. A resolve session that commits nothing leaves
`HEAD == baseline`, so the post-resolve re-drive's rollback recomputes the **identical** ref name
and force-updates it — destroying the only copy of the pre-resolve attempt's work.

### Steps to reproduce

1. A story escalates with a dirty tree; the rollback parks `…/<slug>-<baseline>-0`.
2. Resolve the escalation, committing nothing (`HEAD` stays at `baseline`).
3. Re-drive; the story escalates dirty again.
4. The rollback computes `…/<slug>-<baseline>-0` a second time and overwrites the ref.

### Expected

Every rollback parks its snapshot to a distinct ref.

### Actual

The second snapshot silently replaces the first. The work is unrecoverable — `preserve_keep`
pruning never sees a second ref to keep, and the defer notification (#333) names a ref whose
contents are no longer the attempt it describes.

### Environment

bmad-loop 0.9.0, claude adapter, Linux (WSL2), worktree isolation on.

I have a fix with tests and will open a PR against this issue.

Pawel-N-pl:
## What

The dirty-tree rollback snapshot now probes for an unused ref name instead of deriving one from
`task.attempt` alone.

## Why

`task.attempt` is not monotonic across a story's life — `runs.rearm_escalation` sets it back to 0
— so a post-resolve re-drive recomputes the pre-resolve rollback's exact ref name and overwrites
the only copy of that attempt's uncommitted work. The comment above the ref construction states
the opposite premise, which is what sent me looking.

Fixes #349.

## How

- Add `verify.ref_exists(repo, refname)`, a sibling of `branch_exists` that can see full refnames
  outside `refs/heads/`. Best-effort: any git failure reads as absent.
- In `recovery_flow._preserve_dirty_attempt`, keep `{slug}-{baseline}-{attempt}` as the base name
  and append `-r2`, `-r3`, … until free, so uniqueness is enforced against the refs that exist.
- Correct the comment.

## Testing

Two tests, both confirmed to fail with the fix reverted: a re-arm that previously collided now
parks a second distinct ref with both snapshots intact, and `ref_exists` is covered directly.
Full suite: 3458 passed, 35 skipped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented recovery snapshots from overwriting earlier snapshots when retry attempts reuse the same identifier.
  * Added unique suffixes to distinguish multiple preserved snapshots from the same baseline.
  * Improved handling of recovery references across repeated rollback and re-drive scenarios.

* **Tests**
  * Added coverage confirming preserved snapshots retain the correct content and receive deterministic unique names.
  * Added validation for detecting existing repository references.

### DIFF
--- a/src/bmad_loop/recovery_flow.py
+++ b/src/bmad_loop/recovery_flow.py
@@ -425,9 +425,20 @@ class RecoveryFlow:
         slug = safe_ref_segment(self.state.run_id)
         # ``baseline_commit`` is fixed across the whole dev retry loop, so keying the
         # ref on the baseline alone would make a 2nd dirty rollback reuse the name and
-        # orphan the 1st attempt's snapshot. ``task.attempt`` only ever increments
-        # (never resets), so it uniquely discriminates each retry's recovery ref.
-        ref = f"refs/attempt-preserve-dirty/{slug}-{baseline[:8]}-{task.attempt}"
+        # orphan the 1st attempt's snapshot. ``task.attempt`` discriminates the
+        # retries of one arming but is NOT monotonic across the story's life:
+        # runs.rearm_escalation resets it to 0, and a resolve session that commits
+        # nothing leaves HEAD == baseline, so the post-resolve re-drive's rollback
+        # recomputes the exact {slug}-{baseline}-{attempt} name of the pre-resolve
+        # rollback and would overwrite that snapshot, destroying the only copy of
+        # the first attempt's work. Probe for a free name instead of trusting the
+        # counter: uniqueness is enforced against the refs that actually exist.
+        base_ref = f"refs/attempt-preserve-dirty/{slug}-{baseline[:8]}-{task.attempt}"
+        ref = base_ref
+        serial = 2
+        while verify.ref_exists(workspace.root, ref):
+            ref = f"{base_ref}-r{serial}"
+            serial += 1
         try:
             parked = verify.snapshot_worktree(
                 workspace.root, ref, baseline_untracked=task.baseline_untracked
--- a/src/bmad_loop/recovery_flow.py
+++ b/src/bmad_loop/recovery_flow.py
@@ -433,13 +433,18 @@ class RecoveryFlow:
         # rollback and would overwrite that snapshot, destroying the only copy of
         # the first attempt's work. Probe for a free name instead of trusting the
         # counter: uniqueness is enforced against the refs that actually exist.
+        # The probe runs INSIDE the try: `ref_exists` spawns git, and a timeout or
+        # spawn fault arrives as GitError/GitSpawnError rather than a return code.
+        # Uncaught it would crash the rollback here — the one thing this handler
+        # exists to prevent — so a probe that cannot run degrades into the same
+        # "preservation is observation" path as a snapshot that cannot be written.
         base_ref = f"refs/attempt-preserve-dirty/{slug}-{baseline[:8]}-{task.attempt}"
         ref = base_ref
         serial = 2
-        while verify.ref_exists(workspace.root, ref):
-            ref = f"{base_ref}-r{serial}"
-            serial += 1
         try:
+            while verify.ref_exists(workspace.root, ref):
+                ref = f"{base_ref}-r{serial}"
+                serial += 1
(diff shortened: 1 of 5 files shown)


Pawel-N-pl:
Good catch, and correct on both halves — thanks.

`_run_git` translates a timeout (#156) and a spawn OSError (#343) into `GitError`/`GitSpawnError` rather than a return code, so with the probe above the `try` either would have crashed the rollback uncaught. Moved the loop inside the handler, so a probe that cannot run degrades into the same "preservation is observation" pause as a snapshot that cannot be written.

I kept `ref_exists` propagating those faults rather than swallowing them: reading "git could not run" as "the name is free" would overwrite the snapshot the probe exists to protect. The docstring was overpromising — it now says a non-zero exit reads as absent while spawn/timeout faults propagate typed.

Added `test_ref_probe_git_fault_degrades_like_a_failed_snapshot`, with the ablation named in its docstring: move the loop back above the `try` and it fails with the raw `GitSpawnError` instead of the typed pause.

Pawel-N-pl:
Moving this to draft — I gather a larger rework is planned, so I'd rather not sit in your review queue against code that's about to change.

The branch stays up and CI is green (py3.11–3.14, Windows, trunk, pyright). Happy to rebase it onto whatever lands, or close it if the rework covers this — just say the word. Issue #349 stays open either way, since the defect outlives the patch.

pbean:
Triage 2026-08-08: reviewed against current main — this is correct and ready. Verified the bug is live: recovery_flow.py:430 keys the preserve ref on `task.attempt`, and runs.py:802 sets `task.attempt = 0` in `rearm_escalation`, which falsifies the comment this PR removes (attempt DOES reset). The branch is mergeable/clean, and the probe-inside-the-try plus the `ref_exists` verify sibling are the right shapes. Please mark it ready for review — it lands first in the current queue.

pbean:
@codex review

pbean:
Merged as a plain merge commit — thank you, @Pawel-N-pl, and sorry for the long wait in the queue.

Three things I verified before landing it rather than taking on faith:

**The bug is live and this targets the right path.** `runs.py:802` sets `task.attempt = 0` inside `rearm_escalation`, so the counter is not monotonic across a story's life — which is exactly what the comment you removed at `recovery_flow.py:428` claimed it was. A post-resolve re-drive really could recompute the pre-resolve rollback's ref name and overwrite the only copy of that attempt's work.

**The probe degrades rather than crashes.** You moved the loop inside the `try:` after the first round of review, and that was the right call: `GitSpawnError` subclasses `GitError` (`verify.py:79`), so the `except (verify.GitError, OSError)` at `recovery_flow.py:451` does catch a spawn or timeout fault. A probe that cannot run now degrades into the same "preservation is observation" path as a snapshot that cannot be written, instead of crashing the one handler that exists to prevent a crash there.

**`ref_exists` doesn't mistake a failure for a free name.** `verify.py:791` reads a non-zero exit as "absent", which does fold "missing ref" together with "malformed name" — but the docstring says so plainly, the subsequent ref write surfaces anything real, and the faults that would actually be dangerous to swallow arrive typed and propagate.

One residual I'm noting for the record rather than asking you to change: the probe loop has no explicit cap. It does terminate — `serial` increments monotonically and the existing-ref set is finite — and in practice `scm.preserve_keep` (default 20) prunes the namespace at run start, so the scan is short. With `preserve_keep = 0` (never prune) it grows with the namespace. That's a cost question, not a correctness one, and it does not touch the data-loss fix. I'll track it separately if it ever bites.

Issue #357, opened 2026-07-28
frontmatter: set_frontmatter_status does not preserve the bytes its contract promises


pbean:
Follow-up from #355, where `set_frontmatter_status` was rewritten to verify its edit (the reader/writer split that let a corrupting or no-op write report success). Two non-preservations were pinned there as *characterization* rather than fixed, because both need their own evidence and one of them changes bytes on the standard fixture shape. Filing them together: they are the same contract, the same function, and the same line.

The function's contract is "a minimal in-place line replacement (not a YAML round-trip) so the spec's formatting, comments, and field order survive — only the status value changes". These are the two places it does not hold.

## 1. Line endings are relaid across the whole file

`set_frontmatter_status` reads with `Path.read_text` and writes with `Path.write_text` (`src/bmad_loop/frontmatter.py`). Universal-newline handling means a CRLF spec is read as LF and written back as LF **in its entirety** — not just the status line, and not just the frontmatter. For a function whose contract is "formatting survives", rewriting every line ending in the file is the largest possible violation of it, and it has always been there.

The fix is the one already reasoned out for the sibling writer in `devcontract.py` (see the comment above `_atomic_write_spec`): `read_bytes().decode()` + `write_bytes(....encode())`, with the detected line ending reused.

⚠️ `frontmatter.py`'s module docstring deliberately bars importing `platform_util` (it pulls in `subprocess`), which is why this module writes with `write_text` rather than reusing `atomic_replace`. So the fix cannot simply borrow the sibling's helper — it needs its own CRLF characterization tests.

## 2. A trailing inline comment on the status line is dropped

`_replace_value` keeps everything through the colon and then writes `<space><value>`, so `status: draft  # set by hand` becomes `status: done`. Preserving it requires knowing where the scalar ends, and a `#` inside a quoted value is not a comment — a wrong guess turns a working spec into a `FrontmatterWriteError`. That is exactly the guesswork the verified-edit rewrite exists to stop making, so it was pinned as a deliberate drop rather than half-solved.

Doing it safely now has an oracle it did not have before: the verified edit already re-parses the candidate block, so a comment-preserving render can be *attempted* and simply rejected by `_verified` if it did not mean what it claimed. That is a much better position than the pre-#355 code was in.

## While in here

[...]

PR #364, opened 2026-07-28
fix(frontmatter,verify,devcontract): preserve spec line endings

pbean:
Part 1 of #357.

## The defect

All four spec frontmatter writers read with `Path.read_text`. Universal-newline
handling means a CRLF spec arrives in memory as LF — **in its entirety**, not
just the line being edited. Writing that back relays every line ending in the
file, from functions whose contract is "a minimal in-place line replacement so
the spec's formatting, comments, and field order survive — only the value
changes". Two of the four compounded it with `Path.write_text`, whose
`newline=None` default translates `\n` to `os.linesep`, so on Windows an all-LF
spec came back all-CRLF.

| Writer | Read | Write |
| --- | --- | --- |
| `frontmatter.set_frontmatter_status` | `read_text` ❌ | `write_text` ❌ |
| `verify.set_frontmatter_field` | `read_text` ❌ | `write_text` ❌ |
| `devcontract.reset_spec_status` | `read_text` ❌ | `_atomic_write_spec` ✅ |
| `devcontract.strip_auto_run_result` | `read_text` ❌ | `_atomic_write_spec` ✅ |

`devcontract`'s write side was already byte-preserving, and the comment above
`_atomic_write_spec` had already reasoned this out for `append_auto_run_result`
— which reads bytes. Its two siblings never got the same read.

## The fix

`read_bytes().decode("utf-8")` on all four; `write_bytes(....encode("utf-8"))`
on the two `frontmatter` writers. `frontmatter.py`'s module docstring bars
importing `platform_util` (it drags in `subprocess`), so this cannot borrow
`atomic_replace` — the write stays non-atomic, which is a separate concern from
a byte-preserving one and out of scope here.

`_replace_value` now carries the edited line's **own** terminator rather than
re-emitting a flat `"\n"`, which would have left a CRLF spec with exactly one
bare-LF line — the very line the writer was asked to touch. `splitlines(keepends=True)`
guarantees at most one terminator per line, so `\r\n`, `\n`, a bare `\r`, and a
final line with no terminator each round-trip as authored.

`devcontract._render_status_line` needed no change: `_FM_STATUS_RE`'s
`(?P<rest>.*)$` already captures the `\r` of a raw CRLF line (verified before
touching it, not assumed).

Side effect: this makes `_edit_frontmatter_block`'s insert branch reachable for
the first time — before the byte-level read, the block was always LF by the time
it got there. It is now covered on both writers that can insert.

**Second commit, flagged for the maintainer as a scope call.** Making that branch
live exposed that it hardcoded `"\r\n" if block.endswith("\r\n") else "\n"`,
which agrees with the line it follows for CRLF and LF but not for a bare `\r` —
so an insert into a CR-only spec appended the file's *only* LF line. That is the
same hardcoded-`\n` defect as `_replace_value`'s, in the sibling branch of the
same function, and leaving it would have shipped a writer where one branch
carries endings and the other invents them. `_last_line_ending` reads the
terminator off the block's last line instead. Happy to split this out if you'd
rather part 1 stay strictly to the four read/write swaps.

## Behavior delta, pinned rather than fixed

`_FRONTMATTER_RE` and `AUTO_RUN_HEADING_RE` are line-oriented on `\r?\n`. A
CR-only spec used to be normalized by `read_text` on the way in, so the edit
landed (and the file was rewritten as LF, which was the defect). Reading the
bytes as authored, it finds no frontmatter block / no heading and returns a
silent `False`.

`frontmatter.set_frontmatter_status` is `splitlines`-based and still rewrites
that same shape, so the two siblings now disagree on CR-only input. That
asymmetry is documented at both writers and pinned by
`test_reset_status_no_ops_on_a_cr_only_spec` / `test_strip_auto_run_result_no_ops_on_a_cr_only_spec`,
which assert the no-op *and* that the sibling rewrites the identical bytes.
Widening the patterns to `\r` would change what counts as a line for every
reader keyed on them (and `^` in `MULTILINE` would not follow anyway) — a larger
contract than this fix owns, and no BMAD tool authors CR-only files.

## Tests

New coverage, all byte-exact:

- `tests/test_frontmatter.py` — a `line endings` section: CRLF whole-file
  (only the status line changes, plus `"\n" not in text.replace("\r\n", "")`),
  bare-CR round-trip, and a mixed-ending file where the status line is the odd
  one out on purpose, so a whole-file `nl` detection fails it.
- `tests/test_resolve.py` — `set_frontmatter_field` CRLF replace, CRLF insert, and
  a CR-only insert that pins "the inserted line is never the file's odd one out".
- `tests/test_devcontract.py` — `reset_spec_status` CRLF replace + CRLF insert,
  `strip_auto_run_result` CRLF preservation, an append→strip CRLF byte-for-byte
  round-trip, and the two CR-only characterizations.

`_as_written` (the `os.linesep`-modelling helper) is **deleted** and its ten call
sites reverted to bare byte-exact comparisons. It existed to keep the byte-exact
gate green on Windows by *describing* the relay; with the relay gone, those bare
comparisons are what fails on Windows if the write half regresses.

## Ablations

Every gate was reverted individually by a scripted patch → run → restore, with
the restore sha256-verified against the original (no `git checkout <file>`).

| Ablation | Result |
| --- | --- |
| `frontmatter` read_bytes → read_text | CRLF + bare-CR + mixed all **fail** ✅ |
| `_replace_value` per-line ending → bare `"\n"` | CRLF + bare-CR **fail**; plain-LF flip still passes ✅ |
| `_replace_value` → CRLF/LF only (bare CR dropped) | bare-CR **fails**; CRLF still passes ✅ |
| `_replace_value` → hardcoded `"\r\n"` | mixed **fails**; CRLF still passes ✅ |
| `verify` read_bytes → read_text | both new resolve tests **fail** ✅ |
| `devcontract.reset_spec_status` read revert | all three of its tests **fail** ✅ |
| `devcontract.strip_auto_run_result` read revert | all three of its tests **fail** ✅ |
| `_last_line_ending` → the old `block.endswith("\r\n")` branch | CR-only insert **fails**; CRLF insert still passes ✅ |

The "still passes" rows matter as much as the failures: they show each test
discriminates a specific wrong implementation rather than firing on any change.

**One honest limit.** Reverting `write_bytes` → `write_text` keeps the whole
suite green on Linux — `os.linesep` is `\n` there, so the write half is
POSIX-invisible. The **Windows CI leg is its only oracle**, via the reverted
byte-exact assertions in the characterization half. That is recorded in a
comment in `tests/test_frontmatter.py` so the next reader does not conclude the
write change is untested.

## Verification

- `uv run pytest -q -n auto` — 3671 passed, 24 skipped
- `trunk check --no-fix` — no issues
- `uvx pyright@1.1.411` — 0 errors

Out of scope, staying in #357: the trailing-inline-comment drop (part 2), moving
`set_frontmatter_status`'s tests out of `tests/test_resolve.py`, and making these
writes atomic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -248,6 +248,16 @@ story <id>`, the same annotation a sweep bundle writes. Both sprint and stories 
 
 ### Fixed
 
+- **Spec writers no longer relay a spec's line endings (#357, part 1).** `set_frontmatter_status`,
+  `set_frontmatter_field`, `reset_spec_status`, and `strip_auto_run_result` read specs through
+  `read_text`, whose universal-newline translation handed each writer an all-LF copy of a CRLF spec —
+  so a write contracted to move one value rewrote every line ending in the file (to LF everywhere, and
+  to CRLF for an all-LF spec on Windows). All four now read bytes and decode; the two `frontmatter`
+  writers write bytes too, and a replaced line carries its own terminator instead of a flat `\n`.
+  A CRLF spec stays CRLF, a mixed-ending spec keeps each line's ending, and only the value moves.
+  One pinned delta: a CR-only spec — never authored by a BMAD tool — is now a clean no-op through
+  `reset_spec_status` / `strip_auto_run_result`, whose patterns are line-oriented on `\r?\n`.
+
 - **A spec status the writer cannot rewrite is no longer a silent no-op.** `set_frontmatter_status`
   found the line with `lstrip().startswith("status:")` while every reader parses the block as YAML,
   so the two disagreed in both directions: a quoted key (`"status": x`), a space before the colon,
--- a/src/bmad_loop/devcontract.py
+++ b/src/bmad_loop/devcontract.py
@@ -505,7 +505,21 @@ def reset_spec_status(spec_path: Path, new_status: str) -> bool:
     reaches the other three call sites, which treat it as a repair write."""
     if not spec_path.is_file():
         return False
-    text = spec_path.read_text(encoding="utf-8")
+    # Raw read (not read_text), like the append/strip siblings: universal-newline
+    # translation would hand this writer an all-LF copy of a CRLF spec and
+    # `_atomic_write_spec` would then persist it, relaying every line ending in
+    # the file from a repair contracted to move one value.
+    #
+    # Reading the bytes as authored makes one asymmetry with
+    # `frontmatter.set_frontmatter_status` observable, and it is pinned rather
+    # than fixed: that writer splits with `splitlines`, which treats a bare `\r`
+    # as a line break, so a CR-only spec rewrites fine there. `_FRONTMATTER_RE`
+    # and `_FM_STATUS_RE` are line-oriented on `\r?\n`, so a CR-only spec finds
+    # no frontmatter block here and returns False. Widening them to `\r` would
+    # change what counts as a line for every reader keyed on these patterns
+    # (`^` in MULTILINE still would not follow), which is a larger contract than
+    # this fix owns — no BMAD tool authors CR-only files.
+    text = spec_path.read_bytes().decode("utf-8")
     fm = _FRONTMATTER_RE.match(text)
     if not fm:
         return False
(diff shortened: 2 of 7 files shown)


pbean:
@coderabbitai review

PR #365, opened 2026-07-29
fix(frontmatter): keep a trailing comment on the status line

pbean:
Part 2 of #357.

`frontmatter._replace_value` keeps everything through the colon and then writes `<space><value>`, so a trailing inline comment on the line it edits is dropped: `status: draft  # set by hand` comes back as `status: done`. That is the last part of the writers' "a minimal in-place line replacement … so the spec's formatting, comments, and field order survive" contract that was not honored. Part 1 (line endings, #364) fixed the other one.

The comment and its separating whitespace now carry through, whenever a conservative token pattern can certify where the scalar ends. Anything it cannot read as a bare token falls back to today's full-drop render.

## The issue text is wrong about the oracle, and the fix depends on that

> a comment-preserving render can be *attempted* and simply rejected by `_verified` if it did not mean what it claimed

**This is false, and it is the whole reason the change is shaped the way it is.** `_verified` re-parses the candidate block with `yaml.safe_load`, which strips comments *before* the oracle compares. A fabricated comment is therefore invisible to all three of its gates.

Concretely, for `status: "a # b"` — where the `#` is scalar text, not a comment — a naive split-at-`#` render produces:

```yaml
status: done # b"
```

That parses as `{"status": "done"}` with every other key untouched. It passes "still a mapping", "the key is exactly the value", and "every other key unchanged", the write returns `True`, and the junk lands in the user's spec. Silently.

So the gating here is the **reverse** of everywhere else in this module: the regex is the gate and `_verified` is only the backstop.

```python
_VALUE_COMMENT_RE = re.compile(
    r"^[ \t]*(?P<q>['\"]?)(?P<val>[A-Za-z0-9._-]*)(?P=q)(?P<sep>[ \t]+)(?P<comment>#.*)$"
)
```

Matched against the remainder after the colon (line ending stripped). Only `sep` and `comment` are data — `q` and `val` are **gates**, certifying the scalar boundary. `val` must never be widened to `[^#]*?`; the code comment says so and explains why. `sep`'s `[ \t]+` is load-bearing too: YAML needs whitespace before a `#` for it to open a comment at all.

If a reviewer proposes the "simpler" oracle-only version the issue describes, this is the reason it does not work.

## Ablation

The quoted-`#` test is an "X is absent" assertion, so it was ablated rather than trusted. Widening `val` to `[^#]*?` (scripted patch → run → restore, sha256-verified identical after):

```
>       assert spec.read_bytes().decode() == "---\nstatus: done\ntitle: t\n---\nbody\n"
E         - status: done
E         + status: done # b"
```

It fails on exactly the junk comment, and — the part that matters — the write **succeeded**: `set_frontmatter_status` returned `True` and `_verified` passed it. The other seven tests in the section stayed green under the ablation, so that one test is the sole holder of this line.

## Behavior

| status line | result |
| --- | --- |
| `status: in-review  # c` | `status: done  # c` — separator whitespace as authored |
| `status: 'in-review' # c` | `status: done # c` — comment kept, quotes still dropped |
| `status:  # c` (empty value) | `status: done # c` |
| `status: "a # b"` | `status: done` — fallback, no fabricated comment |
| `status: in-review#x` | `status: done` — `#x` was part of the scalar |
| `status:# c` | full drop (renderer-level only; PyYAML reads that line as a plain scalar, so no spec can reach it — the test asserts that rather than claiming it) |
| comment + CRLF | both survive — stacks on part 1 |

**Quotes are still dropped, deliberately.** `conftest.write_spec` writes `status: '<v>'` and three tests read the result back as an unquoted `status: done` (`tests/test_runs.py:772`, `tests/test_stories_e2e.py:594,793`). Preserving them here breaks those three.

## Inherited ripple

`verify.set_frontmatter_field` uses the same default renderer, so it picks up the carry with no change of its own. Its docstring has always claimed "the spec's formatting, comments, and field order survive" — that was true of every line except the one it edited, and is now true as written. Pinned with a test in `tests/test_resolve.py`, next to that helper's other tests.

`devcontract.reset_spec_status` is untouched: it passes its own renderer and already preserved both the comment and the quote style. Its docstring said the sibling "drops both", which is now half wrong; corrected.

## Verification

- `uv run pytest -q -n auto` → 3678 passed, 24 skipped
- `trunk check --no-fix` → no issues (`--no-fix` because F401 autofix deletes `verify.py`'s re-exports)
- `uvx pyright@1.1.411` → 0 errors

## A sibling with the opposite shape, left alone on purpose

`sprintstatus._set_mapping_value` already preserves inline comments on the board, using the wide `val` pattern this PR refuses (`\S(?:.*?\S)?`) — so `status: "a # b"` there splits to `val='"a'`, `rest=' # b"'` and has the same fabricated-comment defect, with no oracle at all behind it.

It is not fixed here. The board is orchestrator-owned and its values are status tokens and timestamps, and its `val` legitimately contains spaces (`last_updated: 01-06-2026 10:00`), so it cannot borrow this token class — it needs its own shape and its own test. Filed as #366.

What this PR does do is make the asymmetry deliberate in the code, so the next reader does not "unify" the two by widening the spec-side one back into the bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -248,6 +248,16 @@ story <id>`, the same annotation a sweep bundle writes. Both sprint and stories 
 
 ### Fixed
 
+- **A trailing inline comment on a spec's `status:` line survives the write (#357, part 2).**
+  `set_frontmatter_status` and `set_frontmatter_field` kept everything through the colon and dropped
+  the rest, so `status: draft  # set by hand` came back as `status: done` — the last part of
+  "formatting and comments survive" the writers did not honor. The comment and its separating
+  whitespace now carry through whenever a conservative token pattern can certify where the scalar
+  ends; a value it cannot read as a bare token (a quoted `#`, a `#` abutting the value) falls back to
+  today's full drop rather than guessing. The verified-edit oracle is only a backstop here: it strips
+  comments before it compares, so it cannot see a fabricated one. The value's own quotes are still
+  dropped — deliberately, and pinned.
+
 - **Spec writers no longer relay a spec's line endings (#357, part 1).** `set_frontmatter_status`,
   `set_frontmatter_field`, `reset_spec_status`, and `strip_auto_run_result` read specs through
   `read_text`, whose universal-newline translation handed each writer an all-LF copy of a CRLF spec —
--- a/src/bmad_loop/devcontract.py
+++ b/src/bmad_loop/devcontract.py
@@ -459,10 +459,11 @@ def _render_status_line(line: str, m: re.Match[str], value: str) -> str:
     """`reset_spec_status`'s replacement for a matched ``status:`` line: the
     value moves, its quote style and any trailing inline comment stay.
 
-    Deliberately different from `frontmatter._replace_value`, which drops both —
-    that writer's callers read the result back as a bare ``status: done``, this
-    one's tests pin the preservation. Sharing the VERIFICATION is the point; the
-    formatting each already had right."""
+    Deliberately different from `frontmatter._replace_value`, which carries the
+    comment through but drops the quotes — that writer's callers read the result
+    back as a bare ``status: done``, this one's tests pin the quote style.
+    Sharing the VERIFICATION is the point; the formatting each already had
+    right."""
     # Guarantee `key: value` spacing: a bare `status:` (no trailing space) would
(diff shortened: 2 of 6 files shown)


pbean:
@coderabbitai review

Issue #356, opened 2026-07-28
bmad-loop confirm only works on the machine that parked the story


pbean:
Follow-up from #355 (awaiting-operator part 3), where the trade-off was taken deliberately and documented rather than papered over.

## The gap

`bmad-loop confirm` reads `.bmad-loop/operator-actions.json`, which is **machine-local and never committed** (it is registered in the repository's local git exclude). So a story parked on the machine that ran the orchestrator can only be confirmed *on that machine*. A teammate, a fresh clone, or CI sees the parked spec and the parked board but has no way to complete the story with the command.

That is a real product gap, not just an aesthetic one: the human who performs the external action (buy the domain, publish the DNS record, grant the API key) is not necessarily sitting at the box that ran the unattended loop.

## Why it was not simply committed

The index is written from *inside* a story's commit window, in `_finalize_commit_phase`. Every way of committing it there is wrong:

- **on its own path** (`verify.commit_paths`, as `decisions.apply_pre_answer` does) — shifts `HEAD` past the `commit_sha` the park just stamped; and under `scm.isolation = "worktree"` it advances the **target** branch while the unit branch is still out, so an `scm.merge_strategy = "ff"` merge-back fails outright and the unit is kept + escalated.
- **folded into the story's own commit** (written before `finalize_commit`'s `git add -A`, as `_close_declared_deferred` annotations are) — only reachable inside the worktree, but the human reads the index at the project root. It would also invite a merge conflict between two parked stories whose worktrees were created before the first merged.
- **left merely untracked** — dirties the tree the epic-boundary auto-sweep refuses to run on (`_maybe_auto_sweep`), and the next story's `git add -A` sweeps one story's bookkeeping into another story's commit.

## Why the spec alone is not enough

`confirm` cannot fall back to reading the spec when the index has no entry: `spec_file` is reported by the dev session in its result JSON (`engine.py` / `verify.py` set `task.spec_file` from `rj["spec_file"]`) and is **not derivable from a story key** — there is no by-key spec discovery anywhere in the codebase. So the path to a parked story's spec survives only in the index. #355 refuses and explains rather than guessing with a glob.

## What a fix needs

A committed index plus the commit sequencing to make it safe. Sketch, not a design:

- write the index at a point where committing it cannot collide with the story's own commit or the worktree merge-back — most likely **after** `integrate_unit`, once the unit is merged and the target branch is the only thing moving;
[...]

PR #373, opened 2026-07-29
fix(install): guard the worktree exclude's filesystem tail (#359)

### LINKED ISSUE #359: install: _worktree_local_exclude does unguarded filesystem work outside its only try
Surfaced while fixing the park-path guards in #355. That PR contains the fault *for its own caller*; the helper itself is still unguarded for everyone else.

## The gap

`install._worktree_local_exclude` describes itself as "Best-effort — skipped if git can't be queried". Its `try` covers only the `git rev-parse --git-common-dir` subprocess:

```python
    try:
        common = subprocess.run([...], check=True).stdout.strip()
    except (subprocess.SubprocessError, OSError):
        return
    common_dir = Path(common)
    if not common_dir.is_absolute():
        common_dir = (worktree / common_dir).resolve()   # <-- outside the try
    exclude = common_dir / "info" / "exclude"
    exclude.parent.mkdir(parents=True, exist_ok=True)    # <-- outside the try
    existing = exclude.read_text(encoding="utf-8") if exclude.is_file() else ""
    ...
    exclude.write_text(prefix + "\n".join(new) + "\n", encoding="utf-8")   # <-- outside the try
```

Everything after the subprocess is unguarded filesystem work: a `.resolve()`, a `mkdir`, a `read_text`, and a `write_text`. Any of them can raise, and none of them is "git could not be queried" — so a function documented as best-effort propagates on a read-only `.git`, a permission fault, or an undecodable exclude file.

The `.resolve()` is the sharpest of the four because it raises a type the callers' `except OSError` guards do not catch: **below Python 3.13 `Path.resolve` reports a symlink loop as `RuntimeError`** (probed on this box — 3.11/3.12 raise it in default non-strict mode; 3.13 and 3.14 return normally). The repo floor is 3.11 and CI runs 3.11/3.12, so it is live on supported legs. It is reached only on the relative-`--git-common-dir` branch, which is the linked-worktree case.

## Callers

- `operatoractions._exclude_from_git` (the park index) — its caller `engine._index_park` was widened to `except (OSError, RuntimeError)` in #355, so that path is contained. The containment is at the *call site*, not here.
- worktree provisioning, which seeds its tool files through the same helper.

## Fix shape

Widen the helper's own guard to cover its filesystem tail rather than leaving each caller to discover which exception types leak out of a function that says it is best-effort. That keeps the documented contract true at one place instead of N, and matches the observe-degrade / repair-write asymmetry in AGENTS.md — this is bookkeeping, not a repair write, so degrading is correct.

Worth checking whether the worktree-provisioning caller wants the same degrade or would rather hear about it; they may not have the same answer, in which case the return type should say which happened.

pbean:
Closes #359.

`install._worktree_local_exclude` documented itself as "Best-effort — skipped if git can't be
queried", but the only `try` covered the `git rev-parse --git-common-dir` subprocess. Everything
after it was unguarded and propagated out of a function whose contract says it never does:

| Line | Call | Escapes as |
| ---- | ---- | ---------- |
| 751 | `(worktree / common_dir).resolve()` | **`RuntimeError`** on a symlink loop — 3.11/3.12 only, and the repo floor is 3.11 |
| 753 | `exclude.parent.mkdir(...)` | `OSError` (read-only `.git`) |
| 754 | `exclude.read_text(...)` | `OSError`, and **`UnicodeDecodeError`** on an exclude file that is not UTF-8 |
| 760 | `exclude.write_text(...)` | `OSError`, and **`UnicodeEncodeError`** on a pattern carrying a surrogate |

## Why the two arms cannot merge

`OSError` means the opposite thing on each side of the git call. In the subprocess arm it is "no git
to query" — an *expected* skip that must stay silent, since callers hand this plain non-repo
directories routinely. In the tail it is a filesystem fault worth surfacing. Same type, opposite
meaning, distinguishable only by location. So the tail gets its own guard catching exactly
`(OSError, RuntimeError, UnicodeDecodeError)` — explicit rather than `ValueError`-broad, which would
swallow programming errors ("typed escalation over bare except", AGENTS.md). The helper returns
`str | None`: `None` for success **and** for the expected skip, a reason string for a degradation.

## Swallowing alone would have been a regression

The exclude is what "shields [the provisioned tool files] even when a project doesn't" gitignore
them. A silently swallowed failure means the unit's `git add -A` commits `.claude/`, `.mcp.json` and
`_bmad/custom/` into the story merge — a loud crash traded for silent repo pollution. So the
surfacing lands in the same PR:

- `provision_worktree` gains a keyword-only `on_degraded: Callable[[str], None] | None = None`. The
  `list[str]` return contract is untouched, so the lazy `install.__getattr__` re-export and ~20
  existing callers/tests are unaffected.
- `run_isolated` wires it to `journal.append("worktree-exclude-degraded", story_key=..., error=msg)`,
  mirroring the `on_teardown_degraded` lambda a few lines below.

Journal-only, no `gates.notify`: the precedent `worktree-teardown-degraded` (housekeeping failed, run
continues) is journal-only, while `worktree-open-failed` gets notify because the operator must act
*now*. Here the run finishes normally and the harm is conditional and greppable.

## Correcting the issue's framing

**The `.resolve()` branch is the plain-checkout case, not the linked-worktree case.** Verified in a
scratch repo against git 2.55:

```
git -C <plain checkout>  rev-parse --git-common-dir  ->  .git                (relative)
git -C <linked worktree> rev-parse --git-common-dir  ->  /abs/path/repo/.git (absolute)
```

A linked worktree gets an absolute common dir and never reaches the `is_absolute()` fall-through. The
symlink-loop test therefore drives the helper against the main checkout — a linked worktree could not
exercise that line at all.

**The issue's caller list is stale.** `operatoractions._exclude_from_git` was deleted by #356, which
made the park record deliberately committed. There is exactly one live caller left —
`worktree_flow.provision_worktree` at worktree_flow.py:287, reached only from
`WorktreeFlow.run_isolated` — and this covers it. The engine's park-path containment from #355 is
untouched by this change, as is `engine._write_park_record`'s own `except (OSError, RuntimeError)`,
whose justification #356 already rewrote to "insurance".

## Tests

Seven helper/provision tests in `tests/test_install.py`, one engine-wiring test in
`tests/test_engine_worktree.py`. Faults are **injected** via name-filtered monkeypatch (the
`test_engine.py:1553` pattern) rather than built for real: 3.13+ resolves a symlink loop without
raising, so a hand-built loop would false-green on the newer legs. The `UnicodeDecodeError` test uses
real undecodable bytes, since that fault reproduces everywhere, and asserts the legacy bytes survive
untouched.

Every degrade and negative assertion carries its ablation in the docstring, and all six were run:

| Ablation | Fails |
| -------- | ----- |
| drop `RuntimeError` from the tail tuple | symlink-loop test |
| drop `OSError` from the tail tuple | write-fault test + `on_degraded` forwarding test |
| drop `UnicodeDecodeError` from the tail tuple | undecodable-exclude test |
| **inverse:** make the subprocess arm return a reason | both expected-skip tests |
| delete `on_degraded(reason)` in `provision_worktree` | forwarding test |
| delete the `on_degraded=` lambda in `run_isolated` | journal-wiring test |

The two expected-skip tests are negative assertions, so they needed the inverse ablation — a gate
deletion cannot fail them. The engine test's `worktree_seed` setup was checked the same way: without
it, `provision_worktree` short-circuits before the exclude step and the wiring under test is never
reached.

The five existing exclude happy-path tests are green **unmodified**.

## Follow-up commit: the codec fault has a write direction too

Review triage turned up one real gap in the first commit. The tail's guard named
`UnicodeDecodeError`, which is only the READ half. `write_text` raises the sibling
`UnicodeEncodeError`, and the two share no subclass but `UnicodeError` — so the best-effort
contract still leaked:

```
_worktree_local_exclude(repo, ["/vendor/weird-\udcff-name"])
-> UnicodeEncodeError: 'utf-8' codec can't encode character '\udcff': surrogates not allowed
```

Reachable rather than theoretical: `provision_worktree` derives one pattern per `seed_globs`
match through `rel.as_posix()`, so the pattern text comes from a real filename — a repo file
whose name is not valid UTF-8 arrives surrogate-escaped and cannot be written back as UTF-8.

Widened to `UnicodeError`: it covers both directions and stays far short of the
`ValueError`-broad catch the original rationale rejected, so a programming error still escapes.
The new test writes a literal `"\udcff"` rather than `os.fsdecode(b"...\xff...")` — fsdecode
decodes with surrogatepass on Windows, which rejects a lone invalid byte and would have raised
in the test's own setup, reddening only the Windows legs.

| Ablation | Fails |
| -------- | ----- |
| narrow the tuple back to `UnicodeDecodeError` | the new encode test alone |
| drop the codec member entirely | the encode test **and** the existing decode test |

Each half is independently pinned.

## Review outcome

Three findings raised, one design objection declined.

- **Applied:** the tail's codec guard covered only the read half (`UnicodeDecodeError`); `write_text`
  raises `UnicodeEncodeError` and the two share no subclass but `UnicodeError`. Reproduced end-to-end
  through `seed_globs`. Widened, tested, ablated both directions.
- **Applied:** a stale ablation note — after the widening, the decode test still told a reader to drop
  a tuple member that no longer existed. Corrected to the two ablations actually run.
- **Declined:** a proposal to raise a typed error and pause/escalate the unit instead of journaling.
  It inverts this repo's severity ladder — failing to *mount* a worktree at all defers rather than
  escalates ("Defer this unit rather than crash the whole run", worktree_flow.py), and the exclude is
  a second line of defense behind the project's own `.gitignore`. Escalating would also preserve the
  exact bug #359 filed: the run still dies over a best-effort shielding step. Reasoning is in the
  thread.

Two pre-existing hazards in the same helper were verified but deliberately left out of this change,
filed as #374 (the git-query arm decodes stdout strictly, so a repo path with bytes invalid in the
locale encoding still escapes) and #375 (the exclude write is not atomic, so a short write truncates
the operator's shared exclude while the reason string reports that nothing happened). Both want a
decision this PR's scope does not contain — where filesystem paths get decoded, and whether the write
becomes atomic.

## Verification

`uv run pytest -q -n auto` — 3693 passed, 24 skipped. `uvx pyright@1.1.411` — 0 errors. `trunk fmt` +
full `trunk check` — no issues.

## Summary by CodeRabbit

- **Bug Fixes**
  - Worktree setup now handles Git exclude-file and filesystem failures gracefully instead of stopping runs.
  - Non-Git directories and unavailable Git metadata are safely skipped.
  - Exclude-file updates preserve existing formatting and avoid duplicate entries.
  - When exclusion is degraded, the run continues and records a journal entry while still allowing successful completion and merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -260,6 +260,15 @@ story <id>`, the same annotation a sweep bundle writes. Both sprint and stories 
 
 ### Fixed
 
+- **A worktree's local git exclude really is best-effort now (#359).** `_worktree_local_exclude`
+  guarded only its `git rev-parse` call; the filesystem tail behind it (resolve, mkdir, read, write)
+  crashed the run — a symlink loop raises `RuntimeError` on 3.11/3.12, an exclude file that is not
+  UTF-8 raises `UnicodeDecodeError`, and a read-only `.git` raises `OSError`. The tail is guarded and
+  degrades to a reason string, and provisioning journals it as `worktree-exclude-degraded` on the
+  story rather than swallowing it: without the exclude the unit's `git add -A` would commit the
+  provisioned skill trees and tool configs into the story's merge. Git being unqueryable at all
+  (not a repo, no git) stays a silent, expected skip.
+
 - **A finished session whose spec left `status:` blank is rescued again (#369).** The second copy of
   the #358 stringification lived in `devcontract.synthesize_result`, which that fix did not reach: a
   present-but-blank `status:` read as the truthy token `"none"`, so the prose `## Auto Run Result`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -263,7 +263,8 @@ story <id>`, the same annotation a sweep bundle writes. Both sprint and stories 
 - **A worktree's local git exclude really is best-effort now (#359).** `_worktree_local_exclude`
   guarded only its `git rev-parse` call; the filesystem tail behind it (resolve, mkdir, read, write)
   crashed the run — a symlink loop raises `RuntimeError` on 3.11/3.12, an exclude file that is not
-  UTF-8 raises `UnicodeDecodeError`, and a read-only `.git` raises `OSError`. The tail is guarded and
+  UTF-8 raises `UnicodeDecodeError`, a seeded path whose filename is not UTF-8 raises
+  `UnicodeEncodeError` on the way back out, and a read-only `.git` raises `OSError`. The tail is guarded and
   degrades to a reason string, and provisioning journals it as `worktree-exclude-degraded` on the
   story rather than swallowing it: without the exclude the unit's `git add -A` would commit the
   provisioned skill trees and tool configs into the story's merge. Git being unqueryable at all
--- a/src/bmad_loop/install.py
+++ b/src/bmad_loop/install.py
@@ -729,11 +729,23 @@ def _copy_traversable(src, dst: Path, *, skip_existing: bool = False) -> bool:
     return True
 
 
-def _worktree_local_exclude(worktree: Path, patterns: Sequence[str]) -> None:
+def _worktree_local_exclude(worktree: Path, patterns: Sequence[str]) -> str | None:
     """Add anchored ignore patterns to the worktree's local git exclude so the
     provisioned tool files are never staged by the unit's `git add -A`. Uses
     git's standard local-only exclude (never committed or pushed); it does not
-    affect already-tracked files. Best-effort — skipped if git can't be queried.
+    affect already-tracked files.
+
+    Best-effort in two arms that must stay separate, because `OSError` means the
+    opposite thing in each:
+
+    - git unqueryable (not a repo, git missing) is an EXPECTED skip — returns
+      None silently. Callers hand this plain temp dirs routinely.
+    - a filesystem fault AFTER git answered degrades to a returned reason string
+      the caller can surface: `.resolve()` (pre-3.13 a symlink loop raises
+      RuntimeError, not OSError), `mkdir`, an exclude file that is not UTF-8
+      (UnicodeDecodeError), and read/write OSError. Silence here is not cosmetic:
+      without the exclude the unit's `git add -A` commits the tool files this
+      provisioning just wrote into the story's merge.
     """
     # Callers pass POSIX-slash patterns (glob rels via as_posix; config strings as
     # authored); git's exclude is POSIX-slash on every platform, so nothing to fix here.
@@ -745,19 +757,26 @@ def _worktree_local_exclude(worktree: Path, patterns: Sequence[str]) -> None:
             check=True,
         ).stdout.strip()
     except (subprocess.SubprocessError, OSError):
-        return
-    common_dir = Path(common)
-    if not common_dir.is_absolute():
-        common_dir = (worktree / common_dir).resolve()
-    exclude = common_dir / "info" / "exclude"
-    exclude.parent.mkdir(parents=True, exist_ok=True)
-    existing = exclude.read_text(encoding="utf-8") if exclude.is_file() else ""
-    present = set(existing.splitlines())
-    new = [p for p in patterns if p not in present]
-    if not new:
-        return
-    prefix = existing if not existing or existing.endswith("\n") else existing + "\n"
-    exclude.write_text(prefix + "\n".join(new) + "\n", encoding="utf-8")
+        return None
+    try:
+        common_dir = Path(common)
+        if not common_dir.is_absolute():
+            # a PLAIN checkout answers with a relative ".git"; a linked worktree
+            # answers with the main repo's absolute .git (verified, git 2.55), so
+            # this branch is the plain-checkout case, not the worktree one.
+            common_dir = (worktree / common_dir).resolve()
+        exclude = common_dir / "info" / "exclude"
+        exclude.parent.mkdir(parents=True, exist_ok=True)
+        existing = exclude.read_text(encoding="utf-8") if exclude.is_file() else ""
+        present = set(existing.splitlines())
+        new = [p for p in patterns if p not in present]
+        if not new:
+            return None
+        prefix = existing if not existing or existing.endswith("\n") else existing + "\n"
+        exclude.write_text(prefix + "\n".join(new) + "\n", encoding="utf-8")
+    except (OSError, RuntimeError, UnicodeDecodeError) as e:
+        return f"could not update the worktree-local git exclude ({worktree}): {e}"
+    return None
 
 
 def _copy_skills(project: Path, trees: Sequence[str], force: bool) -> bool:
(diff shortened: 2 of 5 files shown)


pbean:
@coderabbitai review

pbean:
@coderabbitai full review

pbean:
Verified against the code and rejecting this one — it is a severity-policy disagreement rather than a defect, and the repo's own escalation ladder points the other way.

**1. It would invert the repo's severity ordering.** Failing to *mount a worktree at all* is strictly more consequential than failing to write its backup shield, and that case does **not** escalate: `worktree_flow.py:448-459` defers the unit, journals `worktree-open-failed`, notifies, and the comment states the rule outright — *"Defer this unit rather than crash the whole run."* Pausing/escalating on a failed `info/exclude` write would make the backup layer louder than the primary failure it backs up.

**2. The exclude is a second line of defense, not the only one.** Per `provision_worktree`'s docstring (`worktree_flow.py:113-114`), the skill trees, hook config and seeded configs "all live in dirs projects gitignore — but the exclude shields them even when a project doesn't." The pollution you describe requires the exclude write to fail **and** the project to omit its tool dirs from `.gitignore`. That conditional harm is exactly why this is journaled and greppable rather than run-stopping.

**3. Escalation is the opposite of what #359 asked for.** The issue is that `_worktree_local_exclude` advertised "best-effort" while its filesystem tail could take down the run (pre-PR the fault propagated through `run_isolated` to `engine.py:1133-1138`, which wraps it in no `try`). Converting that into a typed pause preserves the bug the issue filed — the run still dies over a shielding step — just with better prose.

**4. The doctrine line is already satisfied, and the "repair write" half doesn't reach here.** "Typed escalation over bare except" is met: the tail catches an explicit `(OSError, RuntimeError, UnicodeDecodeError)`, deliberately narrow so programming errors still surface. The "repair writes must raise" half is the spec/ledger-read doctrine from #113 (anchor comment at `devcontract._read_text_or_empty`), which governs writes that repair **orchestrator state** — `reset_spec_status`, `set_frontmatter_status`, `mark_done`. A provisioning shield is not in that category.

**5. There is a closer precedent inside the same function.** A `worktree_seed` entry that silently no-ops leaves the worktree without its MCP/CLI config — arguably worse for the session that follows — and it is reported via the `skipped` return and journaled as `worktree-seed-skipped` while the run continues (`worktree_flow.py:490-498`).

The kernel of the concern is real and the PR does act on it: a swallowed fault would trade a loud crash for silent repo pollution, which is why the surfacing (`on_degraded` → `worktree-exclude-degraded`, carrying `story_key` + `error`) had to ship in the same PR rather than later. Journal-only over `gates.notify` was chosen deliberately against the two precedents above; the alternatives (callback-on-helper, `bool` return, notify, raise) were considered and rejected before implementation.

Leaving the design as-is.

pbean:
@coderabbitai review

pbean:
Good catch on the stale ablation note — fixed in 6f19355.

Widening the tail's tuple to `UnicodeError` left `test_worktree_local_exclude_degrades_on_undecodable_exclude` instructing a reader to "drop `UnicodeDecodeError` from the tail's except tuple", which is no longer a member, so the ablation could not be performed as written. In this repo an ablation note is the durable record that a negative assertion actually discriminates, so a stale one is worth more than cosmetic.

Replaced with the two ablations I re-ran to confirm the corrected wording:

| ablation | decode test | encode test |
| --- | --- | --- |
| drop `UnicodeError` entirely | **fails** | fails |
| narrow to `UnicodeDecodeError` | passes | **fails** |

Stating both is the point — it shows the codec member is pinned independently from each side rather than by one test twice.

PR #376, opened 2026-07-29
fix(install): close the exclude helper's two residual holes (#374, #375)

### LINKED ISSUE #374: _worktree_local_exclude's git-query arm can still raise UnicodeDecodeError
Follow-up from the review of PR #373 (the #359 best-effort guard). Verified, not theoretical — reproduced on this box.

`install._worktree_local_exclude` has two arms. PR #373 guarded the filesystem tail so it degrades to a reason string instead of propagating. The **git-query arm above it** still has one escape the tail now handles on both sides:

```python
common = subprocess.run(
    ["git", "-C", str(worktree), "rev-parse", "--git-common-dir"],
    capture_output=True,
    text=True,          # <-- decodes stdout with the locale codec, errors='strict'
    check=True,
).stdout.strip()
except (subprocess.SubprocessError, OSError):
    return None
```

`text=True` decodes git's stdout strictly. `UnicodeDecodeError` is neither a `SubprocessError` nor an `OSError`, so it escapes.

## Reproduction

POSIX filenames are bytes, so a repo whose absolute path carries bytes invalid in the locale encoding makes `git rev-parse --git-common-dir` emit exactly that. With a stub `git` on `PATH` printing a path containing byte `0xff`:

```
ESCAPED from arm 1: UnicodeDecodeError -> 'utf-8' codec can't decode byte 0xff in position 10: invalid start byte
```

Note a linked worktree gets the **main repo's absolute `.git`** back from `--git-common-dir` (a plain checkout gets a relative `.git`), so the whole main-repo path is in play, not just the worktree's own name.

## Why it matters now

This is pre-existing code, but PR #373 rewrote the docstring to characterize this arm as "git unqueryable (not a repo, git missing) is an EXPECTED skip — returns None silently", and its CHANGELOG entry says the helper "really is best-effort now". The escape falsifies both. It is also the **same codec-fault genus** the tail now handles in both directions (`UnicodeError` covers the read and the write), so the two arms of one function are asymmetric.

## Why it was left out of PR #373

Deliberately, to avoid an unbounded scope spiral. The obvious minimal fix — capture bytes and `os.fsdecode` inside the tail — makes the non-UTF-8 path *work* rather than fault, but then the surrogate-bearing path flows onward into the degrade reason and into `journal.append`, whose JSON write is itself UTF-8. Handling non-UTF-8 filesystem paths is a program-wide concern, not a `_worktree_local_exclude` one, so it wants its own change with its own decision about where paths get decoded.

(Checked while there: this is also a bare `subprocess.run` for git, outside the `verify._run_git` chokepoint that AGENTS.md names as a hard invariant, and it carries no `timeout=`. Worth confirming whether this call is a deliberate exemption — `install.py` importing `verify` may be the reason — or an oversight.)

### LINKED ISSUE #375: _worktree_local_exclude's exclude write is not atomic: a short write corrupts the shared exclude
Follow-up from the review of PR #373 (the #359 best-effort guard). Pre-existing; PR #373 guarded the *propagation* but the underlying write is still non-atomic.

`install._worktree_local_exclude` finishes with a read-modify-**rewrite**:

```python
existing = exclude.read_text(encoding="utf-8") if exclude.is_file() else ""
...
prefix = existing if not existing or existing.endswith("\n") else existing + "\n"
exclude.write_text(prefix + "\n".join(new) + "\n", encoding="utf-8")
```

`Path.write_text` is `open(mode="w")` — which **truncates immediately** — followed by `f.write(data)`. `prefix` carries the operator's pre-existing exclude content, so a fault partway through the write leaves the file truncated mid-content rather than unchanged.

## Consequences

1. **The reason string misdescribes the state.** The degrade reason is `could not update the worktree-local git exclude (...)`, which is accurate only when the *open* fails. When the write fails partway the file **was** updated — destructively. The caller is told nothing happened.

2. **The truncated tail is still a valid git pattern, and it is a prefix of the intended one.** Patterns are written anchored (`/.claude`, `/.mcp.json`, ...). If the cut lands one character into a pattern line, the surviving line is `/` — which git-excludes the **entire worktree**. That is precisely the state `worktree_flow.py` already has an explicit guard against for hookless profiles:
   > hookless profiles have no config_path — and their empty string would render as the pattern "/", git-excluding the entire worktree.
   
   A short write can manufacture that state while the caller sees only "could not update".

3. **Blast radius is the main repo, not just the unit.** For a linked worktree `git rev-parse --git-common-dir` returns the **main** repo's `.git`, so `info/exclude` is shared by the main checkout and every worktree. Truncating it drops the operator's own hand-written excludes repo-wide.

## Reproduction

Injecting a short write followed by `OSError(28, "No space left on device")`:

```
before:   b"# USER'S OWN EXCLUDE\n/secret-local\n"
after:    b"# USER'S OWN EXCLUDE\n/secret-local\n/aaaaa"     <- truncated, unterminated
returned: "could not update the worktree-local git exclude (...): [Errno 28] No space left on device"
```

Trigger is a mid-write fault (ENOSPC, EIO, quota) — plausible rather than routine.

## Note on test coverage

The existing write-fault test in `tests/test_install.py` **cannot** observe this: it monkeypatches `Path.write_text` itself, so the file is never opened and the truncation never happens. Any fix needs a test that lets the real `open` run and fails inside the write.

## Suggested shape

Write via a temp file in the same directory and `os.replace` onto the target, so the exclude is either fully updated or untouched — the repo already has an `atomic_replace` helper for this pattern. That is a change to the write strategy rather than to the exception guard, which is why it was kept out of PR #373.

pbean:
Closes #374. Closes #375.

Follow-up to #359 (PR #373), which guarded `_worktree_local_exclude`'s filesystem tail but left two
holes the review surfaced. Both live in the same helper and both falsify the same best-effort
contract, so they ship together rather than as two PRs racing on adjacent lines.

## #374 — the git-query arm decoded strictly

`subprocess.run(..., text=True)` decodes stdout with the locale codec at `errors='strict'`.
`UnicodeDecodeError` is neither a `SubprocessError` nor an `OSError`, so it escaped the one arm whose
entire contract is *"skip silently"*. POSIX filenames are bytes, so a repo path carrying bytes
invalid in the locale encoding reaches this on an ordinary box:

```
ESCAPED from arm 1: UnicodeDecodeError -> 'utf-8' codec can't decode byte 0xff in position 10
```

Stdout is now captured as **bytes** and decoded with `os.fsdecode` **inside the guarded tail**. Two
reasons for that placement: `fsdecode` uses surrogateescape on POSIX, so a non-UTF-8 path round-trips
back to the filesystem instead of faulting at all; and it can still raise on Windows, where
utf-8/surrogatepass rejects a lone invalid byte — so it belongs where `UnicodeError` is already
caught.

## #375 — the write was not atomic

The update is a read-modify-**rewrite**, and `Path.write_text` opens `"w"`, which truncates *before*
writing. `prefix` carries the operator's own excludes, so a fault partway through left the file cut
mid-content while the degrade reason still said "could not update":

```
before:   b"# OPERATOR'S OWN\n/secret-local\n"
after:    b"# OPERAT"                            <- ablation output, verbatim
```

Three consequences: the reason misdescribes the state; the surviving tail parses as a **valid git
pattern** and is a prefix of the intended one, so a cut one character in leaves the last line as `/`,
excluding the entire worktree — the exact state `worktree_flow.py` already guards against for
hookless profiles; and the blast radius is the **main repo**, since a linked worktree's
`--git-common-dir` points at the main `.git`.

Now written to a scratch file and moved into place with `atomic_replace`, with the scratch file
removed on the failure path. The exclude is either fully updated or untouched.

## One test was replaced after failing its own ablation

The first version of the #374 test monkeypatched `subprocess.run` to return undecodable bytes. It
passed — and **passed identically with `text=True` restored**, because handing the code bytes
directly never runs the stdlib's decoding at all. It could not see the bug it was written for. Driving
a **real stub `git` on PATH** does fail the ablation. It is POSIX-only and skipped on Windows, which
is correct rather than a dodge: Windows paths are UTF-16, so git cannot emit an undecodable one.

The #375 test injects at `Path.open`, not `Path.write_text` — patching `write_text` means the file is
never opened, so the truncation cannot happen and the test would pass against the very bug it exists
to catch. That is the same reason the two pre-existing write-fault tests needed their name filter
widened: the write now lands on the scratch file, and a filter pinned to `"exclude"` alone would stop
injecting anything and pass vacuously.

| Ablation | Fails |
| -------- | ----- |
| restore `text=True` on the subprocess | the stub-git test |
| write directly to `exclude` instead of tmp + `atomic_replace` | the short-write test (file returns truncated) |
| *(rejected test shape)* patch `subprocess.run` instead of stubbing git | **nothing** — which is why it was replaced |

## Verification

`uv run pytest -q -n auto` — 3695 passed, 24 skipped (run three times; one unrelated Textual modal
test, `test_decision_modal_scrolls_when_content_long`, flaked once under parallel load and passes
consistently in isolation and on `main`). `uvx pyright@1.1.411` — 0 errors. `trunk fmt` + full
`trunk check` — no issues.

## Summary by CodeRabbit

* **Bug Fixes**
  * Worktree exclude updates are now written atomically, preventing partial/truncated exclude files and preserving existing content.
  * Exclude update failures no longer leave behind scratch/temp artifacts.
  * Worktree setup now handles repositories with non-UTF-8 paths without crashing, degrading gracefully when needed.
* **Tests**
  * Added failure-injection and coverage for partial-write scenarios and undecodable `git` output, including validation that exclude content remains byte-identical.

---

Follow-up filed, not fixed here: #381 — `atomic_write_text` makes each write all-or-nothing but does not serialize the read-modify-write around it, so two runs provisioning against one repo interleave into a lost update. Pre-existing, reproduced in the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -260,6 +260,20 @@ story <id>`, the same annotation a sweep bundle writes. Both sprint and stories 
 
 ### Fixed
 
+- **A short write no longer truncates the worktree exclude (#375).** The exclude update is a
+  read-modify-rewrite, and `write_text` truncates before writing, so a fault partway through (ENOSPC,
+  EIO) left the operator's own excludes cut mid-content while the degrade reason still reported that
+  nothing had been written. Worse, the surviving tail parses as a valid git pattern and is a prefix
+  of the intended one — cut a character in and the last line is `/`, excluding the whole worktree —
+  and a linked worktree's common dir is the MAIN repo's `.git`, so the damage was repo-wide. Now
+  written to a scratch file and moved into place, so the exclude is either fully updated or untouched.
+
+- **The exclude's git query no longer crashes on a non-UTF-8 repo path (#374).** POSIX filenames are
+  bytes, and `git rev-parse --git-common-dir` was decoded strictly, so a repo path carrying bytes
+  invalid in the locale encoding raised `UnicodeDecodeError` — a type neither arm of the helper
+  caught, out of a function documented as best-effort. Git's output is captured as bytes and decoded
+  with the filesystem codec inside the guarded tail, so such a path round-trips instead of faulting.
+
 - **A worktree's local git exclude really is best-effort now (#359).** `_worktree_local_exclude`
   guarded only its `git rev-parse` call; the filesystem tail behind it (resolve, mkdir, read, write)
   crashed the run — a symlink loop raises `RuntimeError` on 3.11/3.12, an exclude file that is not
--- a/src/bmad_loop/install.py
+++ b/src/bmad_loop/install.py
@@ -762,6 +762,15 @@ def _worktree_local_exclude(worktree: Path, patterns: Sequence[str]) -> str | No
         # out of an arm whose whole contract is "skip silently", and that type is
         # in neither arm's tuple (#374). POSIX filenames are bytes, so a repo path
         # with bytes invalid in the locale encoding reaches this on a normal box.
+        #
+        # Deliberately NOT routed through verify._run_git, the chokepoint AGENTS.md
+        # otherwise requires. Nothing structural forbids it — neither module imports
+        # the other — but _run_git hands back CompletedProcess[str], i.e. the strict
+        # decode this call exists to avoid, and it raises GitError rather than the
+        # (SubprocessError, OSError) this arm treats as an expected skip. Adopting it
+        # needs a bytes mode on the chokepoint first; that is #377, which also covers
+        # the decode fault escaping _run_git's own taxonomy. The cost of standing
+        # outside it meanwhile is this call's missing timeout.
         raw = subprocess.run(
             ["git", "-C", str(worktree), "rev-parse", "--git-common-dir"],
             capture_output=True,
--- a/src/bmad_loop/install.py
+++ b/src/bmad_loop/install.py
@@ -781,7 +781,9 @@ def _worktree_local_exclude(worktree: Path, patterns: Sequence[str]) -> str | No
     try:
         # fsdecode, so a non-UTF-8 repo path round-trips back to the filesystem
         # instead of faulting. It can still raise on Windows (utf-8/surrogatepass
-        # rejects a lone invalid byte), which is why it sits inside this try.
+        # rejects a lone invalid byte), which is why it sits inside this try —
+        # defensive placement rather than a path under test: the regression test
+        # is POSIX-only because git on Windows has no such path to hand back.
         common_dir = Path(os.fsdecode(raw).strip())
         if not common_dir.is_absolute():
             # a PLAIN checkout answers with a relative ".git"; a linked worktree
@@ -807,8 +809,15 @@ def _worktree_local_exclude(worktree: Path, patterns: Sequence[str]) -> str | No
         # The helper rather than a hand-rolled tmp+replace: its temp name is
         # unique, and EVERY linked worktree of a repo resolves to this same
         # common dir, so a fixed `.tmp` sibling is a collision between two runs
-        # provisioning against one repo. It also fsyncs before the replace and
-        # carries the mode over, which a bare replace resets.
+        # provisioning against one repo. It also fsyncs before the replace, and
+        # carries the mode over when the exclude already exists, which a bare
+        # replace resets; one it creates from nothing gets the helper's
+        # deliberate 0600 rather than the umask default.
+        #
+        # Atomicity per write is NOT isolation across the read above and this
+        # write: they are unserialized, so two such runs interleaving lose one
+        # side's patterns outright, both returning None. That race predates
+        # #375 and survives this fix — it is #381.
         atomic_write_text(exclude, prefix + "\n".join(new) + "\n")
     # UnicodeError, not UnicodeDecodeError: the write raises the ENCODE sibling.
     # Still far short of ValueError-broad, so a programming error still escapes.
--- a/src/bmad_loop/install.py
+++ b/src/bmad_loop/install.py
@@ -17,6 +17,7 @@ orchestrator's signal watcher is CLI-agnostic.
 from __future__ import annotations
 
 import json
+import os
 import re
 import shutil
 import subprocess
@@ -28,6 +29,7 @@ from typing import Any, NamedTuple
 
 from .adapters.profile import ALIASES, CLIProfile, ProfileError, load_profiles
 from .checks import Finding
+from .platform_util import atomic_replace
 from .policy import POLICY_TEMPLATE
 from .process_host import get_process_host
 
@@ -739,12 +741,10 @@ def _worktree_local_exclude(worktree: Path, patterns: Sequence[str]) -> str | No
     opposite thing in each:
 
     - git unqueryable (not a repo, git missing) is an EXPECTED skip — returns
-      None silently. Callers hand this plain temp dirs routinely. One residual
-      escape survives here and is tracked in #374: `text=True` decodes git's
-      stdout strictly, so a repo path carrying bytes invalid in the locale
-      encoding raises UnicodeDecodeError, which is neither arm's type. Left out
-      of the #359 fix deliberately — decoding filesystem paths safely is a
-      program-wide decision, not this helper's.
+      None silently. Callers hand this plain temp dirs routinely. Nothing is
+      DECODED in this arm: git's stdout is captured as bytes and decoded in the
+      tail below, because a decode fault is a degrade (git answered; we could
+      not read it), not the silent skip this arm means (#374).
     - a filesystem fault AFTER git answered degrades to a returned reason string
       the caller can surface: `.resolve()` (pre-3.13 a symlink loop raises
       RuntimeError, not OSError), `mkdir`, read/write OSError, and either
@@ -758,16 +758,22 @@ def _worktree_local_exclude(worktree: Path, patterns: Sequence[str]) -> str | No
     # Callers pass POSIX-slash patterns (glob rels via as_posix; config strings as
     # authored); git's exclude is POSIX-slash on every platform, so nothing to fix here.
     try:
-        common = subprocess.run(
+        # bytes, not text=True: strict decoding here would raise UnicodeDecodeError
+        # out of an arm whose whole contract is "skip silently", and that type is
+        # in neither arm's tuple (#374). POSIX filenames are bytes, so a repo path
+        # with bytes invalid in the locale encoding reaches this on a normal box.
+        raw = subprocess.run(
             ["git", "-C", str(worktree), "rev-parse", "--git-common-dir"],
             capture_output=True,
-            text=True,
             check=True,
-        ).stdout.strip()
+        ).stdout
     except (subprocess.SubprocessError, OSError):
         return None
     try:
-        common_dir = Path(common)
+        # fsdecode, so a non-UTF-8 repo path round-trips back to the filesystem
+        # instead of faulting. It can still raise on Windows (utf-8/surrogatepass
+        # rejects a lone invalid byte), which is why it sits inside this try.
+        common_dir = Path(os.fsdecode(raw).strip())
         if not common_dir.is_absolute():
             # a PLAIN checkout answers with a relative ".git"; a linked worktree
             # answers with the main repo's absolute .git (verified, git 2.55), so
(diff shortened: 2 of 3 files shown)


pbean:
@coderabbitai full review

pbean:
All three addressed — two were real, and the `/tmp` one was a genuine catch. Thanks.

**1. Fixed scratch name racing on the shared common dir — valid, already fixed in 99ab745.** You reviewed the commit before it. The fix is not a hand-rolled unique name but `platform_util.atomic_write_text`, whose docstring names this exact hazard ("no fixed `.tmp` sibling for a concurrent writer of the same file to collide with") and which also fsyncs before the replace and carries the mode over — both of which a bare `os.replace` resets, since it swaps in a new inode.

Moving to it broke three test seams in a way worth recording, because each would have failed *silently*: the helper writes via `os.fdopen` on an `mkstemp` fd, so the two `Path.write_text` patches and the `Path.open` patch all stopped firing and the tests passed **vacuously**. They now patch `install.atomic_write_text` and `io.open` respectively — `io.open` being the one seam both shapes share, so the ablation keeps biting. The leftover-scratch assertion also named a literal filename, which is meaningless once the name is random; it now asserts the directory contains nothing but `exclude`.

**2. Stub `git` leaking real directories into shared `/tmp` — correct, fixed in 0066682.** Confirmed on this box before fixing: `/tmp/repo-<0xff>-name` existed, created by the test. `_worktree_local_exclude` does `mkdir(parents=True)` on whatever common dir it is handed, so the hardcoded path escaped `tmp_path` entirely. The stub now builds its path from `tmp_path` via `os.fsencode`, keeping the undecodable byte inside the sandbox, and the test asserts nothing was created outside it. The ablation was re-run against the new stub: restoring `text=True` still fails it.

**3. Route through `_run_git` — declined here, filed instead (#377).** You are right that this bypasses the chokepoint, and right that nothing structural forbids it: neither module imports the other. But `_run_git` returns `CompletedProcess[str]`, which is precisely the strict decode this PR exists to remove, and it raises `GitError` rather than the `(SubprocessError, OSError)` this arm treats as an expected skip. Adopting it would reintroduce the bug.

Taking your second option — an explicit exception with the rationale in the code, including the real cost (this call has no timeout). The bytes mode the chokepoint would need is #377, which also covers a bigger instance of the same defect: `_run_git`'s own tuple omits `UnicodeDecodeError`, so it escapes untyped past all ~50 `except GitError` guards. Reproduced on `-z` output, `git diff` content, and `worktree list --porcelain`.

pbean:
@codex review

pbean:
@coderabbitai full review

pbean:
Triaged on `0066682`, fixed in `28ae720` + `77b2df1`.

**@coderabbitai — your last review read stale code.** Its own review-info block says it covered *"between 87a4b5e and 2eb8f62"*, and it posted at 05:34:53Z, before `99ab745` (05:39:37Z) and `0066682` (05:44:27Z) existed. So all three findings were against the first commit:

- **Route through `_run_git`** — answered above and already in the code at `install.py:766-773`, as promised.
- **Fixed `.tmp` scratch name (nitpick)** — superseded by `99ab745`, which moved to `atomic_write_text`.
- **Stub git leaking into `/tmp`** — you marked this addressed yourself in `0066682`.

The two `full review` invocations both replied "Full review finished" while rate-limited, which is indistinguishable from a clean review. Flagging that because a stale review that reports success is worse than no review.

**@greptile-apps — the `prepend=False` note was right, and worth more than its severity.** `MonkeyPatch.setenv(name, value, prepend: str | None = None)` consumes it as `if prepend and name in os.environ`, so `False` is falsy and behaves exactly like omitting it. No behavioral effect, but the parameter takes a separator string, and anyone "correcting" it to `prepend=True` gets `value + True + os.environ[...]` → `TypeError`. Dropped, and the replaced comment records why PATH is *replaced* rather than prepended: the stub has to be the only resolvable `git`, or the box's real one answers first with a decodable path.

Pulling that thread found three more holes in the same test, all measured rather than reasoned:

1. **It passed vacuously whenever the stub could not exec.** The only assertion was `reason is None or "could not update" in reason` — which the git-unqueryable arm satisfies by returning `None`. On a noexec temp dir (ordinary CI hardening), with no `/bin/sh`, or on a filesystem that drops the exec bit, `subprocess.run` raises `OSError`, the first arm swallows it as the expected skip it is, and the test goes green having exercised nothing — taking the `text=True` ablation with it. Measured: at mode `0644` every assertion still passed. It now pins that the tail ran and the pattern landed on the decoded path.

2. **The leak check was aimed at the wrong place.** `Path("/tmp").glob("repo-*-name")` is non-recursive and `tmp_path` sits three levels down, so it could never see this test's own output — while it *could* fail for anyone whose box still carried litter from the hardcoded-path version. Replaced by the containment pin above, which catches a revert to a hardcoded path directly.

3. **The path was interpolated into the shell unquoted**, which reopened the `/tmp` leak through a different door. A temp root carrying whitespace (`TMPDIR`, `--basetemp`, a username with a space) word-splits into a second `printf` argument, and `printf '%s\n'` repeats its format per argument — so the helper is handed a path with an embedded newline and `mkdir(parents=True)`s it as a *sibling* of the basetemp pytest reaps. Measured with `--basetemp='.../my user'`: it created `my\nuser/...` outside the sandbox **and the test still passed**. Now single-quoted with the standard `'\''` escape.

All three ablations re-run against the final shape: `text=True` restored fails; a non-executable stub fails; the unquoted interpolation under a spaced `--basetemp` fails.

**One production finding, filed rather than fixed: #381.** `atomic_write_text` makes each *write* all-or-nothing; it does not serialize the read-modify-write around it. Every linked worktree of a repo resolves to the same `--git-common-dir`, so two runs provisioning against one repo interleave into a clean lost update — reproduced with two worktrees:

```
wt1 returned: None
/from-wt1 present: True
/from-wt2 present: False   <-- lost
```

Both calls returned `None`. That is exactly what the helper's docstring says silence must not cause ("without the exclude the unit's `git add -A` commits the tool files this provisioning just wrote"). Pre-existing — `write_text` lost the update the same way — and not covered by #379, which is scoped to truncation. `77b2df1` corrects the comment that invited the reading that concurrency was handled, along with two other overclaims: the mode is carried only when the exclude already exists (a new one gets the helper's deliberate `0600`), and the Windows note now reads as defensive placement rather than contradicting the test's POSIX-only skip.

pbean:
@coderabbitai full review

pbean:
@coderabbitai full review

pbean:
@codex — both correct. The first one I had already been handed and had triaged **wrong**; your framing is what makes it a defect rather than a docs nit.

**1. Creation permissions — real regression, fixed in `399d2de`.** An earlier review pass raised the `0600` change and I closed it as documentation-only, on the grounds that `atomic_write_text`'s docstring already declares the choice deliberate ("there is no prior mode to carry over, and the restrictive choice is the safe one"). That reasoning was about the generic helper, not this call site. You went to the consequence here, which is where it turns:

```
$ chmod 0000 .git/info/exclude && git status --porcelain
warning: unable to access '.git/info/exclude': Permission denied
?? tooldir/
$ git add -A && git diff --cached --name-only
tooldir/tool.txt          # exit 0
```

git does not fail on an unreadable exclude — it warns, exits 0, and stages the files the exclude existed to shield. So the narrowed mode reproduces the exact harm the helper's docstring says its silence must not cause, and reaches it through a write that *succeeded*, meaning not even the degrade-reason channel fires. The same docstring I had leaned on says the opposite of what I took from it two paragraphs on: *"Callers writing genuinely shared, multi-user state need more than this helper."* This caller is that.

Fixed by creating the file the way the replaced `write_text` did — `O_CREAT` under the umask — before handing it to the helper, which then carries that mode across the replace. End-to-end: mode `0644`, `git status` clean, no warning. An empty exclude is equivalent to an absent one for git, so a fault after the touch leaves no more damage than the missing file it replaced.

The regression test pins against a **measured** `write_text` probe rather than a hardcoded `0644`, and pins the umask at `0022` — at a box's own `0077` the ablated code also lands `0600` and the ablation would not bite. Verified: with the `touch()` removed it fails, reporting `0o600`.

**2. Sandbox assertion scoped to ambient state — already fixed in `28ae720`,** and your two failure directions are exactly right. It was worse than order-dependent: `Path("/tmp").glob(...)` is non-recursive while `tmp_path` sits three levels down, so it could never observe this test's own output in the first place, while it *could* fail for anyone whose box still carried litter from the hardcoded-path version of this test. Replaced with the positive containment assertion you describe — asserting the path attributable to the stub — which also catches a revert to a hardcoded path directly and depends on no ambient state.

That same commit closed two more in the same test: it passed **vacuously** whenever the stub could not exec (the only assertion was satisfied by the git-unqueryable arm's `None`, so a noexec temp dir made the test unfailable and its `text=True` ablation evaporate — measured at mode `0644`), and the stub's path was interpolated into the shell unquoted, so a temp root with whitespace word-split into a second `printf` argument and leaked a newline-named directory beside the basetemp pytest reaps, while still passing.

Ablations re-run against the final shape: `text=True` restored fails; a non-executable stub fails; the unquoted interpolation under a spaced `--basetemp` fails; the mode fix removed fails.

pbean:
@codex review

pbean:
@coderabbitai full review

pbean:
@codex review

Issue #377, opened 2026-07-29
_run_git's decode fault escapes the GitError taxonomy on -z output


pbean:
Found while sweeping for other copies of #374. This is the same genus but at the `_run_git` chokepoint, which makes it strictly more severe: the exception escapes **untyped**, defeating the very taxonomy that function exists to provide.

## The defect

`verify._run_git` decodes git's stdout with `text=True` (locale codec, `errors='strict'`) and translates only two exception classes:

```python
    except subprocess.TimeoutExpired as exc:
        raise GitError(...)
    except OSError as exc:
        raise GitSpawnError(...)
```

Its own docstring states the contract:

> Two failures are raised by `subprocess.run` *before* any return code exists — a timeout (#156) and a spawn-level OSError (#343) — so left uncaught either would bypass every `except GitError` guard and crash the run.

A `UnicodeDecodeError` is a **third** such failure, raised inside `subprocess.run` before any return code exists, and it is not in the taxonomy — so it does exactly what the docstring says must not happen.

## Why ordinary git commands are safe (and which are not)

Git quotes non-ASCII paths by default (`core.quotePath`), so `git status --porcelain` returns pure ASCII (`?? "weird-\377-name.txt"`) and decodes fine. **`-z` disables that quoting** and emits raw path bytes — which is exactly what `_git_raw` exists for:

> Like `_git` but returns stdout verbatim (no strip, no stderr merge) — for NUL-delimited (`-z`) output...

## Reproduction

A repo containing one file whose name is not valid UTF-8 (POSIX filenames are arbitrary bytes):

```
verify._git(repo, "status", "--porcelain")     -> rc=0 ok   out='?? "weird-\\377-name.txt"'
verify._git_raw(repo, "status", "--porcelain", "-z")
    -> RAISED UnicodeDecodeError -> 'utf-8' codec can't decode byte 0xff in position 9
       is GitError? False
```

## Affected call sites

All three `-z` invocations:

| line | call | reached from |
| --- | --- | --- |
| `verify.py:779` | `status --porcelain -z -uall` in `dirty_paths` | `clean_incoming_collisions` (merge pre-flight) |
| `verify.py:804` | `diff --name-only -z` in `branch_incoming_paths` | `clean_incoming_collisions` |
| `verify.py:2013` | `ls-files -z` in `commit_paths` | `decisions.py:161`, `cli.py:1859` |

[...]

Issue #360, opened 2026-07-28
flaky: test_decision_modal_scrolls_when_content_long fails intermittently under xdist load


pbean:
Filing an observation rather than a reproduction, so it is not silently absorbed. It was chased down at the time and could not be attributed to any change; it has not recurred since.

## What happened

`tests/test_tui_app.py::test_decision_modal_scrolls_when_content_long` failed **twice** during work on #355, both on full-suite `pytest -n auto` runs. It has not failed since, across a lot of runs.

| tree | runs | failures |
|---|---|---|
| pre-change (changes stashed) | 8, random seed | 0 |
| working tree, `--randomly-seed=1..6` fixed | 6 | 0 |
| committed tree | 8, random seed | 0 |
| after the remaining #355 work landed | ~4 full-suite `-n auto` | 0 |

So: 2 failures in ~22 runs, then 0 in roughly 26 consecutive. Fixed seeds never reproduce it, which rules out test ordering as the cause.

## Why it is probably load, not logic

It is a Textual `pilot` test — `await until(...)`, `scroll_visible(animate=False)` plus a single `pause()`, then an `_on_screen` assertion. That shape is sensitive to event-loop starvation: if the scroll has not been applied by the time the single `pause()` yields, the assertion reads a stale position. Both failures happened while the machine was concurrently running stash/pop and lint passes, i.e. under unusually heavy load on top of xdist.

Nothing in #355 touches decisions, the dashboard, or the modal (its only `tui/app.py` edit widens an `except` tuple in the *replan* handler), and the pre-change tree was probed 8 times without a failure — so it is pre-existing if it is real.

## Suggested fix if it recurs

Replace the single `pause()` with the same `await until(...)` predicate the test already uses elsewhere, so the assertion waits for the scroll position to settle instead of assuming one event-loop turn is enough. That removes the timing assumption without weakening what the test checks.

Origin of the test: PR #280 (issue #275, docked modal buttons).

Low priority — but if it starts reproducing at a higher rate, this is the record that it was seen before and was not caused by whatever change is in flight at the time.

pbean:
Another occurrence, for the frequency record: [run 30475807082](https://github.com/bmad-code-org/bmad-loop/actions/runs/30475807082), `test (py3.13)` on PR #388.

```
>           assert _on_screen(app, opt8)
E           AssertionError: assert False
tests/test_tui_app.py:1289
1 failed, 3686 passed, 35 skipped in 129.22s
```

[...]

PR #382, opened 2026-07-29
fix(verify): decode verify-command output with errors="replace" (#378)

### LINKED ISSUE #378: run_verify_commands decodes operator command output strictly; a non-UTF-8 byte crashes the run
Found while sweeping for copies of #374. Same genus as #377 but a **different guard and a different fix**, so filed separately: this call does not route through `_run_git`.

`verify.run_verify_commands` runs the operator's policy verify commands and decodes their output strictly:

```python
        try:
            proc = subprocess.run(  # nosec B602
                command,
                shell=True,
                cwd=cwd,
                capture_output=True,
                text=True,
                timeout=COMMAND_TIMEOUT_S,
            )
            output = (proc.stdout + proc.stderr)[-2000:]
            results.append(CommandResult(command, proc.returncode, output))
        except subprocess.TimeoutExpired:
            results.append(CommandResult(command, -1, "timed out"))
```

`except subprocess.TimeoutExpired` is the entire net. A `UnicodeDecodeError` raised while decoding the child's output escapes it.

## Why this is the most likely of the family to fire

The decoded bytes are arbitrary test-suite output, not a path — `core.quotePath` offers nothing here. Any non-UTF-8 byte anywhere in stdout/stderr does it: a latin-1 fixture filename appearing in a traceback, a `bytes` repr in an assertion diff, a compiler or linter diagnostic in a non-UTF-8 locale.

## Consequence

The exception escapes before any `results` entry is appended, so the whole list is lost — not just the offending command. It propagates `run_verify_commands` → `verify_commands_outcome` → `_dev_phase` → `engine.py`, and a deterministic verify gate is reported as a run crash with `UnicodeDecodeError`, pointing nowhere near the test suite that actually produced the byte. Re-running reproduces it forever, because the offending output is deterministic.

## Suggested fix

`errors="replace"` on this call. The output is captured purely to show a human/repair session what failed (`[-2000:]`, i.e. already lossy), so mojibake on one line is strictly better than losing the run — and unlike the `_run_git` sites, nothing here is fed back to a filesystem API, so surrogate round-tripping is not needed. `tui/launch.py:439` already uses `errors="replace"` for the same reason and is a good model.

Pre-existing; not introduced by #373 or #376.

pbean:
## Problem

`verify.run_verify_commands` runs the operator's policy verify commands through the host shell and captures their output with `text=True` — the locale codec at `errors="strict"`. `except subprocess.TimeoutExpired` is the entire net, so a `UnicodeDecodeError` raised while decoding a child's stdout/stderr escapes it.

The decoded bytes are arbitrary test-suite output, not a path: any non-UTF-8 byte does it — a latin-1 fixture filename in a traceback, a `bytes` repr in an assertion diff, a compiler or linter diagnostic in a non-UTF-8 locale.

## Consequence

The exception escapes *before* any `results` entry is appended, so the whole list is lost — not just the offending command. It propagates `run_verify_commands` → `verify_commands_outcome` → `_dev_phase` → engine, and a deterministic verify gate is reported as a run crash with `UnicodeDecodeError`, pointing nowhere near the test suite that actually produced the byte. Re-running reproduces it forever, because the offending output is deterministic.

## Fix

`errors="replace"` on the one call, plus the function's first docstring recording why.

Locale decode (`text=True`) deliberately stays, and `encoding="utf-8"` is deliberately **not** pinned the way `tui/launch.py` does: that child is bmad-loop's own UTF-8 CLI, whereas these children are arbitrary operator tools whose output follows the host locale (which, on every supported config, decodes UTF-8 anyway — POSIX PEP 538/540 coercion, Windows `PYTHONUTF8=1`). Surrogate round-tripping is not needed either: unlike the `_run_git` sites, nothing here is fed back to a filesystem API. The tail is display-only feedback and already lossy at `[-2000:]`, so a replacement character costs nothing while exit codes and the rest of the tail keep driving classification.

Sibling `_run_git` decode work is #377 — a different guard and a different fix, out of scope here.

## Tests

`tests/test_verify.py`, new `# ---- undecodable verify output (issue #378)` section. Both tests drive a **real child process** (`sys.executable` running a generated script that writes `b'before \xff after\n'`), never a monkeypatched `subprocess.run` — patching `run` hands the code `str` and never exercises the stdlib's decode at all, so such a test passes identically with the bug restored (the #374 regression test's docstring records that same finding).

- `test_verify_commands_undecodable_output_keeps_every_result` — drives `run_verify_commands` directly with `(bad-child rc=5, "exit 0")`; asserts the exit codes are `[5, 0]` (the offender's rc preserved **and** the later command's result not lost) and that the text on both sides of the bad byte survives. Deliberately asserts no `U+FFFD` literal — which glyph appears is locale-dependent, while the contract is "neither raises nor drops".
- `test_verify_commands_undecodable_failure_stays_fixable_retry` — through `verify_commands_outcome`; the run stays a fixable/retryable ordinary failure (not an env fault), with `rc=1` and the tail reaching the repair session's feedback, i.e. `env_fault_reason` ran over the replaced text without raising.

**Ablation performed:** removing `errors="replace"` fails both new tests with `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 7`, raised inside `subprocess._translate_newlines` — the real decode path. Restored by edit.

Full suite green (3698 passed), pyright 1.1.411 clean, full `trunk check` clean.

Closes #378

## Summary by CodeRabbit

* **Bug Fixes**
  * Verify-command results are no longer lost when command output contains undecodable bytes.
  * Undecodable output is safely displayed with replacement characters while exit codes and remaining output continue to determine verification status.
  * Failures with undecodable output are correctly classified as fixable and retryable rather than environment faults.
* **Tests**
  * Added regression coverage for undecodable output and result preservation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -260,6 +260,14 @@ story <id>`, the same annotation a sweep bundle writes. Both sprint and stories 
 
 ### Fixed
 
+- **One undecodable byte of verify output no longer crashes the run (#378).** Verify commands are
+  arbitrary operator tools, and their captured output was decoded strictly — a child emitting bytes
+  invalid in the run's encoding raised `UnicodeDecodeError` out of `run_verify_commands`, losing
+  every command's result and crashing the dev phase instead of classifying the failure. Output now
+  decodes with `errors="replace"`: the tail is display-only feedback for a human or repair session,
+  so a replacement character costs nothing, while exit codes and the rest of the tail keep driving
+  classification.
+
 - **A short write no longer truncates the worktree exclude (#375).** The exclude update is a
   read-modify-rewrite, and `write_text` truncates before writing, so a fault partway through (ENOSPC,
   EIO) left the operator's own excludes cut mid-content while the degrade reason still reported that
--- a/src/bmad_loop/verify.py
+++ b/src/bmad_loop/verify.py
@@ -1629,6 +1629,15 @@ def env_fault_reason(result: CommandResult, cwd: Path) -> str | None:
 
 
 def run_verify_commands(policy: Policy, cwd: Path) -> list[CommandResult]:
+    """Run each of the policy's verify commands, one CommandResult apiece.
+
+    Output decodes with ``errors="replace"`` (#378): the children are arbitrary
+    operator tools whose bytes are not ours to constrain, the captured tail is
+    display-only feedback for a human or a repair session (already lossy at
+    ``[-2000:]``), and one undecodable byte must not raise mid-loop and lose
+    *every* command's result. Decoding stays on the locale codec (``text=True``)
+    precisely because these are host tools — contrast tui/launch.py, which pins
+    ``encoding="utf-8"`` because its child is our own UTF-8 CLI."""
     results = []
     for command in policy.verify.commands:
         try:
@@ -1640,6 +1649,7 @@ def run_verify_commands(policy: Policy, cwd: Path) -> list[CommandResult]:
                 cwd=cwd,
                 capture_output=True,
                 text=True,
+                errors="replace",
                 timeout=COMMAND_TIMEOUT_S,
             )
             output = (proc.stdout + proc.stderr)[-2000:]
--- a/tests/test_verify.py
+++ b/tests/test_verify.py
@@ -1,4 +1,5 @@
 import dataclasses
+import io
 import os
 import subprocess
 import sys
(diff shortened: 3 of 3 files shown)


pbean:
@codex review

pbean:
@coderabbitai full review

pbean:
## Bot triage

| Bot | Channel | Outcome |
| --- | --- | --- |
| Codex | issue comments (it posts as a comment/review, not inline) | Clean on `95a3233a6b` |
| CodeRabbit | issue comments | Rate-limited both times — "Full review finished" arrived with "your included review limit is currently reached", i.e. no findings were produced. Re-reviewing against the new head. |
| Greptile | review body | Blocked: trial credit limit reached, no review possible on this PR |

Since two of three bots produced nothing, I ran an independent adversarial pass over the diff. It found one real hole in **my own tests**, now fixed in 022f6ef.

### The tests were vacuous under a non-UTF-8 single-byte locale

Byte `0xff` is undecodable only in UTF-8/ASCII. Every ISO-8859-x and cp125x codec maps all 256 byte values, so under such a locale the strict decode never raises — and both new tests passed **with the fix ablated**:

| | UTF-8 host | ISO-8859-15 host |
| --- | --- | --- |
| fix present | 2 passed | **2 passed** ← asserts nothing |
| fix ablated | 2 failed | **2 passed** ← silent vacuity |

Reproduced under `LC_ALL=et_EE.iso885915`. CI never saw it (Linux legs are UTF-8; Windows sets `PYTHONUTF8=1`), so the hole was developer-box-only — but it is exactly what the AGENTS.md ablation rule exists to catch, and nothing in the file said the tests were locale-conditional.

**Fix.** No single byte is undecodable in every codec, so re-picking the byte cannot work; the tests now skip when the host codec can decode `0xff`, and say so. The guard probes `TextIOWrapper(encoding=None)` rather than naming a codec, because that is the very default `subprocess`'s text mode resolves for the child's streams.

| | UTF-8 host | ISO-8859-15 host |
| --- | --- | --- |
| fix present | 2 passed | 2 skipped |
| fix ablated | 2 failed | 2 skipped |

No false-pass cell remains.

**Second ablation.** With the codec now pinned, the test asserts `U+FFFD` rather than only "did not raise". That earns its place: switching the source to `errors="ignore"` fails the new assertion while every pre-existing assertion still passes — so the weaker test could not tell replacement from silent byte-dropping.

Also dropped an overclaim: the second test said `env_fault_reason` runs over the replaced text, but on POSIX it returns at its `sys.platform != "win32"` guard without ever reading the tail. Only the Windows leg exercises that path.

### Pushback considered and declined

- **Pin `encoding="utf-8"`** — deliberate. These children are arbitrary operator tools whose output follows the host locale; `tui/launch.py` pins UTF-8 because its child is our own CLI. The docstring records the contrast.
- **Use `surrogateescape`** — nothing here round-trips to a filesystem API; the tail is display-only and already lossy at `[-2000:]`.
- **Also fix `_run_git` / `plugins/bus.py`** — separate typed-taxonomy work, tracked as #377 and #383 respectively.

### CI note

The earlier red `test (py3.14)` leg was `test_decision_modal_scrolls_when_content_long`, which is untouched by this diff and failed identically on **main** at 01:42 and 02:32 today, before this branch existed. Known flake, tracked as #360. It passed on re-run.

Local: full suite 3698 passed, `trunk check` clean, pyright scope unaffected (tests are outside `include`).

pbean:
@codex review

pbean:
@coderabbitai full review

Issue #384, opened 2026-07-29
provision_worktree's exclude patterns are repo-wide and permanent: new files under .claude/skills silently stop being staged in the main checkout


tjoignant:
`provision_worktree` shields a *transient* worktree by writing repo-wide, permanent exclude patterns. The patterns name directories that projects legitimately track (`.claude/skills`, `.claude/settings.json`), and nothing ever removes them — so long after the run, every **new** file under those paths is invisible to `git add -A` in the **main checkout**.

Observed on bmad-loop 0.9.0, git 2.30.0, macOS.

## What happens

`worktree_flow.py:286-293` builds the shield set from the CLI profiles:

```python
patterns = {f"/{p.skill_tree}" for p in profiles}                              # -> /.claude/skills
patterns |= {f"/{p.hooks.config_path}" for p in profiles if not p.hookless}    # -> /.claude/settings.json
patterns |= {f"/{rel}" for rel in seeded}
reason = _worktree_local_exclude(worktree, sorted(patterns))
```

`_worktree_local_exclude` (`install.py:734`) resolves its target with `git rev-parse --git-common-dir` (`install.py:787-793`). From a linked worktree that answers with the **main repo's** `.git`, so the patterns are appended to the repository-wide `.git/info/exclude`.

The defaults in `adapters/profile.py:50,62` are `.claude/settings.json` and `.claude/skills`, so a stock `claude` profile writes exactly:

```
/.claude/settings.json
/.claude/skills
```

Two properties combine badly:

1. **Scope** — the file is shared by the main checkout and every worktree, so a shield meant for one unit applies to the user's own working tree.
2. **Lifetime** — `_worktree_local_exclude` is the only writer and has no counterpart remover; `gc_run_worktrees` reclaims the worktree and leaves the lines behind. `isolation` back to `"none"` doesn't matter; the lines outlive the feature that wrote them.

The docstring argues the write is safe because it "does not affect already-tracked files." True, and that is precisely what makes it hard to notice: `.claude/skills/` is full of tracked files that keep diffing normally, while every *newly created sibling* silently disappears from `git status` and `git add -A`.

## Impact in my repo

`/.claude/skills` sat in `.git/info/exclude` across two BMAD-METHOD installer upgrades. Both upgrade commits captured only modifications to files that were already tracked; **51 new files were never staged**, including three whole skills (`bmad-review`, `bmad-editorial-review`, `bmad-deep-recon`).

[...]

Issue #236, opened 2026-07-21
safe_rollback matches git's English "did not match" — benign restore no-op raises GitError under a localized git


pbean:
### Description

Reported by @Polloinfilzato in #233 as a locale-dependent test failure — but the root cause is in
production code, and the blast radius is the rollback path, not the suite.

`safe_rollback` (`src/bmad_loop/verify.py`) restores each preserve dir from the pre-reset stash
snapshot and tolerates the benign failure — a preserve dir holding only untracked files — by
substring-matching git's **English** message:

```python
rc, out = _git(repo, "checkout", snapshot, "--", d)
if rc != 0 and "did not match" not in out:
    raise GitError(...)
```

`_run_git` spawns git with the inherited environment and no locale override, so under a localized
git (e.g. `LANG=it_IT.UTF-8`) the benign message is translated, the substring never matches, and
the benign no-op **raises `GitError`** — a resolved re-drive that should auto-recover instead
lands in the rollback failure path.

### Steps to reproduce

- `LC_ALL= LANG=it_IT.UTF-8 .venv/bin/pytest tests/test_engine.py::test_rollback_or_pause_resolved_auto_recovers`
  fails (matches the reporter's observation: fails identically on clean main, passes under
  `LC_ALL=C`).

### Fix shape

Force the C locale at the single spawn point, `_run_git`, so every git message is stable English:

- Build the child env as `{**(env or os.environ), "LC_ALL": "C"}` — note the merge must preserve
  the explicit-env path (`_git_env` passes a custom env for the throwaway `GIT_INDEX_FILE`
  snapshot staging; clobbering it would break `snapshot_worktree`).
- That also future-proofs any other message matching; today `"did not match"` appears to be the
  only message-content match on a git result, but the chokepoint fix makes the class impossible
  rather than patching the instance.
- The tolerance behaviour is pinned by `tests/test_verify.py` ("benign 'pathspec did not match'
  case is tolerated"); a regression test could run the benign case with a localized `LANG` in the
  parent env and assert it still doesn't raise.

Refs #233 (where it was reported).

PR #387, opened 2026-07-29
fix(verify): translate git's decode fault and add a bytes mode (#377)

### LINKED ISSUE #377: _run_git's decode fault escapes the GitError taxonomy on -z output

pbean:
## What

`verify._run_git` now translates a strict-decode fault into `GitError`, and gained a bytes mode
(`binary=`, plus the public `git_bytes` accessor).

## Why

Fixes #377

The chokepoint exists to give git's pre-returncode faults a type its ~50 `except GitError` guards
can catch. It translated a timeout (#156) and a spawn `OSError` (#343) — but `UnicodeDecodeError` is
a third such fault, and being a `ValueError` it matched neither arm, so it escaped **untyped**: the
one thing its own docstring says must not happen. One POSIX filename that is not valid in the run's
encoding is enough, wherever git's own quoting is off — the `-z` callers (merge pre-flight),
`worktree list --porcelain` (crashes the *unguarded* stale-run reconcile at every run/sweep start),
and `git diff`, whose content is verbatim (a failed unit's forensic capture bypasses the valve that
preserves its uncaptured work).

The bytes mode is the second half #377 asks for: `install._shield_git` stands outside the chokepoint
AGENTS.md requires because it needs raw bytes *and* needs a non-zero rc as an **answer** (`config
--get` of an unset key exits 1), and its docstring names this issue as the blocker.

## How

- `except UnicodeDecodeError` → `GitError`, verbatim as the issue sanctions. Not
  `errors="surrogateescape"` — #377 defers that, since surrogates would flow into UTF-8 journal/JSON
  writes.
- Keyword-only `binary: bool = False` with `@overload`s so the return type is
  `CompletedProcess[bytes]` when set; the only body change is `text=not binary`. Bytes callers keep
  the `LC_ALL=C` pin (#236) and the `_git_timeout_s` bound (#156), and the two faults with no rc to
  return still raise.
- Public `git_bytes(repo, *args)` so no other module reaches for the private name. Argv shape is
  unchanged — `_run_git`'s messages index `cmd[3]`.

**No caller on `main` yet:** #385/#386 are still open, so `install._shield_git` does not exist here.
This is the API half; the rewiring stays with #384's stack.

## Testing

5 tests in `tests/test_verify.py`, all on the conftest `project` fixture. The decode test uses a
**real** bad-byte filename and real git — monkeypatching `subprocess.run` hands the code `str` and
never runs the stdlib's decoding, so that version passes with the fix ablated
(`tests/test_install.py:1996` documents the same trap). It reuses the `needs_strict_codec` guard
from #378 (hoisted to the top of the file, since it now serves both), because 0xff decodes fine
under ISO-8859-x/cp125x.

Full suite `3703 passed, 24 skipped` (main: 3698 + 24, i.e. exactly the 5 new tests);
`pyright@1.1.411` 0 errors; `trunk fmt` + full `trunk check` clean.

### Ablations run (AGENTS.md rule)

| Deleted | Result |
| --- | --- |
| `except UnicodeDecodeError` arm | `test_z_output_undecodable_path_becomes_git_error` FAILS with raw `UnicodeDecodeError` ✅ |
| `binary=True` in `git_bytes` | `..._returns_bytes_and_reads_rc_as_an_answer` + `..._inherits_locale_pin_and_timeout` FAIL ✅ |
| `except subprocess.TimeoutExpired` arm | `test_git_bytes_timeout_still_becomes_git_error` FAILS with raw `TimeoutExpired` ✅ |
| `except OSError` arm | `test_git_bytes_spawn_oserror_still_becomes_git_spawn_error` FAILS with raw `OSError` ✅ |

Plus the vacuity check the #378 follow-up exists for: under `LC_ALL=et_EE.iso885915` the decode test
**skips** with its stated reason rather than falsely passing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Git operations no longer crash when encountering filenames with non-UTF-8 characters.
  * Undecodable output is now reported as a clear Git error.
  * Git timeouts and startup failures are consistently surfaced as application errors.

* **Improvements**
  * Added support for safely processing raw Git output, improving handling of unusual filenames.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -260,6 +260,18 @@ story <id>`, the same annotation a sweep bundle writes. Both sprint and stories 
 
 ### Fixed
 
+- **A non-UTF-8 filename no longer crashes the run past every git guard (#377).** `verify._run_git`
+  is the sole git spawn point and exists to give git's faults a type its ~50 `except GitError`
+  guards can catch, but it decoded git's output strictly and translated only a timeout and a spawn
+  `OSError`. `UnicodeDecodeError` is a third fault raised before any return code exists, and being a
+  `ValueError` it matched neither arm — so one file whose name is not valid in the run's encoding
+  (POSIX filenames are arbitrary bytes) escaped untyped. Reachable wherever git's own quoting is
+  off: the `-z` callers (merge pre-flight collision detection), `worktree list --porcelain` — which
+  crashed the unguarded stale-run reconcile at every run and sweep start — and `git diff`, whose
+  content is verbatim, so a failed unit's forensic capture bypassed the valve that preserves its
+  uncaptured work. Such output is now a `GitError` like any other git failure. Making those paths
+  usable rather than merely non-fatal is a separate change.
+
 - **One undecodable byte of verify output no longer crashes the run (#378).** Verify commands are
   arbitrary operator tools, and their captured output was decoded strictly — a child emitting bytes
   invalid in the run's encoding raised `UnicodeDecodeError` out of `run_verify_commands`, losing
--- a/src/bmad_loop/verify.py
+++ b/src/bmad_loop/verify.py
@@ -87,13 +87,23 @@ class GitSpawnError(GitError):
 def _run_git(
     cmd: list[str], repo: Path, *, env: dict[str, str] | None = None
 ) -> subprocess.CompletedProcess[str]:
-    """Sole spawn point for git subprocesses. Two failures are raised by
-    `subprocess.run` *before* any return code exists — a timeout (#156) and a
-    spawn-level OSError (#343) — so left uncaught either would bypass every
-    `except GitError` guard and crash the run. Both are translated here into
-    the GitError taxonomy — observation guards degrade, unguarded paths fail
-    typed — with the spawn class marked as `GitSpawnError` for the callers
-    that must distinguish an environment fault from git refusing.
+    """Sole spawn point for git subprocesses. Three failures are raised by
+    `subprocess.run` *before* any return code exists — a timeout (#156), a
+    spawn-level OSError (#343), and a strict-decode fault on the child's output
+    (#377) — so left uncaught any of them would bypass every `except GitError`
+    guard and crash the run. All are translated here into the GitError taxonomy
+    — observation guards degrade, unguarded paths fail typed — with the spawn
+    class marked as `GitSpawnError` for the callers that must distinguish an
+    environment fault from git refusing.
+
+    The decode fault is real, not theoretical: POSIX filenames are arbitrary
+    bytes, and while `core.quotePath` C-quotes them to ASCII for ordinary
+    porcelain, `-z` disables that quoting (`dirty_paths`, `branch_incoming_paths`,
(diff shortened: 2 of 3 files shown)


pbean:
@codex review

PR #388, opened 2026-07-29
chore: pin pyright in the dev group and run it via uv

pbean:
## What

Pins `pyright==1.1.411` in the `dev` dependency group and switches the typecheck job from `uvx
pyright@1.1.411` to `uv run pyright`.

## Why

`uv run pyright` — the obvious local command, and the one AGENTS.md's "pyright at the CI-pinned
version" implies — **did not work**: pyright was never a dependency, so it failed with `Failed to
spawn: pyright`. The only way to typecheck locally was to read the workflow file and retype its
literal. Found while working #377, where that failure looked like a broken environment rather than a
missing dep.

Worse, the version lived in exactly one place — a string in `ci.yml` — with nothing tying it to the
project. Any contributor invoking a bare `pyright` got whatever was on their machine, silently
checking against a different version than the gate.

## How

- `pyright==1.1.411` in the `dev` group. **Exact, not a floor** like the other dev deps: pyright is
  a gate, and a newer one adds checks that fail the build on untouched code. 1.1.411 is both what
  the job already pinned and the current PyPI release, so this pins the status quo, not a bump.
- The job runs `uv run pyright`, resolving that pin from `uv.lock`. One pin, in the lock, used by
  both — no sync test needed, because there is no second copy to drift from.
- AGENTS.md spells the command out in the dev-environment block instead of pointing at the workflow.

Runtime path is unchanged: `uvx pyright@1.1.411` was already fetching this same PyPI package, whose
wheel bundles the pyright dist (`PYRIGHT_PYTHON_USE_BUNDLED_PYRIGHT` defaults on), so no npm or node
download is introduced.

## Testing

- `uv sync --locked --all-extras` clean (what CI runs — fails on a stale lock).
- `uv run pyright` → `0 errors`, and `uv run pyright --version` → `pyright 1.1.411`, matching the
  version the job pinned before.
- **Proved the gate still bites, rather than trusting a green run:** injected
  `def _gate_probe(x: int) -> str: return x` into `model.py` (a strict-list module) and `uv run
  pyright` reported `2 errors` (`reportReturnType`, `reportUnusedFunction`); back to 0 once
  reverted.
- Full suite `3698 passed, 24 skipped`; `scripts/sync_version.py --check` ok; `trunk fmt` + full
  `trunk check` clean.

No CHANGELOG entry: dependency-groups are not installed by consumers, so nothing about the shipped
package changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,8 +132,11 @@ jobs:
       - name: Install the project
         # Populates .venv (venvPath/venv in the pyright config) so third-party and
         # optional-extra imports (textual, pyte, httpx, …) resolve instead of being
-        # flagged as missing.
+        # flagged as missing. Also installs pyright itself, which is pinned in the
+        # `dev` dependency group rather than named here — one pin, in the lock.
         run: uv sync --locked --all-extras
 
       - name: Pyright (basic mode)
-        run: uvx pyright@1.1.411
+        # `uv run`, not `uvx pyright@<version>`: the same command a contributor
+        # runs locally, resolving the same pinned version from uv.lock.
+        run: uv run pyright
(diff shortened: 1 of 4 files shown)


pbean:
@codex review

Issue #405, opened 2026-07-30
upstream bmad-build-auto rename (bmad-method >= 6.10.1-next.33) breaks validate/run/sweep


pbean:
Upstream BMAD-METHOD PR #2651 (first shipped in `bmad-method@6.10.1-next.33`, 2026-07-30)
renamed the dev-primitive skill `bmad-dev-auto` → `bmad-build-auto`, leaving a forwarding
shim named `bmad-dev-auto` (a lone SKILL.md with an interactive customization-migration
gate that can HALT unattended sessions with zero disk writes). On an upgraded target
project, `bmad-loop validate` fails `skills.base-incomplete` (the shim lacks the
`step-04-review.md` + `customize.toml` markers), the same preflight hard-blocks
`run`/`sweep`/`resume`, and worktree isolation would never copy `bmad-build-auto`.

The rename itself is contract-clean (spec `status:` flip, `## Auto Run Result`,
"folder+id dispatch" marker, terminal statuses, Stop-event completion all unchanged; only
the no-spec fallback result filename changed to `bmad-build-auto-result-*`). Two other
upstream changes rode the same window and also break bmad-loop:

- **PR #2601**: SKILL.md is now a renderer stub running `_bmad/scripts/render_skill.py`
  via uv — HALTs with no spec write when the script is missing. Step files and
  customize.toml still exist in the skill dir (markers remain valid).
- **PR #2640**: deferred findings now land in the spec's frontmatter `deferred:` YAML
  list, not `deferred-work.md` — the sweep pipeline silently starves.

Everything on main since the v0.9.0 tag is next-major work and must not ship, so the fix
is cut as hotfix **0.9.1** from tag `v0.9.0` (commit `7255174`) on a `release/0.9.x`
branch. Upstream npm `latest` (6.10.0) is still pre-rename, so 0.9.1 must keep working
against **both** skill eras.

## Scope of the fix

1. **Resolve the primitive on disk** — prefer `bmad-build-auto`, fall back to a
   marker-complete `bmad-dev-auto`; never invoke or accept the shim. New validate checks
   `skills.base-shim`, `skills.dev-renderer`, `skills.customize-legacy`.
2. **Thread the resolved name** through every session prompt (dev, review, repair,
   stories dispatch, sweep bundles) and the no-spec fallback result marker.
3. **Harvest spec-frontmatter `deferred:` findings** into the deferred-work ledger,
   era-agnostically (gated on content presence, never on the installed skill name).

[...]

PR #430, opened 2026-08-02
chore(ci): run CI and auto-publish on release/* maintenance branches

pbean:
## What

Add `release/*` to the `push` branch filter in both `.github/workflows/ci.yml`
and `.github/workflows/release.yml`, and correct `release.yml`'s header comment,
which described the trigger as a version bump landing on `main` only.

Nothing else changes.

## Why

Both workflows fire only on `push: branches: [main]`. A maintenance branch
therefore gets **no checks at all** on its merge commits, and a version bump
landing there never publishes — `release.py publish` simply has no trigger
outside trunk.

The 0.9.1 hotfix hit exactly that. It was cut from `v0.9.0` onto
`release/0.9.x` (main carries next-major work that must not ship in a patch),
merged as #406 with zero checks, and had no path to a `v0.9.1` tag short of a
hand-created one.

## Effect

**Any `release/*` branch will now auto-publish on a version bump**, the same
way `main` does: pushing a merge commit that bumps the canonical version fires
the publish job, which creates the `vX.Y.Z` tag and GitHub release from the
matching `## [X.Y.Z]` CHANGELOG section, targeting that branch's head. Those
branches also get full CI on their merge commits for the first time.

`release.py publish` is idempotent — it no-ops when the tag already exists — so
the two triggers cannot fight: whichever branch carries a given version first
wins the tag, and the other side's run does nothing. That is what keeps a
forward-port of already-released content from re-tagging it.

The same change was shipped to `release/0.9.x` as #429 so v0.9.1 could publish;
this PR carries it to trunk.

Refs #405.

## Summary by CodeRabbit

* **Chores**
  * Continuous integration now runs for updates to both the main branch and release branches.
  * Releases can be initiated from either the main branch or a release branch.
  * Updated release workflow documentation to reflect the supported branch options.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: [main]
+    branches: [main, release/*]
   pull_request:
 
 # Cancel superseded runs for the same ref.
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,13 @@ on:
     branches: [main, release/*]
 
 concurrency:
-  group: release-${{ github.ref }}
+  # Deliberately ref-independent: `main` and a `release/*` branch can carry the
+  # same version, and keying on `github.ref` would put them in separate groups and
+  # let both publish runs proceed at once. One group repo-wide serializes them, so
+  # the second run checks out *after* the first pushed the tag and its `tag_exists`
+  # probe short-circuits cleanly. Never cancel-in-progress — a cancelled publish
+  # leaves the tag unmade.
+  group: release-publish
   cancel-in-progress: false
 
 permissions:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,12 +1,13 @@
 name: Release
 
-# Auto-publish on merge: when a version bump lands on `main`, create the matching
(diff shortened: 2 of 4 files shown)


pbean:
@coderabbitai Verified the race and it is real, but it is out of scope for this PR — filed
as #431 instead.

Confirming the mechanism rather than taking it at face value: `tag_exists` is
`git rev-parse -q --verify refs/tags/<tag>` against the **local** clone, and nothing
fetches between `actions/checkout` and the check. So the guard reflects tag state at
checkout time, not the live remote — two runs that both check out before either creates the
tag both pass it, both call `gh release create`, and the loser exits non-zero under
`check=True`. And `group: release-${{ github.ref }}` does not serialize them, since two
branches are two refs. Your read was correct on both halves.

Not fixing it here for two reasons:

1. This PR is deliberately the trigger change and nothing else. Its sibling #429 already
   merged into `release/0.9.x` and published v0.9.1; letting the two diverge would muddy the
   forward-port.
2. The blast radius is a spuriously red workflow run on the losing branch — the release that
   wins is still correct. It is not a correctness bug in the published artifact.

One note on the suggested fix: a shared concurrency group alone does not close the window,
because the second job's checkout can still predate the first job's tag push. #431 records
that, and lands on the `gh release create` conflict-tolerance half as the one that actually
closes it.

Issue #433, opened 2026-08-02
forward-port the 0.9.1 hotfix (#405) to main


pbean:
Forward-port the 0.9.1 hotfix to `main` as a sequenced program of sub-phases.

PR #406 merged into `release/0.9.x` on 2026-08-02 and shipped 57 commits (`+~2,000` net
source lines across 18 files, `+12,076` test lines) after ~20 adversarial bot-review rounds.
`main` is 274 commits ahead of the v0.9.0 tag the hotfix was cut from, so this is a
**re-implementation, not a cherry-pick**: `worktree_flow.py` owns provisioning,
`resolve_review_layers`/`_CUSTOMIZE_OVERRIDES` exist and must follow the resolved name,
worktree excludes are worktree-scoped, and pyright is a configured CI gate that has never run
on any of this code.

`bmad-build-auto` has **zero occurrences** on `main` today — nothing pre-landed.

## Sub-phases

Fifteen sessions, fourteen PRs. Every sub-phase leaves `main` green and coherent.
CONTRIBUTING's one-concern-per-PR rule was waived *for the hotfix only*; it applies here.

| # | Sub-phase | PR | Closes |
| --- | --- | --- | --- |
| 6A | resolution seam (+ tracking issue) | 1 | — |
| 6B | prompt threading + dry-run banner | 1 | **#393** |
| 6C | small independent contracts + pyright canary | 2 | — |
| 6D | verify pathspec hardening | 3 | **#423** |
| 6E | #414 refusal | 4 | supersedes #414 |
| 6F | renderer preflight | 5 | resolves **#410** |
| 6G | provisioning totality | 6 | **#421**, **#422** |
| 6H | `_bmad` seeding + escalations | 7 | — |
| 6I | deferral parser + ledger primitives + model fields | 8 | — |
| 6J | harvest + proof-of-work exclusion | 9 | — |
| 6K | snapshot / restore / disarm + ledger classification | 10 | **#420** |
| 6L | accept-gated bundle close + reopen | 11 | — |
| 6M | isolation carry + crash replay | 12 | — |
| 6N | #425 + #426 + residuals | 13 | **#425**, **#426** |
| 6O | CHANGELOG + program close-out | 14 | — |

6C, 6D, 6E are order-independent among themselves. 6F→6G→6H is a chain; 6I→6J→6K→6L→6M is a
chain; the two chains are independent after 6E. Ship-worthy release points: after **PR 1**
[...]

PR #441, opened 2026-08-03
fix(verify): force literal git pathspecs, and read status stdout alone (#433, 6D)

### LINKED ISSUE #423: verify.py interpolates paths into git pathspecs unescaped: _exclude_specs over-excludes, disagrees with _path_under_any, and safe_rollback's preserve checkout over-restores
`verify.py` builds git pathspecs by interpolating configured paths, so a path carrying `*`, `?` or `[…]` matches a **wider** set than the one it names. Every affected call is on the over-match side; none of them under-matches.

Sibling of #401, which is the same class one layer over: there the interpolation is into **gitignore** syntax, here it is into **git pathspec** syntax. Neither escapes.

## Status

**Items 1 and 2 shipped in #406** (`hotfix/0.9.1-build-auto`, commits `f0b34b3` + `e8979b6`), via a `_literal_specs` helper mirroring `_exclude_specs`. **Items 3, 4 and 5 remain open and are what this issue now tracks.**

Two corrections to this issue's own text, both established by measurement while fixing 1 and 2 — kept rather than deleted, because each was load-bearing in the decision to defer:

- **The deferral rationale below was wrong.** It read *"neither a regression it introduced nor reachable on any configuration it ships."* Both halves fail for those two sites. `git log -S "def path_tracked"` returns `e370528` alone and `git merge-base --is-ancestor e370528 release/0.9.x` is false — `path_tracked` and its only exposed call site are entirely new on that branch. And `commit_paths` became **unattended** in `dc4f934`: until then its only caller was `decisions.py:155`, behind the operator-invoked `bmad-loop decisions`; `_carry_harvested_deferrals` put it inside `_defer`, on the run path, with no human present.
- **Item 2's "already declares safe" claim named the wrong lines.** `verify.py:325-327` declares the **GitError degrade** safe, not the over-match. The real gap was `:310-316`, whose enumeration of considered false-positive sources (`core.quotePath`, tracked-but-deleted, `--error-unmatch`, `check-ignore`) omitted pathspec globbing entirely. The fix carries a new paragraph there.

### Still live on `main`

`main` is **not** already fixed, and the forward-port is not a clean cherry-pick:

- `path_tracked` does not exist on `main` at all — item 2 is unique to the hotfix branch.
- `main`'s `commit_paths` still has all three bare operands, **plus a fourth unescaped `ls-files -z -- *missing` pathspec** that `release/0.9.x` does not carry. Its failure direction looks different — the membership test is `r in tracked`, and an over-match returns the *neighbour's* name — so it likely **drops** the rel rather than over-including. Not measured; measure before porting.
- `main` hand-rolls `r.replace("\\", "/")` where `.as_posix()` fixes the same Windows-separator problem at the source. The port should replace it, not keep both.

### Version floor (was a blocking unknown)

Measured, not read off the docs, as this issue required. `:(literal)` behaves identically at **git 2.20.4** (alpine:3.9 — the floor `install._WORKTREE_CONFIG_GIT` declares) and at 2.55.0, non-root: metachar operand → empty; ordinary path → lists; directory prefix `_bmad/render` → lists beneath; tracked-but-deleted → still lists; `add` of an ignored path → rc 1. This mattered more than it looked: below the floor the prefix would read as a literal *filename*, match nothing and answer `False` — which **authorizes the unlink**, inverting the failure direction from "skip a delete" to "delete a tracked file".

---

## Measured (git 2.55.0)

Fixture: sibling pairs `docs[a]`/`docsa`, `a*b`/`axb`, `q?r`/`qxr`, each holding `f.md`.

**1. `commit_paths` — FIXED on `release/0.9.x`, still live on `main`.** `verify.py:1838-1866` documents itself as *"commit exactly `paths` (and nothing else) … safe to call out of band"*, and its `add`/`status`/`commit` all took the rels as bare pathspecs. It was not:

```
add -- docs[a]/f.md   staged=['docs[a]/f.md', 'docsa/f.md']
add -- a*b/f.md       staged=['a*b/f.md',     'axb/f.md']
add -- q?r/f.md       staged=['q?r/f.md',     'qxr/f.md']
```

An unrelated user file is staged and committed under a story's name. Note the operand is a **superset** — it matches its own literal text *and* the glob's reach — because `match_pathspec_item` compares literally before falling through to fnmatch. That is also why the under-match direction is impossible.

A second harm this issue did not originally record: with the ledger **gitignored** (the shape `_carry_harvested_deferrals` is built to hit), `git add` refuses an explicitly-named ignored path but *skips* a globbed one. So the plain form could exit **rc 0** having staged nothing, `status` found no change, and the carry reported success having committed no ledger and journalled no `harvest-carry-uncommitted` — a silent loss where the explicit form has a record. rc 0 requires a glob-reachable neighbour to exist: with none it is still rc 1; with a *clean* neighbour it stages nothing; with a *dirty* one it stages the operator's edit.

**2. `path_tracked` (`verify.py:328`) over-matched — FIXED on `release/0.9.x`; does not exist on `main`.** With the metachar path *untracked* and only the collider tracked:

```
rel='docs[a]/f.md'  tracked? plain=True  literal=False
rel='a*b/f.md'      tracked? plain=True  literal=False
rel='q?r/f.md'      tracked? plain=True  literal=False
```

So `_ledger_is_gits_to_restore` answered `True` for an untracked ledger and the harvest revert's `unlink()` was skipped — the ledger revert silently no-opped for such a project, leaving a finding about discarded code open for a later sweep. Found by Codex on #406 (`dc4f934`); reproduction and ablation table on that thread.

**3. `_exclude_specs` (`verify.py:280-282`) over-excludes, but only for specific shapes** — worth stating precisely, because the general form of this claim is *false*:

```
prefix='docs[a]'  victim='docsa/f.md'   victim-excluded=False   <- sibling DIR: NOT swallowed
prefix='docs[a]'  victim='docsa'        victim-excluded=True    <- sibling FILE at root: swallowed
prefix='q?r'      victim='qxr/f.md'     victim-excluded=False
prefix='q?r'      victim='qxr'          victim-excluded=True
prefix='a*b'      victim='axb/f.md'     victim-excluded=False
prefix='doc*'     victim='docsa/f.md'   victim-excluded=True    <- a `*` that can swallow `/`
```

The pathspec has to match the **whole** path, so a same-length `[…]`/`?` collision only reaches a sibling *file*, never a sibling directory's contents; a `*` reaches everything because it crosses `/`. Where it does fire, `has_changes_since` (`:208`, the `diff --quiet` at `:237`) and `attempt_dirty` (`:247`, the same call at `:270`) under-report the diff — a false "no changes", the same failure #406 fixed at `_harvest_gate_exclude` by a different route.

**4. `_exclude_specs` (git glob) and `_path_under_any` (`:285-287`, a Python `startswith`) disagree** on exactly those rows — the two halves of one `has_changes_since` call answer differently:

```
prefix='docs[a]'  path='docsa'      git-excludes=True   python-under=False
prefix='q?r'      path='qxr'        git-excludes=True   python-under=False
prefix='doc*'     path='docsa/f.md' git-excludes=True   python-under=False
```

The tracked half is excluded while the untracked half is still counted, so the two branches of the same function disagree about what "under the artifact dir" means.

**5. `safe_rollback`'s `checkout -- <preserve dir>` (`verify.py:648`)** — same shape-dependence:

```
checkout -- 'docs[a]'   target=restored  victim='dirty'    (unaffected)
checkout -- 'doc*'      target=restored  victim=restored   (an unrelated file reverted)
```

## Reachability

`implementation_artifacts` is read verbatim out of `_bmad/bmm/config.yaml` and only `{project-root}`-substituted and resolved (`bmadconfig.py:131-146`); nothing sanitizes it. `*` and `?` are illegal in Windows path names, so the realistic carrier is `[`/`]` — which narrows the live exposure to the `victim`-is-a-file rows of 3/4. No fixture in the repo has ever carried one, and no report exists; the remaining items are latent-correctness, not a live bug.

## Fix shape (remaining items)

`:(exclude,literal)` in `_exclude_specs` (item 3) and `:(literal)` on `safe_rollback`'s preserve dirs (item 5). The floor is settled — `:(literal)` is measured safe at 2.20.4, see above — so that blocker no longer applies.

Item 4 is a **behaviour decision**, not an escaping fix: it is about what "under the artifact dir" means. Fixing only the git half does not merely paper over the disagreement — it makes both halves agree *on the literal reading*, which is the one `_path_under_any` already implements, so it is the right direction rather than a coin flip.

pbean:
Sub-phase **6D** of the #433 forward-port. Independent of 6C (merged as #439); branched from `main`.

`implementation_artifacts` is read verbatim out of the operator's `_bmad/bmm/config.yaml`
(`bmadconfig._resolve` only `{project-root}`-substitutes and resolves it — nothing sanitizes it),
and git reads a positional operand as a **pathspec**, not a path. So a configured name carrying
`[`, `]`, `*` or `?` matched a wider set than the one it named. Every affected call is on the
over-match side; none under-matches, because `match_pathspec_item` compares literally before
falling through to fnmatch.

## What changed

- **`commit_paths`** — all four operands (`ls-files -z`, `add`, `status`, `commit`) literalised, so
  it keeps its own first promise, "commit exactly `paths` (and nothing else)", instead of staging a
  glob-colliding sibling's edit under a story's name. A second, quieter harm with the ledger
  **gitignored** (the shape `_carry_harvested_deferrals` hits): `git add` refuses an explicitly
  named ignored path but *skips* a globbed one, so the bare form could exit rc 0 having staged
  nothing and report success having committed no ledger.
- **`.as_posix()`** replaces `str()` when building the rels, which makes the hand-rolled
  `r.replace("\\", "/")` at the membership test dead. Deleted in the same change — a dead
  normalizer beside a glob operand reads as protection that isn't there.
- **`_exclude_specs`** — now `:(exclude,literal)`. An artifacts dir whose name carries `*` was
  excluding the sibling tree it glob-matches, so `has_changes_since`/`attempt_dirty` reported a
  changed dev attempt as CLEAN. This also realigns the git half with `_path_under_any`, the Python
  `startswith` filtering the untracked half of the *same* `has_changes_since` answer — the two
  disagreed on exactly the shapes that glob (#423 items 3–4). Literal is the reading
  `_path_under_any` already implements, so agreeing on it is a fix, not a coin flip.
- **`safe_rollback`** — the preserve `checkout` is a **write**, so a glob reaching a sibling hands
  back a change the reset had just discarded (item 5).
- **`worktree_clean`** — rewritten onto `_run_git`, reading **stdout alone**. See below.
- **New `_literal_specs` and `path_tracked`**, ported from `release/0.9.x`.

## `worktree_clean` — a second bug found while porting

`_git` returns `(proc.stdout + proc.stderr).strip()`. `status` exits 0 while still writing to
stderr — an unexecutable `core.fsmonitor` hook, an unknown `core.fsyncMethod` — and against the
merged stream that chatter is indistinguishable from a porcelain record, so a **pristine tree
answers DIRTY**. Seven callers gate on this and three of them (`cli.py:268`, `:1023`, `:1261`)
refuse the command outright, so a host with a noisy git config could never start a run and the
message named no file. `path_tracked` carries the same stdout-alone rule for the same reason; the
error path keeps the merge, where stderr is the only informative half.

## `path_tracked` ships with zero callers

Deliberate, and constraint 4 of the program's hard ordering: `git.render-tracked` (**6F**) and
`_ledger_is_gits_to_restore` (**6K**) are its consumers, and both need it to exist first. That is
*inert*, not silent — neither ruff's `E4,E7,E9,F` set nor pyright basic reports an unused
module-level function.

## Version floor

`:(literal)` **and** the combined `:(exclude,literal)` are measured identical at **git 2.20.4**
(the floor `install._WORKTREE_CONFIG_GIT` declares, via `alpine:3.9`) and at **2.55.0**. #423 had
measured only the plain `:(literal)` form; the combined one this PR introduces was unmeasured, so
that gap is now closed. This matters more than it looks: below a git that understands the prefix it
reads as a literal *filename*, matches nothing and answers `False` — which **authorizes an unlink**,
inverting the failure direction from "skip a delete" to "delete a tracked file". The floor must not
drop.

## Verification

`pytest` 3953 collected, green apart from 4 pre-existing `test_module_skills_sync` failures that are
local dev-box skill drift (identical on a clean `main` checkout). `pyright` 0/0/0 — unchanged from
the `origin/main` baseline. `trunk check --all --no-fix`: 242 files, no issues.

**Ablation matrix, run on both the 3.13 and 3.14 lanes** (whole suite each row; every row's total
3953, so no baseline drift, and the src digest was verified against the known-clean value before
each row). Both lanes returned **identical** results:

| Row | Ablation | New reds |
| --- | --- | --- |
| R1 | `:(exclude,literal)` → `:(exclude)` | 2 |
| R2 | `path_tracked` literal operand → bare | 4 |
| R3 | `path_tracked` stdout → stdout+stderr | 1 |
| R4 | `worktree_clean` stdout → stdout+stderr | 1 |
| R5 | `commit_paths` add/status/commit literal → bare | 1 |
| R6 | `commit_paths` `ls-files` literal → bare | **0** |
| R7 | `safe_rollback` preserve literal → bare | 1 |
| R8 | `.as_posix()` → `str()` | **0** |

The two zero-red rows are reported rather than papered over, since an edit no ablation reddens and
a test no ablation reddens are the same signal:

- **R8 is unpinnable on this platform, not unpinned.** `str(PosixPath("a/b"))` and `.as_posix()` are
  byte-identical on POSIX; the change only bites where `str()` yields backslashes. **Windows CI is
  its sole oracle** — and both Windows jobs run this suite.
- **R6 is redundant-by-design.** #423 measured this operand safe and the mechanism holds: the
  membership test is `r in tracked`, so a glob over-match returns the *neighbour's* name and `r` is
  **dropped**, never wrongly kept. Kept anyway — three literal operands beside one bare one is the
  next reader's trap — and the code comment says exactly that rather than implying it fixes a bug.

## Notes for review

- Metachar fixtures are built in **Python, never a shell**: fish/zsh glob-expand `docs[a]` / `a*b` /
  `q?r` into non-existence, and `git rm --cached '<glob>'` takes a pathspec too — both produced void
  fixtures (tests passing because they tested nothing) on #406.
- `*` and `?` are **illegal in Windows filenames**, so only those fixture rows are Linux-skipped.
  The `[…]` rows run on both platforms, and per #423's reachability analysis brackets are also the
  realistic carrier, since the value comes from operator config.
- The `safe_rollback` test asserts the **under**-revert direction, not #423's. The issue's fixture
  snapshots *before* the operator's edit, so the bare operand destroys it; `safe_rollback`'s
  snapshot is a `stash create` of the *dirty* tree, so the real harm inverts — the glob hands the
  dev attempt's change back after the reset was meant to discard it.

Refs #433. Closes #423.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of configured file paths containing wildcard or special characters during Git operations.
  * Prevented harmless Git warnings from incorrectly marking clean working trees as modified.
  * Improved reliability of tracking checks, targeted commits, exclusions, and rollback restoration.
  * Preserved correct error reporting when Git commands genuinely fail.

* **Tests**
  * Added coverage for special-character paths, ignored files, deleted tracked files, rollback behavior, and Git diagnostic output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -260,6 +260,25 @@ story <id>`, the same annotation a sweep bundle writes. Both sprint and stories 
 
 ### Fixed
 
+- **A configured path carrying `[`, `]`, `*` or `?` no longer makes git act on the wrong files
+  (#423).** `implementation_artifacts` reaches git verbatim out of the operator's
+  `_bmad/bmm/config.yaml`, and git reads a positional operand as a _pathspec_, not a path — so such
+  a name matched a wider set than the one it named, always on the over-match side. `commit_paths`
+  broke its own "commit exactly these paths (and nothing else)" promise, staging an unrelated
+  sibling's edit under a story's name; with the ledger gitignored it could also exit clean having
+  committed nothing, where the fixed form leaves a record. `has_changes_since` and `attempt_dirty`
+  excluded a sibling tree along with the artifacts dir and reported a changed attempt as CLEAN, and
+  their git half disagreed with the Python half about what "under the artifact dir" means.
+  `safe_rollback`'s preserve restore reached siblings too, handing back a change the reset had just
+  discarded. Every operand is now literal, and the two halves agree.
+
+- **A noisy git config no longer makes a pristine tree read as dirty.** `worktree_clean` compared
+  git's stdout and stderr merged, so `status` exiting 0 while warning about an unexecutable
+  `core.fsmonitor` hook or an unknown `core.fsyncMethod` was indistinguishable from a porcelain
+  record. Seven callers gate on it and three of them refuse the command outright, so such a host
+  could never start a run — and the message named no file. Only stdout is read now; the error path
+  still reports both.
+
 - **An undecodable `policy.toml` or `_bmad/bmm/config.yaml` is reported, not a crash.**
   `UnicodeDecodeError` is a `ValueError`, not an `OSError`, so a config file saved in UTF-16 or
   latin-1 escaped every `except (PolicyError, OSError)` handler in the codebase — and those handlers
--- a/src/bmad_loop/verify.py
+++ b/src/bmad_loop/verify.py
@@ -211,15 +211,40 @@ def last_commit_for(repo: Path, path: Path) -> str:
 
 
 def worktree_clean(repo: Path) -> bool:
-    # The orchestrator's own config file (.bmad-loop/policy.toml) is excluded:
-    # the TUI settings editor rewrites it, and a tracked config edit must not
-    # count as a "dirty tree" that blocks run/sweep/validate or forces a commit.
-    # Scope is policy.toml only — the deferred-work ledger also lives under
-    # .bmad-loop/ and is meant to be committed (see sweep._commit_ledger).
-    rc, out = _git(repo, "status", "--porcelain", "--", ".", f":(exclude){POLICY_FILE_REL}")
-    if rc != 0:
-        raise GitError(f"git status failed in {repo}: {out}")
-    return out == ""
+    """True when the tree holds no change git would report.
+
+    The orchestrator's own config file (.bmad-loop/policy.toml) is excluded: the TUI
+    settings editor rewrites it, and a tracked config edit must not count as a "dirty
+    tree" that blocks run/sweep/validate or forces a commit. Scope is policy.toml only
+    — the deferred-work ledger also lives under .bmad-loop/ and is meant to be
+    committed (see sweep._commit_ledger).
+
+    Reads `stdout` ALONE rather than `_git`'s stdout+stderr merge, for the reason
+    :func:`path_tracked` spells out: `status` exits 0 while still writing to stderr (a
+    `core.fsmonitor` hook that cannot exec, an unknown `core.fsyncMethod`, a stale
+    index advisory), and against the merged stream that chatter is indistinguishable
+    from a porcelain record — a pristine tree answers DIRTY. That direction is not
+    benign here: seven callers gate on it, and `cli.py`'s three refuse the command
+    outright, so a host with a noisy git config could never start a run and the
+    message would name no file. The error path keeps the merge, where stderr is the
+    only informative half."""
+    proc = _run_git(
+        [
+            "git",
+            "-C",
+            str(repo),
+            "status",
+            "--porcelain",
+            "--",
+            ".",
+            f":(exclude){POLICY_FILE_REL}",
+        ],
+        repo,
+    )
+    if proc.returncode != 0:
+        merged = (proc.stdout + proc.stderr).strip()
+        raise GitError(f"git status failed in {repo}: {merged}")
+    return proc.stdout.strip() == ""
 
 
 def same_commit(a: str, b: str) -> bool:
@@ -317,12 +342,47 @@ def attempt_dirty(
 
 
 def _exclude_specs(dirs: tuple[str, ...]) -> list[str]:
-    """git pathspec `:(exclude)<dir>` args for each repo-relative dir prefix."""
-    return [f":(exclude){d}" for d in dirs]
+    """git pathspec `:(exclude,literal)<dir>` args for each repo-relative dir prefix.
+
+    `literal` for the same reason as :func:`_literal_specs` — git reads a positional
+    operand as a PATHSPEC, so `[`, `]`, `*` and `?` in an operator-configured dir are
+    wildmatch metacharacters — but the harm here runs the other way: an over-matching
+    exclusion HIDES a diff instead of exposing a file. `has_changes_since` and
+    `attempt_dirty` both spend these on `diff --quiet . :(exclude)<dir>`, so a dir
+    whose name carries a `*` excludes a sibling tree as well and the attempt reads
+    CLEAN when it changed — the same false "no changes" that a dev attempt's dirtiness
+    check exists to prevent (#423 item 3).
+
+    It also realigns this half with :func:`_path_under_any`, the Python `startswith`
+    that filters the untracked half of the very same `has_changes_since` call. The two
+    disagreed on exactly the shapes that glob (#423 item 4): the tracked half excluded
+    a path the untracked half still counted, so one function's two branches answered
+    differently about what "under the artifact dir" means. Literal is the reading
+    `_path_under_any` already implements, which is why fixing the git half is the right
+    direction and not a coin flip.
+
+    Measured, not read off the docs: only a `*` (which crosses `/`) reaches a sibling
+    DIRECTORY's contents; a same-length `[…]`/`?` collision reaches a sibling FILE
+    only, because the pathspec has to match the whole path. Both magic words go in ONE
+    comma-separated `:(...)` prefix — the global `--literal-pathspecs` /
+    `GIT_LITERAL_PATHSPECS` form would disarm the `:(exclude)` magic too and silently
+    stop excluding anything at all. Verified identical at git 2.20.4 (the floor
+    `install._WORKTREE_CONFIG_GIT` declares) and 2.55.0."""
+    return [f":(exclude,literal){d}" for d in dirs]
+
+
+def _literal_specs(rels: list[str]) -> list[str]:
+    """git pathspec `:(literal)<rel>` for each repo-relative posix path — the operand
+    form that means "this path", not "this glob". See `path_tracked` for why."""
+    return [f":(literal){r}" for r in rels]
 
 
 def _path_under_any(path: str, prefixes: tuple[str, ...]) -> bool:
-    """True if repo-relative posix `path` equals or sits under any `prefixes` dir."""
+    """True if repo-relative posix `path` equals or sits under any `prefixes` dir.
+
+    The literal reading of "under", and since #423 item 4 the one `_exclude_specs`
+    agrees with — the two filter the tracked and untracked halves of a single
+    `has_changes_since` answer and must not disagree."""
     return any(path == p or path.startswith(p.rstrip("/") + "/") for p in prefixes)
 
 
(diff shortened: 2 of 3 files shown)


pbean:
@codex review

pbean:
@coderabbitai review

pbean:
@coderabbitai Thanks — reviewed and **not taking this one**, on measurement rather than preference.

The guideline (`AGENTS.md`: *terse, scannable, imperative*) is real, and "scannable" here is served by the **bolded lead sentence** every entry in this file opens with — which both of these do. But the proposed shape is not what the file converges on. Counting the `### Fixed` entries under `## [Unreleased]` (n = 42, by line count per bullet):

```
min 4 · median 9 · mean 11.7 · max 53
[4,5,5,5,5,5,6,6,6,6,7,7,7,7,7,8,8,8,8,8,9,9,9,9,10,10,10,
 11,11,11,11,11,12,12,13,14,14,14,22,35,53,53]
```

- the pathspec entry is **11 lines** — 15 of 42 entries are at or above it, so it sits mid-band, not at the tail;
- the `worktree_clean` entry is **6 lines** — only 6 of 42 entries are shorter than it.

The proposed condensation is one to two lines. **No entry anywhere in `## [Unreleased]` is ≤ 2 lines** (n = 61, minimum 3), so adopting it would make these the two shortest entries in the section — a move away from the convention, not toward it.

On the duplication point specifically: restating the mechanism is the house style here, deliberately — a CHANGELOG reader is the person who is *not* going to open `verify.py`, and the docstrings carry the measurements (git 2.20.4/2.55.0 floors, the `match_pathspec_item` fall-through, the over-match direction proof) that the entries do not. The two entries directly above these — `An undecodable policy.toml …` (8 lines) and `diagnose pseudonymizes the spec name …` (10 lines) — are the same shape, and you reviewed both on #439 two days ago without raising it.

Trimming only these two would leave the section internally inconsistent, so they stay as written. No other findings in this round; merging.

Issue #393, opened 2026-07-30
[BUG] validate and run blocked: bmm >=6.10.1 renamed the dev primitive to bmad-build-auto, leaving bmad-dev-auto a shim


tjoignant:
### Description

On a project running BMad Method **6.10.1-next.34**, `bmad-loop validate` fails and `bmad-loop run` aborts before spawning any session. bmm renamed the dev primitive `bmad-dev-auto` → **`bmad-build-auto`** and left `bmad-dev-auto` behind as a deprecation shim (a lone `SKILL.md` that forwards). bmad-loop 0.9.0 still targets the old name and content-probes it for marker files that now live under the new name.

There is no way to configure around it, and every released tag predates the rename, so the loop is unrunnable on current bmm.

### Steps to reproduce

1. Install bmm 6.10.1-next.34 in a project (`npx bmad-method install`), with the `claude` skill tree.
2. `uv tool install "bmad-loop[tui] @ git+https://github.com/bmad-code-org/bmad-loop.git"` (resolves to `main` @ `5c86f98`, version 0.9.0).
3. `bmad-loop init --project . --cli claude` — succeeds; hooks, skills, policy all fine.
4. `bmad-loop validate --project .`

### Expected behavior

Preflight recognises the project's dev primitive and passes, so `run` can start. bmm ships a complete dev primitive here — it is simply called `bmad-build-auto`, and both marker files bmad-loop demands are present in it.

### Actual behavior

```
FAIL: .claude/skills/bmad-dev-auto is incomplete (missing step-04-review.md, customize.toml) — reinstall it from the bmm module
```

Every other check passes (config, policy, sprint-status, git, tmux, process host, `claude` found, hooks registered). The suggested remediation cannot fix it: reinstalling bmm re-renders the same shim, by design.

Critically, this is **not** validate-only. `missing_base_skills` returns severity `problem`; `_require_base_skills` (`cli.py:494`) maps that to `False`, which gates **`run`** at `cli.py:912`, plus `cli.py:1144` and `cli.py:1199`. The run aborts before any session spawns.

### Analysis

bmm 6.10.1 moved the primitive and shimmed the old name. From the project's `_bmad/_config/skill-manifest.csv`:

```csv
"bmad-build-auto","bmad-build-auto","One iteration of an unattended development loop. Use when invoked by name.","bmm","_bmad/bmm/4-implementation/bmad-build-auto/SKILL.md"
[...]

PR #444, opened 2026-08-03
Resolve the dev primitive skill dynamically (bmad-build-auto / bmad-dev-auto)

ron332:
## Summary
- The bmm module renamed the inner dev primitive skill `bmad-dev-auto` → `bmad-build-auto`, leaving `bmad-dev-auto` installed as a permanent one-file forwarding shim (no `step-04-review.md`/`customize.toml` of its own).
- `bmad-loop` still hardcoded `bmad-dev-auto` as the primitive everywhere: the `validate` marker check, review-layer resolution, `_bmad/custom/` override paths, the stories-dispatch probe, and worktree provisioning's copy list. This made `bmad-loop validate` report a correctly-installed post-rename project as `skills.base-incomplete`.
- Worse than the validate cosmetic: review-layer resolution was reading the forwarding shim's file instead of the real primitive's `step-04-review.md`/`customize.toml` — a functional bug that would affect live dev/review sessions on any post-rename bmm install, not just `validate` output.
- Adds `resolve_primitive_skill(project, tree)` as the single place that decides which skill is a project's real dev primitive: prefers `bmad-build-auto` whenever it's installed there at all (existence, not completeness — a half-upgraded install still resolves to `bmad-build-auto` so the marker gap is reported against the right name), falling back to `bmad-dev-auto` only when `bmad-build-auto` is entirely absent (pre-rename bmm installs).
- Every consumer (`missing_base_skills`, `resolve_review_layers`, `_merged_review_layers`, `missing_stories_support`, the `_bmad/custom/` override filenames, and `provision_worktree`'s copy list via the new `base_skills_for_tree`) now resolves through that one function instead of a hardcoded literal, so they can never disagree about which skill is authoritative for a given project/tree.
- `DEV_BASE_SKILLS`, `BASE_SKILLS`, `STORIES_PROBE_SKILL` are kept exported under their legacy values for backward compatibility with existing call sites.

## Test plan
- [x] `pytest tests/test_install.py` — 234 passed (2 pre-existing/unrelated failures on `main`, macOS/APFS undecodable-filename tests, verified via `git stash`)
- [x] Full suite `pytest tests/` — 3816 passed, 48 skipped, 1 pre-existing/unrelated failure (same macOS/APFS issue), reproduced identically before and after this change
- [x] `ruff check` / `ruff format --check` clean on all touched files
- [x] `pyright` clean (0 errors) on `install.py`, `worktree_flow.py`, `cli.py`
- [x] Added `TestResolvePrimitiveSkill` (5 cases) plus coverage for `missing_base_skills`, `resolve_review_layers`, customize-override precedence, `base_skills_for_tree`, and `provision_worktree` against a `bmad-build-auto` project
- [x] Verified end-to-end against a real post-rename bmm project: `bmad-loop validate` goes from a false `FAIL: skills.base-incomplete` to a clean pass

## Summary by CodeRabbit

* **New Features**
  * Added support for resolving the preferred development skill while retaining compatibility with legacy installations.
  * Isolated worktrees now receive the appropriate project-specific development and review skills.
  * Installation diagnostics identify the selected development skill.
  * Installation now excludes runtime files and policy configuration from version control.

* **Documentation**
  * Updated validation and dispatch messages to reference the resolved development skill instead of a fixed name.

* **Tests**
  * Added coverage for preference, fallback, customization, worktree provisioning, and legacy compatibility.

### DIFF
--- a/src/bmad_loop/cli.py
+++ b/src/bmad_loop/cli.py
@@ -370,7 +370,7 @@ def cmd_validate(args: argparse.Namespace) -> int:
     if profiles and not any(f.severity == "problem" for f in base_findings):
         report.ok(
             "skills.base",
-            "upstream skills present (bmad-dev-auto + review layers)",
+            "upstream skills present (dev primitive + review layers)",
             {"trees": list(dict.fromkeys(p.skill_tree for p in profiles))},
         )
     report.extend(base_findings)
@@ -492,17 +492,17 @@ def _mux_set(project: Path, args: argparse.Namespace) -> int:
 
 
 def _require_base_skills(project: Path, pol, *, require_stories: bool = False) -> bool:
-    """Preflight the upstream skills the orchestrator drives (bmad-dev-auto + the
-    review layers it invokes inline).
+    """Preflight the upstream skills the orchestrator drives (the resolved dev
+    primitive + the review layers it invokes inline).
 
     Returns True when everything is in place; otherwise prints the problems and
     returns False so the caller can abort before spawning any session (a missing
     skill would otherwise stall as an `Unknown command` until the run times out).
     Warnings are printed but never abort — only ``problem`` findings block.
 
-    ``require_stories`` additionally content-probes bmad-dev-auto for folder+id
-    dispatch — stories mode needs a newer skill than sprint mode, so an older
-    install must fail loudly here rather than HALT `no stories.yaml`-style at
+    ``require_stories`` additionally content-probes the resolved primitive for
+    folder+id dispatch — stories mode needs a newer skill than sprint mode, so an
+    older install must fail loudly here rather than HALT `no stories.yaml`-style at
     dispatch time."""
     from .adapters.profile import ProfileError, get_profile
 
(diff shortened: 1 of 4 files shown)


pbean:
Superseded — #436 landed the same end-to-end resolution on `main` on 2026-08-02 (tracking issue #393), a few hours before this was opened, which is also why it now conflicts.

`install.resolve_dev_primitive` (install.py:244) is the single resolver, and every consumer named here goes through it or its total form `dev_primitive_or_default` (install.py:362): `missing_base_skills` (install.py:726), `resolve_review_layers` (install.py:547), `missing_stories_support` (install.py:611), `_customize_overrides` (install.py:225 — resolved name only, never both eras), and worktree provisioning — the seed floor and the completeness gate resolve via `dev_primitive_or_default` (worktree_flow.py:113 and :201), while the copy list unions a `BASE_SKILLS` that names both eras copy-if-present (install.py:200-205, consumed at worktree_flow.py:94), so a build-auto checkout seeds build-auto and a pre-rename one seeds dev-auto. Different shape from `base_skills_for_tree` here, same outcome. `validate` reports the name that actually resolved, per tree (cli.py:443).

Pinned in tests/test_install.py by `test_resolve_dev_primitive_matrix` (the 8-row resolution matrix) plus the per-consumer tests: `test_resolve_review_layers_reads_the_resolved_primitives_customize`, `test_review_layers_ignore_the_legacy_dir_on_a_renamed_project`, `test_stories_probe_follows_the_resolved_primitive`, and `test_provision_worktree_copies_the_renamed_dev_primitive`.

One deliberate difference, and the reason this can't be adopted verbatim: `resolve_dev_primitive` returns `None` for a legacy install missing either marker. On disk that shape is indistinguishable between the post-rename forwarding shim and a truncated pre-rename install, so `missing_base_skills` emits `skills.base-shim` naming both causes and the one remediation they share — including the interactive migration gate that HALTs an unattended session with nothing written to disk (tests/test_install.py::test_shim_only_install_is_refused_with_the_halt_hazard_named). `resolve_primitive_skill` here is total and returns `bmad-dev-auto` for that same tree, so the shim-specific diagnosis would degrade to a generic `skills.base-incomplete` against the shim's own name.

Thanks for the report and for verifying end to end against a real post-rename project — closing in favor of what already landed.

Issue #443, opened 2026-08-03
Plumb `paths.project` through worktree provisioning so `repo_root != project` is supported, and retire the refusal


pbean:
Supersedes #414, and replaces the refusal that closed it in the short term.

#414 offered two fix shapes: (1) plumb `paths.project` into `provision_worktree` for the reads that mirror what `init`/`validate` touch, keeping `repo_root` for git operations; or (2) refuse the combination until (1) exists. The forward-port of the 0.9.1 hotfix took **(2)** — `bmadconfig.worktree_isolation_conflict` now refuses `isolation = "worktree"` under a `repo_root` override at `validate`, at `run`/`sweep`/`resume`, in the auto-triggered child sweep, in the dry-run banner and in the TUI's pre-launch guard. That was a deliberate scope decision, not a verdict on the design: shape (1) rewrites every disk read in `provision_worktree`, and it has an open design question shape (2) does not.

This issue is shape (1). Landing it **deletes the refusal**, and that deletion is part of the work — see "What this removes" below.

### The open design question

`bmadconfig.load_paths` *requires* `project/_bmad/bmm/config.yaml` (`src/bmad_loop/bmadconfig.py:194-196`), so the `_bmad/` surface is under `project` by definition and `repo_root/_bmad/` generally does not exist. But "which root wins" is not one answer — it is one answer *per surface*, and each has to be decided rather than defaulted:

- the **hook relay** is pointed at, never copied: `init` writes `project/.bmad-loop/bmad_loop_hook.py` (`install.py`), and provisioning bakes an absolute path from `repo_root` into the worktree's hook config (`src/bmad_loop/worktree_flow.py:828`). This one is unambiguous — the relay only exists under `project`.
- **skill trees** and **`_bmad/custom/` overrides** are copied from `repo_root` (`src/bmad_loop/worktree_flow.py:1023-1030`, `src/bmad_loop/worktree_flow.py:405`) and probed under `project` by `missing_base_skills` / `resolve_review_layers`. The preflight is the authority on which layers a run gets, so these follow `project`.
- **`seed_files` / `seed_globs`** (`src/bmad_loop/worktree_flow.py:858-865`, `src/bmad_loop/worktree_flow.py:930-943`) use `repo_root` as both the copy source *and* the `is_relative_to` containment base. These are operator-configured and may legitimately name paths outside `project` but inside the git root — this is the one where `repo_root` is arguably still right, and changing it is a behaviour change for existing monorepo users.
[...]

Issue #414, opened 2026-07-31
Worktree provisioning reads repo_root while init/preflight read project — under repo_root != project the seed and its gates go inert


pbean:
Worktree provisioning reads the "main repo" at `paths.repo_root`; `init`, `validate` and the run preflight write and probe at `paths.project`. The two are the same path by default, so the split is invisible — but `_bmad/bmm/config.yaml` may set `repo_root` (`bmadconfig.py:86-87`) to decouple the git root from the project dir (documented for monorepos in `README.md:446`, `docs/FEATURES.md:73`). When it does, every provisioning read misses, and the checks meant to notice go inert instead of firing.

`load_paths` requires `project/_bmad/bmm/config.yaml` (`bmadconfig.py:72-74`), so the `_bmad/` surface is by definition under `project`. `repo_root/_bmad/` is not required to exist, and generally does not.

### The paired reads (main)

| surface | written / probed at `project` | seeded / probed at `repo_root` |
|---|---|---|
| hook relay | `install.py:2344` (`init` writes `project/.bmad-loop/bmad_loop_hook.py`) | `worktree_flow.py:142` (`relay = repo_root / HOOK_SCRIPT_REL`, baked absolute into the worktree hook config) |
| base skills (dev primitive + review layers) | `install.py:405` `missing_base_skills(project, trees)`, called from `cli.py:366`, `cli.py:515` | `worktree_flow.py:235` `src = repo_root / tree / skill` |
| review-layer resolution | `install.py:483` `resolve_review_layers(project, tree)` | `worktree_flow.py:229` `resolve_review_layers(repo_root, tree)` |
| `_bmad/custom/` overrides | `install.py:130` `CUSTOMIZE_DIR`, resolved under `project` by the preflight | `worktree_flow.py:253` `src_customize = repo_root / CUSTOMIZE_DIR` |

Call site: `worktree_flow.py:489-492` passes `self.paths.repo_root` into `provision_worktree` (`worktree_flow.py:70-78`). Nothing passes `paths.project`.

The comment at `worktree_flow.py:241-247` states the intent exactly — "the preflight resolves review layers through these files, but the run inside the worktree resolves them again from ITS OWN project root ... without this the two disagree, and validate approves a layer set the isolated run never runs" — and then reads the wrong root.

### The renderer surface (hotfix/0.9.1-build-auto, PR #406)

The 0.9.1 renderer port adds a third layer with the same split, plus two completeness checks that go inert rather than fail:

[...]

PR #445, opened 2026-08-03
feat(cli,tui): refuse worktree isolation under a repo_root override (#433, 6E)

pbean:
Sub-phase 6E of the 0.9.1 forward-port (#433): ports the **#414 refusal**, not the plumbing.

Worktree provisioning reads `paths.repo_root` for every surface it seeds *off disk* — the upstream skill trees, `_bmad/` and the `_bmad/custom/` overrides inside it, each `seed_files`/`seed_globs` entry — and bakes the hook relay's absolute path from it into the worktree's hook config. `init`, `validate` and the run preflight write and probe those same surfaces under `paths.project`. `bmadconfig.load_paths` *requires* `project/_bmad/bmm/config.yaml`, so `_bmad/` is under `project` by definition and `repo_root/_bmad/` generally does not exist at all.

The two are the same path by default, so the split is invisible — until `_bmad/bmm/config.yaml` sets `repo_root`, which is the documented monorepo knob. Then the preflight approves a surface the isolated run never receives, and the seed-completeness gates go **inert rather than fire**: the session dispatches into a worktree with no dev primitive, and stops with no result and nothing journaled naming the cause.

## Why a refusal and not the fix

Deliberate scope call, carried over from the 0.9.1 line. Shape (1) — plumb `project` through provisioning — rewrites every disk read in `provision_worktree`, which is character-for-character the diff of two later sub-phases, and "which root wins" is a **separate decision per seeded surface**: the relay only exists under `project`, but operator-configured `seed_files` may legitimately name a path outside it and inside the git root. `ProjectPaths.rebased` encodes `project == repo_root` besides, so it cannot express the divergent case at all.

That work is filed as **#443**, which names every line this PR adds as something it deletes. `worktree_isolation_conflict`'s docstring says so in the source, so a later reader does not mistake the refusal for the intended end state.

## What lands

`bmadconfig.worktree_isolation_conflict(paths, isolation) -> str | None` is the **sole producer** of the text, so the five consumers cannot drift:

| site | disposition |
|---|---|
| `cmd_run`, `cmd_sweep`, `_resume_paused_run` | `_reject_isolation_conflict` → rc 1 |
| `_sweep_factory.factory` | **raises** `RuntimeError` |
| `cmd_validate` | `report.fail("policy.isolation-repo-root", …)`, keeps going |
| `_warn_preflight_would_abort` | `problems.insert(0, …)` — named first |
| `tui/app.py::_guarded` | error toast, ahead of the clean-tree gate |

Ordering is load-bearing in three places and each is pinned by a test:

- **ahead of the clean-tree and queue gates** in `cmd_run`/`cmd_sweep` and in the TUI. This refusal says the configuration cannot run *at all*, so answering "commit or stash first" would send the operator to fix something that is not the problem. The fixtures leave the tree genuinely dirty so a wrongly-ordered gate speaks instead.
- **`_sweep_factory` raises rather than returns.** By the time the engine calls the factory it has latched the trigger and journaled `sweep-auto-trigger`, and it reads a plain return as success — a bare decline would be recorded as `sweep-auto-finished`, i.e. a child sweep that ran and finished when none was ever launched. Raising lands on the `sweep-auto-failed` + notify path an unparseable `policy.toml` already takes.
- **`validate` reports rather than aborts**, so its other gates still run.

`_resume_paused_run` gets it too: resume re-reads both files off disk, so a run started before the override was added must not finish its remaining stories through provisioning the preflight would now refuse. It is refused before the `run-resume` journal entry, so the journal does not record a resume that never happened.

## Main-side adaptations (this is a re-implementation, not a cherry-pick)

- **`_warn_preflight_would_abort` keeps main's `severity == "problem"` filter.** The 0.9.x version had none. Main additionally carries `test_dry_run_banner_stays_silent_for_a_warning_only_finding`, whose docstring calls the filter *"a deliberate correction, not a port artefact"* — so only the parameter changed (`project: Path` → `ProjectPaths`, at all three call sites) and the insert was added. Dropping the filter would have printed warnings as `FAIL:` lines.
- Its docstring's claim that it *"reads the same finding list `_require_base_skills` gates on"* became narrower than the behaviour once the banner also reported the isolation conflict — which `_require_base_skills` does not check. Rewritten to name **both** gates it mirrors, and to say which refusals are deliberately not in the banner (dirty tree, queue, run id).
- `_findings_by_check` does not exist on main; its equivalent is `_validate_findings`. `_make_validate_pass` takes `policy=`, not `policy_text=`. `_render_findings` was inlined in one TUI-render test on main and is now a helper.
- The `except (BmadConfigError, PolicyError, OSError)` fall-through in the TUI guard depends on both loaders converting `UnicodeDecodeError` (a `ValueError`, not an `OSError`) into their own typed error. **Verified present on main** at `policy.py:684` and `bmadconfig.py:125` — they landed in 6C — and an ablation row proves the `PolicyError` leg is reachable, not merely present.
- 0.9.x also has a paragraph in a seed-completeness helper in `install.py` citing these guards. Main has none of those helpers yet, so it has no home here; it is recorded on #443.

## Verification

- **3964 collected**, 3936 passed, 24 skipped. That is `main`'s 3953 + the 11 tests added here. The 4 reds are pre-existing `test_module_skills_sync` local skill-drift on the dev box and do not occur on a clean checkout.
- `uv run pyright` **0 errors, 0 warnings, 0 informations** — zero delta from `main`. `checks.py` is in the strict list and gained an id only, never logic.
- `trunk check --all --no-fix` — 242 files, no issues.

### Ablation matrix — every row against the **whole** suite

Collected total constant at **3964** on all 11 runs (no baseline drift); the full `src/` digest was re-verified pristine after every row.

| row | edit ablated | new reds | which |
|---|---|---|---|
| A | `cmd_run` refusal | 1 | `test_start_refuses_…[run]` — fails on `assert REFUSAL in err`, having got *"git worktree is not clean"*: the later gate speaks, which is exactly the ordering claim |
| B | `cmd_sweep` refusal | 1 | `test_start_refuses_…[sweep]` — no cross-bleed between the two params |
| C | `_resume_paused_run` refusal | 1 | `test_resume_refuses_…` — fails `engine constructed`, i.e. resume proceeded to build |
| D | `_sweep_factory` raise | 1 | `test_auto_sweep_refuses_…` — `DID NOT RAISE RuntimeError` |
| E | `cmd_validate` finding | 1 | `test_validate_refuses_…` — rc 0 where 1 was asserted |
| F | dry-run banner insert | 1 | `test_dry_run_banner_names_the_isolation_refusal_first` — banner still prints, so this pins the isolation **line**, not banner silence |
| G | TUI `_guarded` refusal | 1 | `test_worktree_isolation_…_blocks_launch` |
| G2 | TUI refusal **moved below** the clean-tree gate | 1 | same test — the ordering claim, which G alone cannot pin |
| G3 | `except` tuple narrowed to drop `PolicyError` | 1 | `test_unreadable_policy_falls_through_…` — uncaught `PolicyError` takes the app down |
| H | `.resolve()` on both sides of the path compare | 1 | `test_worktree_isolation_conflict_compares_normalized_paths` |
| I | the `VALIDATE_CHECKS` id | 1 | `test_validate_refuses_…` — the registry assert bites, so a future id cannot skip registration |

Every row reddens exactly one test, inside the edit's blast radius, and no row reddened nothing.

Row H is the one that matters most for a refusal gate: **a false positive here is worse than the bug it forecloses**, since it would refuse an ordinary isolated project whose `repo_root` merely spells the same directory a different way. `load_paths` resolves both sides, but nothing obliges a hand-built `ProjectPaths` — or a future caller — to have done so.

Refs #433, #414. The plumbing is tracked as #443.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented worktree-isolated operations from starting when a different repository root is configured.
  * Added clear conflict reporting across validation, run, sweep, resume, dry-run, and TUI workflows.
  * Ensured auto-triggered sweeps fail safely instead of reporting successful completion.

* **Documentation**
  * Documented that custom repository roots cannot be combined with worktree isolation.


pbean:
@codex review

pbean:
@coderabbitai review

pbean:
Taking this one — applied in `3eb37d7`, essentially your wording.

Worth saying why, since I declined the same-shaped finding on #441 two hours ago and do not want that to read as inconsistency. There I measured the section: the entries sat at the 64th and 14th percentile of the 42 `### Fixed` entries, and the proposed 1–2 line form had **no precedent anywhere** in `## [Unreleased]` (n = 61, minimum 3 lines) — so the proposal was out-of-band, not the entries.

This proposal is **6 lines**, comfortably inside that distribution. So both versions were in-band, which stopped it being a measurable question and made it a judgement one — and on the merits you are right. What your version drops is the *mechanism* (why `repo_root/_bmad/` does not exist, what `load_paths` requires, why the seed-completeness gates went inert), and in this PR that text is not merely paraphrased elsewhere, it is **verbatim in `worktree_isolation_conflict`'s docstring**, which this same PR adds as the canonical explanation. Duplicating it in the changelog earns nothing and gives it a second place to drift.

Kept everything user-facing: the behaviour change, all five refusal surfaces, and the #443 pointer for the plumbing that retires this.

No other findings this round.

PR #453, opened 2026-08-04
fix(engine): harden ledger snapshot cleanup

### LINKED ISSUE #420: Ledger snapshot disarm is skipped when _rollback_or_pause raises anything but RunPaused
Found while closing out PR #406 (the pre-harvest ledger snapshot work, `0a8088f`). Recorded rather than fixed there because it is unreachable from the policy every test in that family uses.

## What

`_dev_phase`'s non-fixable RETRY leg latches the pause so its `finally` can tell that leg apart (`engine.py:2078-2090`):

```python
paused = False
try:
    self._rollback_or_pause(task)
except RunPaused:
    paused = True
    raise
finally:
    ...
```

The `except` catches **only** `RunPaused`. `_rollback_or_pause` (`engine.py:1140`) does not always pause — with `scm.rollback_on_failure = on`, and unconditionally inside a mounted unit worktree, it performs a `git reset --hard` to the attempt baseline. That reset can raise `verify.GitError`, and the preserve-ref parking around it can raise `OSError`.

On that path the exception is neither `RunPaused` nor caught here, so:

1. `paused` stays `False`, so the `finally` reads the leg as a plain rollback rather than a pause; and
2. the disarm of `pre_harvest_ledger` / `pre_harvest_ledger_captured` never runs, while `run()`'s `finally: self._save()` (`engine.py:420`) still persists the task — leaving a **still-armed snapshot on disk** for a task whose attempt died.

## Why it is not already caught

Every test in the ledger-snapshot family runs on the default `rollback_on_failure = off`, where `_rollback_or_pause` raises `RunPaused` by construction and the latch fires correctly. The uncovered combination is `rollback_on_failure = on` (or worktree isolation, which always auto-recovers) **plus** a git/IO failure during the reset — two conditions, neither of which any current fixture supplies together.

## Severity

Bounded. The residue is state in `state.json` on a task that is going terminal, and nothing reads `pre_harvest_ledger` on a terminal task — the restore path gates on `pre_harvest_ledger_captured` only when a replay reaches the same attempt. The concrete risk is a resume that finds an armed snapshot from an attempt that never completed, and restores ledger text captured before a harvest that may since have been reverted by other means.

## Shape of a fix

Either widen the latch to the exception types `_rollback_or_pause` can actually raise, or — probably better — move the disarm out of the exception-shaped bookkeeping entirely, so it is unconditional on leaving the attempt rather than conditional on which exception left it. Worth checking the same question at the other `except RunPaused` (`engine.py:307`).

Whichever shape, it needs a fixture that sets `rollback_on_failure = on` and makes the reset fail — the ledger family has no such fixture today, which is the actual gap.

pbean:
## Summary

- restore, rebase, disarm, and durably save the pre-harvest ledger snapshot on every non-fixable RETRY unwind, including a failing automatic reset
- preserve an earlier pause/reset exception when ledger restoration also fails, journal the secondary repair fault, and complete durable disarm before propagating the appropriate error
- spend an inherited snapshot in the main-only `awaiting-operator` park before its commit/human window
- cover tracked, untracked, ignored, and external ledgers; crash replay; operator edits; retry chains; all four decision paths; and direct state-file durability boundaries

## Decision handling

| Decision | Snapshot handling | Durability |
| --- | --- | --- |
| PROCEED | Disarm after acceptance, before review handoff | Explicit save because resume may bypass `_dev_phase` |
| RETRY, non-fixable | Restore, rebase from captured snapshot, then disarm; fixable RETRY stays armed | Returned rollback uses next attempt save; pause/reset failure saves while unwinding |
| DEFER | Disarm without restoring the harvest; defer intentionally keeps the entry | `_defer` persists |
| ESCALATE | Disarm while preserving the tree/knowledge | `_escalate` persists before pausing |

If ledger restoration itself raises `OSError`, cleanup journals it and still disarms/saves. An earlier unwind remains authoritative; without an earlier unwind, the restore fault is raised after its disarm is durable.

The separate main-only `awaiting-operator` park can resume a persisted DEV_VERIFY task without re-entering `_dev_phase`. After confirming the awaiting-operator status, it spends and explicitly saves any inherited snapshot before entering the commit and human-wait window.

## RETRY root cause

An ordinary untracked harvest-created ledger survives because `RecoveryFlow.safe_reset()` passes `keep=(".bmad-loop", *protected_relpaths())` and `verify.safe_rollback()` skips it. This is not caused by `git reset --hard`: Git's own reset revert applies only to a tracked ledger. Ignored ledgers are invisible to `untracked_files()` and therefore require explicit filesystem-snapshot restoration. External ledgers are outside the reset's reach and are restored explicitly as well.

## Durability witnesses

Removing the relevant local `_save()` calls produced 0 red in ordinary in-process result tests because `run()` saves again in its `finally` block. That does not cover the SIGKILL-sized durability window. The new tests read `state.json` directly before ambient saves: before harvest at arm time, before review handoff on PROCEED, at the `run-paused` event, at the #420 `run-crash` event, and before the awaiting-operator commit window. Restore-fault tests perform the same disk probe for pause, reset-error, and returned-rollback cases.

## Ablation

`RecoveryFlow.protected_relpaths()` was temporarily forced to return `()` and restored with the inverse patch after both whole-suite lanes:

- Python 3.13: 4,254 collected; 4,217 passed, 24 skipped, 13 failed. Four failures are the known local `.claude/.agents` seeded-skill drift. Of the nine shield-removal reds, the Phase 6K witness changed `ledger_present_after_reset` from `[True]` to `[False]`; the other eight are existing engine/recovery/stories tests that also depend on protected artifact paths.
- Python 3.14: 4,254 collected; the same 13 failures reproduced, plus one unrelated TUI scrolling timing failure (4,216 passed, 24 skipped, 14 failed). The isolated TUI rerun passed immediately.

## Verification

- Python 3.13 pre-review full suite: 4,226 passed, 24 skipped, plus only the four known local seeded-skill drift failures (4,254 collected)
- Final head `uv run pytest tests/test_engine.py -q`: 345 passed
- Python 3.14 full baseline: 4,226 passed, 24 skipped, plus the same four known local failures; focused suite before the review follow-up: 342 passed
- `uv run pyright`: 0 errors, 0 warnings, 0 informations
- `trunk fmt`: clean
- `trunk check --no-fix`: clean
- `git diff --check`: clean

Part of #433.

Closes #420

### DIFF
--- a/src/bmad_loop/engine.py
+++ b/src/bmad_loop/engine.py
@@ -1489,6 +1489,7 @@ class Engine:
                         # rollback. Detect the active unwind without limiting it to
                         # RunPaused: reset/preserve failures are the #420 gap.
                         unwinding = sys.exc_info()[0] is not None
+                        restore_error: OSError | None = None
                         # Recovery resets code/spec state but protects artifact
                         # directories through `_safe_reset`'s
                         # keep=(".bmad-loop", *self._protected_relpaths()) shield.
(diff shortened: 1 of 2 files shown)


pbean:
@codex review

Issue #471, opened 2026-08-06
worktree isolation breaks codex stages: .codex/hooks.json is not seeded, so the session registers no hooks and stalls


LuckierTrout:
## Summary

With `scm.isolation = "worktree"`, a **codex-routed stage registers no hooks at all**, because `.codex/hooks.json` is not in the codex profile's `seed_files`. The session runs, ends, emits no `Stop`, and the orchestrator stalls it to timeout — a silent hang with no error.

The claude profile is unaffected only by coincidence: its hook config path *is* one of its seeded files.

## The asymmetry

| profile | `seed_files` | `[hooks] config_path` | seeded? |
|---|---|---|---|
| `claude` | `.mcp.json`, **`.claude/settings.json`**, `.claude/settings.local.json` | `.claude/settings.json` | ✅ incidentally |
| `codex`  | `.codex/config.toml` | **`.codex/hooks.json`** | ❌ |

`.codex/hooks.json` is gitignored (it is machine-local state written by `bmad-loop init`), so a `git worktree` checkout — which takes tracked files only — omits it. The comment above codex's `seed_files` already articulates exactly this hazard for MCP config:

> Project-scoped Codex config (MCP servers under `[mcp_servers]`) — gitignored, so a worktree checkout omits it and isolated sessions lose their MCP servers.

The same reasoning applies to the hook registration, with a worse failure mode: losing MCP servers degrades a session, but losing hooks means the orchestrator never receives `Stop` and cannot tell the session finished.

## Environment

- `bmad-loop` 0.9.0 (`uv tool install` from git `f512288f4e7ca4f028da324b280762fd413b9f90`)
- `codex-cli` 0.146.0, macOS arm64
- `bmad-loop init --cli claude --cli codex`
- Policy: `scm.isolation = "worktree"`, `[adapter.dev] name = "codex"`

## Reproduce

1. `bmad-loop init --cli claude --cli codex`
2. Set `scm.isolation = "worktree"` and route any stage to codex, e.g.
   ```toml
   [adapter.dev]
   name = "codex"
   ```
3. `bmad-loop run --story <key>`
4. The codex session spawns in `.bmad-loop/runs/<id>/worktrees/<story>/`, which contains no `.codex/hooks.json`. No event files are written for it, and the run stalls to `session_timeout_min`.

## Workaround (confirmed working)

```toml
[scm]
worktree_seed = [".codex/hooks.json"]
```

[...]

Issue #392, opened 2026-07-30
[BUG] Worktree provisioning writes tracked hook config to shared info/exclude


StirlingGoetz:
## Summary

With `scm.isolation = "worktree"` and the Codex adapter, worktree provisioning adds `/.codex/hooks.json` to the repository's shared `.git/info/exclude`, even when `.codex/hooks.json` is already tracked and was not modified during provisioning.

This makes the tracked hook configuration appear ignored and can break repository-hygiene or pre-commit checks.

## Environment

- bmad-loop: 0.9.0
- Installed source revision: `232f3f7159004e960e786a6568ebe79daecf388d`
- Upstream `main` also affected at: `ad22f7966b0d0626bea3a22f0617f55c875b7810`
- Adapter: Codex
- Git: 2.54.0
- OS: Linux
- SCM policy:
  - `isolation = "worktree"`
  - `branch_per = "story"`
  - `seed_adapter_defaults = true`

## Steps to reproduce

1. Create a Git repository with a tracked `.codex/hooks.json`:

   ```bash
   git add .codex/hooks.json
   git commit -m "Track Codex hook configuration"Confirm no tracked files are ignored:
git ls-files -ci --exclude-standard

Configure bmad-loop to use the Codex adapter and worktree isolation.

Start a run that provisions an isolated story worktree.

Inspect the common Git exclude file:
git check-ignore --no-index -v .codex/hooks.json

Check for tracked-and-ignored files again:
git ls-files -ci --exclude-standard



### Expected behavior

Expected behavior
Worktree provisioning should:
Never make an already-tracked file ignored.
Avoid changing ignore behavior for the main checkout or unrelated worktrees.
Exclude only untracked files actually created by provisioning.
Remove any provisioning-owned metadata during worktree/run cleanup.

### Actual behavior

Actual behavior
After provisioning:
.git/info/exclude:...:/.codex/hooks.json .codex/hooks.json
and:
.codex/hooks.json
is returned by git ls-files -ci --exclude-standard.
provision_worktree() unconditionally adds every non-hookless profile's hook configuration to the exclusion set:
patterns |= {
    f"/{p.hooks.config_path}"
    for p in profiles
    if not p.hookless
}
_worktree_local_exclude() then resolves git rev-parse --git-common-dir and writes to:
common_dir / "info" / "exclude"
For a linked worktree, that is the main repository's shared Git directory. The exclusion therefore affects the main checkout and every linked worktree; it is not worktree-local.
[...]

PR #483, opened 2026-08-07
test(worktree): pin #392's teardown-residue arc end to end

pbean:
The one leg of #392's requested regression test that was never written.

The reporter asked for four things, and their fourth — *"deletes the run/worktree and verifies no exclude residue remains"* — had no test in their scenario. `test_shield_dies_with_the_worktree` pins the same lifetime from #384's angle, but with a plain fixture and no `ls-files -ci` probe, so nothing walked their arc in one piece: track the hook config → provision → probe **both** checkouts → tear down → prove no residue.

## Fixture

**codex**, not claude. `.codex/hooks.json` is codex's `config_path` and is **not** one of its `seed_files` — the reporter's exact shape. claude's `config_path` (`.claude/settings.json`) doubles as a seed file, which is the coincidence #471 reports, so claude cannot express it. Asserted as a precondition in the test so a profile change cannot quietly make it stop biting.

## Ablations, run singly

| ablation | reddens |
|---|---|
| drop the tracked-file filter (`_drop_inert_tracked_file_patterns`) | the in-worktree probe returns `.codex/hooks.json` — the reporter's path verbatim |
| point the shield write at the common dir | `assert private.is_file()`, so the guard is not vacuous |
| write the patterns to **both** excludes | `shared.read_bytes() == before`, so the residue comparison is live |

The post-teardown probe is **defense in depth, not an independent pin**, and the comment at the line says so rather than leaving a reader to assume otherwise: no single ablation reddens it alone, because every ignore source the main checkout can see after teardown is equally visible before it, so the earlier probes fail first. What it adds over the byte comparison is reach — a residue that is not the shared exclude's bytes (a `.gitignore` left in the tree, a surviving `core.excludesFile`) answers there and nowhere else in this test.

## Scope

Test-only. No behavior change, so no CHANGELOG entry — the behavior shipped in #385 (scope + lifetime) and #478 (the tracked-file filter).

Suite 4446 green, 25 skipped. pyright clean, `trunk check` clean.

## Issues

Covers the last untested ask in #392. The tracked-**directory** half of that report stays a documented non-fix and is tracked separately.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for cleaning up tracked hook configuration.
  * Verified worktree removal fully clears private configuration and ignore state.
  * Confirmed shared exclude settings remain unchanged.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -253,8 +253,40 @@ def make_validate_document(findings, *, stories_on: bool = False, spec_folder: s
     return documents.validate_document(report, stories_on, spec_folder)
 
 
+@pytest.fixture(scope="session", autouse=True)
+def _isolate_ambient_git_ignores(tmp_path_factory: pytest.TempPathFactory):
+    """Shadow the two ignore sources git reads from OUTSIDE the repo, for every test.
+
+    Developer boxes routinely carry `.claude/` or `.codex/` in a global gitignore —
+    ignoring your AI tooling everywhere is the obvious thing to do — and without this
+    the shield tests silently measure that instead of the shield. The failure is not
+    only a red: a global `.codex/` makes `git add -A` skip the hook config, so the
+    file the test believes it TRACKED is untracked, the tracked-file filter is never
+    reached, and `ls-files -ci --exclude-standard` answers "" because an untracked
+    file is not tracked-and-ignored. The test passes having exercised nothing.
+
+    Two sources, and both are needed: `core.excludesFile` from `~/.gitconfig`
+    (suppressed by pointing GIT_CONFIG_GLOBAL at a file that does not exist), and its
+    documented fallback `$XDG_CONFIG_HOME/git/ignore`, which still applies once the
+    key is unset. Session-scoped so `_project_template`'s own `add -A` is covered too.
+
+    `GIT_CONFIG_NOSYSTEM` is deliberately NOT set, for the reason
+    `test_shield_falls_back_to_home_when_xdg_is_relative` records: it would suppress
+    Git-for-Windows' system `core.autocrlf` and make unrelated tracked files read as
+    modified. A system-level excludes file therefore stays reachable — tests that
+    depend on a path really being tracked assert that as a precondition."""
+    env = tmp_path_factory.mktemp("git-env")
+    mp = pytest.MonkeyPatch()
+    mp.setenv("GIT_CONFIG_GLOBAL", str(env / "no-such-gitconfig"))
+    mp.setenv("XDG_CONFIG_HOME", str(env / "xdg"))
+    yield
+    mp.undo()
+
+
 @pytest.fixture(scope="session")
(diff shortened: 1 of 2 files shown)


pbean:
@codex review

pbean:
@codex Valid, and thanks — addressed in `d29372a`, though the fix moved and the diagnosis widened.

**The blast radius is 7 tests, not one.** `tests/conftest.py` pinned no git environment at all, so this is a harness gap rather than something about the new test. Reproduced with a global `core.excludesFile` containing `.codex/` and `.claude/`:

```
FAILED tests/test_cli.py::test_run_aborts_when_the_renderer_entry_source_is_missing
FAILED tests/test_install.py::test_shield_tracked_hook_config_is_not_excluded
FAILED tests/test_install.py::test_shield_tracked_hook_config_leaves_no_residue_after_teardown
FAILED tests/test_install.py::test_shield_tracked_skill_tree_keeps_its_pattern
FAILED tests/test_install.py::test_shield_keeps_patterns_when_tracked_probe_fails
FAILED tests/test_install.py::test_shield_never_touches_main_checkout
FAILED tests/test_install.py::test_shield_outranked_degrade_leaves_no_permanent_repo_format_change
```

**The consequence is worse than a red.** You wrote that a force-add leaves the `ls-files -ci` assertions failing. The more dangerous branch is the other one: when the commit is not empty, `add -A` simply skips the ignored path, the file the test believes it tracked is **untracked**, and `-ci --exclude-standard` answers `""` — because an untracked file is not tracked-*and*-ignored. Measured in a sandbox: the tracked-file filter is never reached and the test passes green having exercised nothing.

## What landed

Session-scoped autouse fixture shadowing both out-of-repo sources git reads — `core.excludesFile` from `~/.gitconfig`, and its documented `$XDG_CONFIG_HOME/git/ignore` fallback, which still applies once the key is unset. Session scope so the project template's own `add -A` is covered too.

**One third of your suggestion is deliberately not taken.** You asked to pin the global/system/XDG configuration; `GIT_CONFIG_NOSYSTEM` is left alone. `test_shield_falls_back_to_home_when_xdg_is_relative` already records why: it suppresses Git-for-Windows' system `core.autocrlf`, after which unrelated tracked files read as modified. Since that leaves a system-level excludes file reachable, the new test now asserts its own precondition — `verify.path_tracked_file(repo, hook_rel)` — so an ambient rule that slips through fails loudly on the precondition rather than passing on an untracked path.

## Ablations

| ablation | result |
|---|---|
| drop the `GIT_CONFIG_GLOBAL` pin | 7 tests redden under an ambient global ignore |
| drop the `XDG_CONFIG_HOME` pin | 6 tests redden under an ambient XDG ignore |
| let the commit succeed with the path still ignored | the new precondition assert reddens, naming `.codex/hooks.json` |

Suite green under a clean environment and under both ambient sources.

pbean:
@codex review

Issue #437, opened 2026-08-02
Review session can revert the orchestrator's own sprint-status write and dead-end the run


didostap:
### Description

The review prompt that bmad-loop dispatches tells the session to keep its hands off
deferred-work.md, but there is no equivalent sentence for sprint-status.yaml. I hit that gap.

Here is the review prompt from my run's journal.jsonl, verbatim:

  /bmad-dev-auto <spec-path> - If this review defers new findings, append them to
  the deferred-work ledger as NEW entries only; do NOT modify, re-open, or rewrite
  existing ledger entries - the orchestrator owns their status and resolution.

So the loop already knows it has to say "hands off, the orchestrator owns this" for one of
its two ledgers. There is nothing like it for the sprint board.

What happened: my story spec's Boundaries section had a line saying "Never edit epics.md,
the PRD, the spine, or sprint-status.yaml." The dev session finished, the orchestrator
flipped the row from backlog to done as its sign-off. The review session then read the
spec, saw the uncommitted done in git status, filed it as a [high] finding for "directly
violating the intent-contract's Never edit ... sprint-status.yaml", reverted the row to
backlog, and committed the revert.

From the reviewer's side that is a defensible read. It had no way to know that write came
from the orchestrator and not from a misbehaving dev session. My spec prohibited *writing*
the file, and the reviewer turned that into a mandate to *revert someone else's write*.

The contradiction detection worked exactly as advertised. on_status_contradiction =
"escalate" caught it and paused with a very clear message instead of burning the remaining
review cycles. No complaints there. The escalation is the symptom being caught correctly,
the bug is that it was reachable at all.

Two things I think are worth fixing:

1. Extend the review prompt appendix to cover sprint-status.yaml, same shape as the
   deferred-work sentence.

2. Optionally, commit the orchestrator's sprint-status write together with the dev pass
   rather than leaving it uncommitted. What the reviewer actually saw was a loose
   modification in git status with no author attached to it, which is what made it look
   like a violation.

Two things I found while digging that might matter to whoever picks this up:

The CHANGELOG entry for #334 says "Review prompts are unchanged by design: forbidding the
revert would make a correct reviewer comply and the story would commit without sign-off."
I do not think that holds any more. verify_review checks the spec's frontmatter status
first and returns retry there, before it ever reads the board. So a reviewer that withholds
sign-off through the spec already blocks the commit, and the board revert was never the
channel doing the work. Worth noting that #334 is still under Unreleased, so if this gets
[...]

Issue #261, opened 2026-07-22
Session read-back adopts another story's spec: a review that produced nothing is scored done


Polloinfilzato:
### Description

A session's read-back locates "the artifact this session produced" by scanning a **shared**
directory for the **most recently modified** qualifying `*.md`, with no constraint that it belongs
to the story being driven. When the session dies having produced nothing, a *different story's*
spec — written into that directory moments earlier by a concurrent run's merge-back — is adopted
as its result, and its `status: done` becomes this story's verdict.

Observed twice in one night, on two different call paths. On the review leg it merged unreviewed
code; on the dev leg it caused the review to be **skipped entirely**.

Unlike the rest of this failure family (#224, #127, #88, #160), this fails in the **unsafe**
direction: every prior member loses good work; this one lands unverified work.

### Evidence — a natural experiment with two controls

Run `20260722-025215-9ffc`, story `i-12`, `dev=claude` / `review=codex`, mux=tmux, macOS,
`isolation="worktree"`, four runs launched as separate processes.

(The `i-…` story keys are this project's own epic-naming convention, carried by a small local patch to the sprint-status key parser. It changes only which keys are *accepted* — artifacts are located by mtime and heading, never by key shape — so it plays no part in what follows.)

Three review sessions wedged identically that night. Only one was scored a pass:

| story | review-1 launched | died | pane log | hook events | qualifying foreign `.md` present | verdict |
|---|---|---|---|---|---|---|
| i-9 | 04:14:56 | 04:58:22 | 2 B | none | **no** | `crashed` ✅ |
| i-13 | 04:23:07 | 04:59:20 | 0 B | none | **no** | `crashed` ✅ |
| **i-12** | **04:55:25** | **05:36:41** | **0 B** | **none** | **yes — `spec-i-13-*.md`, mtime 05:15:27** | **`completed`** ❌ |

The only delta is the presence of a foreign artifact. (Control on the control: the `review-2` /
`review-3` passes of i-9 and i-13 — same codex adapter — wrote normal `SessionStart`+`Stop`
events and multi-MB logs, so the absent events are anomalous, not normal for this adapter.)

What the journal then recorded for i-12:

```jsonl
{"kind":"session-end","task_id":"i-12-…-review-1","status":"completed","tokens":null,"log_pos":0}
{"kind":"spec-reconcile-skipped-out-of-tree","story_key":"i-12-personality-projection",
[...]

PR #491, opened 2026-08-07
fix(prompt): name the sprint board's owner in dispatched prompts (#437)

pbean:
Refs #437 — phase 1 of 3. Adapts the design of #446 (didostap) onto current `main`; see "Why not #446's branch" below.

## The defect

Under the shipped default `scm.isolation = "none"`, `_post_dev_state_sync` advances `sprint-status.yaml` immediately after the dev session, because `verify.verify_dev` reads the board it writes. The story's single commit lands only at the end of the review loop (`verify.finalize_commit`). **Between those two points there is no commit at all**, so every session dispatched in that window — the review session above all — opens on an uncommitted, unattributed modification to `sprint-status.yaml`.

An external reporter hit exactly that. Their spec's Boundaries section said "Never edit … sprint-status.yaml"; the review session read the orchestrator's own write as a violation of that contract, filed it `[high]`, reverted the row `done → backlog`, and committed the revert. The #334 sign-off-regression gate then correctly escalated and paused the run — millions of weighted tokens on a story where both sessions agreed the work was finished.

The orchestrator already says "hands off, the orchestrator owns this" for the deferred-work ledger, one sentence earlier in the same prompt. There was nothing like it for the board.

## What changed

Two new seams beside `_operator_park_instruction`, injected the way that clause already is. Both are **bare sentences with no leading separator**, **backtick-free**, and neither contains the substring `append` (the review seam's whole-prompt ban, `tests/test_engine.py`).

- **`_sprint_board_instruction()`** — the prohibition. Rides all three `_generic_dev_prompt` legs and `_review_prompt`.

  > sprint-status.yaml is owned by the orchestrator: never write it, and never revert a change to it. A row at done or awaiting-operator is the orchestrator's own bookkeeping — not a defect to fix, and not proof that the work is verified.

  `awaiting-operator` is named because `_post_dev_state_sync` writes that row too; defending only `done` would leave a parked row reading as an unattributed edit. **"not proof that the work is verified"** is deliberate and stronger than #446's "not a defect": the row is written *before* `_verify_dev_artifacts` runs, and a repair session is dispatched precisely when that verification failed — so it opens on a red tree under a `done` row, and "not a defect" alone would leave it free to treat the row as evidence.

- **`_board_handback_redirect()`** — the way out, **review prompts only**.

  > If the story cannot be finished without a human decision, finalize the spec to status: blocked and say why. That is the hand-back channel; the board is not.

  The trigger is narrowed from #446's "if you judge the story genuinely unfinished": a review pass is itself a dev-primitive run whose job is to fix what it finds or defer it, so the broad form would hand it a run-halting early exit on cycle 1 of 3. `blocked` is the one status that both withholds the commit and reaches a human (`devcontract` synthesizes a `CRITICAL`; `decide_review_session` pauses). "say why" is load-bearing — the `## Auto Run Result` detail becomes the escalation text a human reads.

  The empty-when-no-board gate lives **inside this method**, not in `_review_prompt`, so the invariant is directly assertable (`StoriesEngine()._board_handback_redirect() == ""`) rather than an emergent property of the caller — and `_review_prompt` stays structurally identical to `_generic_dev_prompt` (filter-and-join, no special case).

- **Separators** move into the two callers, so one seam owns them all: an em dash after the bare story key (the one leg with no terminal punctuation), a plain space after a sentence. Both are plain spaces at the review seam because every clause ends in a full stop and an em dash would render `. — `. `_operator_park_instruction` loses its own leading `" — "`.

- **`StoriesEngine`** overrides `_sprint_board_instruction()` to `""` (no board in this mode), which drops the redirect with it and leaves that mode's review prompt byte-identical to today's. **`SweepEngine` needs no change**: it overrides `_dev_prompt`, so bundle dev prompts never reach the seam, but it does not override `_review_prompt`, so sweep reviews inherit both halves — correct, since a sweep runs inside a project whose board exists and is just as revertible.

- **Docs**: four places claimed the orchestrator does not write the board, contradicting `AGENTS.md` and `sprintstatus.py`'s own docstring. A docs defect propagated into a user's spec and then into a run failure. Fixed in `src/bmad_loop/__init__.py`, `docs/FEATURES.md` (×2), `docs/tui-guide.md`, scoped the way CodeRabbit pushed didostap to on #446: the orchestrator owns every write it makes **while a run is in flight**; your own `bmad-sprint-planning` / `bmad-correct-course` runs still write the board as they always did. Plus one new `### Failure handling & resilience` bullet and one CHANGELOG entry.

## Why not #446's branch

Its base is 126 commits behind and predates `4d90532` (#433), which **rewrote `_review_prompt`** to delete the affirmative "append them to the ledger as NEW entries" instruction — post-BMAD-METHOD#2640 the primitive files its own `defer` findings and `_harvest_spec_deferrals` picks them up, so the affirmative form double-files every finding. A verbatim rebase reverts that fix and fails main's guard test, which asserts `"append" not in prompt.lower()` over the whole prompt; #446's `LEDGER_SENTENCE` test constant pins the deleted wording verbatim. Its CHANGELOG hunk also amends a sentence `6615ad5` already removed. Main's ledger sentence is kept byte-for-byte here, and the #334 entry is untouched — line 127 already reads "`status: blocked` stays the hand-back channel", which is exactly what the redirect implements.

The design is didostap's; credited via `Co-authored-by`.

## Three things worth naming

**1. A deliberate sweep coverage gap.** `test_bundle_dev_prompt_has_no_board_clause_but_the_review_prompt_does` asserts on `_dev_prompt`'s **return value**. If someone later moved the injection into `_run_session`, that test would stay green while bundle prompts silently gained the clause. Closing it means asserting on the journaled `session-start` prompt; the builder is the correct layer today, so this is left open knowingly rather than missed.

**2. The park/redirect vocabulary overlap is unreachable, not merely unlikely.** The redirect's trigger ("cannot be finished without a human decision") overlaps the park clause's ("actions only a HUMAN can perform outside the repo"), and park says "Never use the blocked status" while the redirect says "status: blocked". They never co-occur in one prompt because each rides a different builder — and more strongly, **a parked story is never dispatched a review session at all**: `_review_and_commit` early-returns at `engine.py:1654-1655` on `_park_awaiting_operator`, which fires on spec frontmatter reading `awaiting-operator`; its docstring says "skipping the review loop" outright. The only residual is a story the dev pass should have parked and finalized `done` instead — and there the redirect opens no new path (`blocked` is the skill's native escape, and `devcontract` already synthesizes a CRITICAL from it) and displaces nothing better (park is dev-only, so a review session cannot park either way). Pause-and-reach-a-human beats a false-green `done`. `test_no_assembled_prompt_mixes_the_park_contract_with_the_blocked_redirect` pins the separation anyway, since nothing structural enforces it.

**3. One guard branch is unreachable, and its ablation says so.** `if tail else ""` on the **review** seam is live — `StoriesEngine` inherits `_review_prompt` and empties both clauses, so production reaches the empty-tail branch. On the **dev** seam it is not: `_generic_dev_prompt` has exactly one caller, both subclasses override `_dev_prompt`, and the plugin seam rewrites `proposed_prompt` rather than overriding the method — so on the one class that reaches it the prohibition is unconditionally non-empty. The guard is kept for symmetry and pinned with a monkeypatch. Ablation row A15b below records that making the dev separator unconditional **does not redden** with `[operator] enabled = false` alone; only emptying the board clause too reaches the branch.

## Tests

Nine new tests (7 `test_engine.py`, 1 `test_stories_engine.py`, 1 `test_sweep.py`). **Every presence, ordering and junction constant is a test-local literal, never a call to the builder under test** — `"" in s` is `True` and `s.index("")` is `0`, so `board in prompt`, `prompt.index(board) < …`, `f"{board} {park}" in prompt` and `prompt.endswith(park)` all pass vacuously under exactly the ablation they exist to catch. Three junction literals each span the **end** of one clause and the **start** of the next; that is the only shape that can detect a changed separator.

`tests/test_engine.py`'s `"append"` ban and the #261 `expected_spec` pin both pass **unmodified** — neither clause contains a path or a directory separator, so the fresh bare-key leg still does not pin and the review prompt still leads with the spec path.

### Ablation results — every row applied, run, restored

| # | Ablation | Named assertion | Result |
| --- | --- | --- | --- |
| A1 | prohibition body → `return ""` | `clause.strip()`; `BOARD_OWNED in prompt` on review + 3 dev legs | reddened (3 tests) |
| A2 | delete the no-board gate inside the redirect | `_board_handback_redirect() == ""`; stories `"status: blocked" not in` | reddened (2) |
| A3 | restore the affirmative-append ledger sentence | new `"append" not in prompt.lower()` **and** the pre-existing guard | reddened (2) |
| A4 | drop board from `_review_prompt`'s tuple | `BOARD_OWNED in prompt` | reddened |
| A5 | drop redirect from `_review_prompt`'s tuple | `BLOCKED_INVITE in prompt` | reddened |
| A6 | swap the two review halves | `index(BOARD_OWNED) < index(BLOCKED_INVITE)` | reddened |
| A7 | review ledger↔board seam → `f" — {tail}"` | `LEDGER_BOARD_JOIN in prompt` | reddened |
| A8 | review board↔redirect join → `" — ".join` | `BOARD_REDIRECT_JOIN in prompt` | reddened |
| A9 | drop board from `_generic_dev_prompt`'s tuple | `BOARD_OWNED in prompt`, all 3 legs | reddened (2) |
| A10 | restore the leading `" — "` inside the park clause | `BOARD_PARK_JOIN in prompt` | reddened |
| A11 | reorder board after park | `index(BOARD_OWNED) < index(PARK_HEAD)` | reddened |
| A12 | backtick-**wrap** a token in the board clause | `` "`" not in clause `` (+ `findall[-1] == feedback`) | reddened (2) |
| A12b | **odd stray** backtick in the board clause | `` "`" not in clause `` | reddened (2) |
| A13 | add the redirect to `_generic_dev_prompt`'s tuple | `"status: blocked" not in prompt` (first), then the count | reddened (2) |
| A14 | make the **review** separator unconditional | stories exact-equality + `.endswith(…resolution.)` | reddened (2) |
| A15 | make the **dev** separator unconditional | monkeypatched → `prompt == "/bmad-dev-auto 1-1-a"` | reddened |
| A15b | same, **without** the board monkeypatch (park off only) | `prompt == prompt.strip()` | **SURVIVED — expected, see note 3** |
| A16 | remove the `StoriesEngine` override | `_sprint_board_instruction() == ""` + review-prompt equality | reddened |
| A17 | append a third clause **after** the park clause | `prompt.endswith(PARK_TAIL)` | reddened |

Six of these (A8, A10, A12, A13, A14, A17) are corrections to an earlier table whose named assertions did **not** redden — four because the assertion was derived from the builder's own return value, one because the separator arithmetic yields period-space-**space**-em-dash rather than the `". — "` being watched for, and one because `index(board) < index(park)` catches a swap but not a third clause appended after park.

A12/A12b are worth calling out: an *odd* stray backtick leaves ``re.findall(r"`([^`]*)`", …)`` returning the same two matches, so the last-backticked-token assertion passes through it. Only a wrapped **pair** reddens that form — hence the parity-independent `` "`" not in clause `` assertion on both builders. (`tests/test_engine.py:6206` has the same blind spot today; not touched here.)

## Verification

- `uv run pytest -q -n auto` → 4459 passed, 25 skipped
- `uv run pyright` → 0 errors
- `trunk check` (no path filter, re-run after the last edit) → no issues
- `docs/FEATURES.md`'s table did **not** reflow: column 2 is padded to 279 and the replacement cell is 193 chars, so `git diff` shows exactly one changed table row.
- End-to-end sanity: read the real dispatched prompts out of a mock-adapter run's `journal.jsonl` `session-start` records (not `run --dry-run`, whose review preview is a placeholder that does not reproduce the appendix). Both assemble correctly — dev ends on the park contract, review ends on the redirect, and neither contains `. — `, a double space, or a trailing space.

## Follow-up (outstanding, hence `Refs` not `Closes`)

- **Phase 2**: #437's second ask — committing the board write together with the dev pass. Research says it is inert against this failure (the review's `{diff_output}` covers tracked *and untracked* changes since `baseline_revision`, so the hunk is in the reviewed diff either way) and actively harmful (`finalize_commit` never runs on the escalate path, and `rearm_escalation` advances `baseline_commit` to HEAD — baking a `chore(sprint)` commit permanently below the new baseline for a story that never went done). To be confirmed and recorded on the issue.
- **Phase 3**: plugin workflow sessions dispatched at `post_dev_phase`, `post_review_result` and `pre_commit_gate` run inside the exact window this hardens and receive neither clause. The `WORKFLOW_COMPLETION_CONTRACT` seam appends **after** the session-gate hooks, so the prohibition belongs there and gains a property this injection lacks — unstrippable by a plugin prompt rewrite.

Co-authored-by: didostap <dribnucko@gmail.com>

## Summary by CodeRabbit

* **Documentation**
  * Clarified that sprint-board updates are managed by the orchestrator during active runs.
  * Documented how development, review, sweep, stories, and plugin sessions handle board states.
  * Clarified dashboard behavior and sprint-board editing outside active runs.

* **Behavior Updates**
  * Development and review sessions are instructed not to modify or revert the sprint board.
  * Review sessions can report human-dependent blockers using a blocked status.
  * Stories mode continues to omit sprint-board ownership instructions.

* **Tests**
  * Added coverage for prompt guidance, board-state handling, and session-specific behavior.


pbean:
@codex review

PR #505, opened 2026-08-08
fix(recovery): bound the preserve-ref probe and refuse to reuse an occupied name

pbean:
Follow-up hardening on the preserve-ref probe that landed in #353. Not a regression in that PR — the data-loss fix is sound and stays as written; this closes the one gap the review flagged and merged over.

## The gap

`recovery_flow.py` probed for a free `refs/attempt-preserve-dirty/*` name with an uncapped loop:

```python
while verify.ref_exists(workspace.root, ref):
    ref = f"{base_ref}-r{serial}"
    serial += 1
```

That **terminates** — `serial` only climbs and the ref set is finite — but terminating is not the same as being **bounded**. The iteration count is whatever the namespace happens to hold, one git spawn apiece, on a path that only runs while a crashed attempt is being rolled back. `scm.preserve_keep` (default 20) prunes the namespace at run start, so it is short in practice and unbounded at `preserve_keep = 0` (never prune).

## The fix

Cap the scan at `PRESERVE_REF_PROBE_LIMIT` (100: the base name plus `-r2`..`-r100`) and **raise** on exhaustion rather than falling through to the last candidate — reusing an occupied name is the precise data loss #349 fixed.

The new `verify.PreserveRefExhaustedError` subclasses `GitError`, so the existing `except (verify.GitError, OSError)` at `recovery_flow.py:451` degrades it into the same journal-and-decide path as a snapshot that cannot be written. Preservation is observation: exhaustion must not crash the rollback. A distinct type so a caller can tell "the namespace is full" from "git said no" — those want opposite remedies — mirroring why `GitSpawnError` is its own type (`verify.py:79`).

Deliberately a module constant, not a policy field: it is a runaway backstop, not a tuning knob, and `scm.preserve_keep` already governs the namespace.

## What I did not change

- **The degradation path.** I checked rather than assumed: `GitSpawnError(GitError)` at `verify.py:79` means the handler already catches spawn and timeout faults. Correct as merged.
- **`ref_exists` folding "missing ref" and "malformed name" into `rc != 0 → absent`.** The names are built from `safe_ref_segment(run_id)` + hex + an int, so a malformed one is effectively unreachable, and if it happened the subsequent `update-ref` fails loudly into the same handler. Documented honestly in its own docstring; no change warranted.

## Verification

`tests/test_recovery_flow.py::test_ref_probe_is_bounded_and_refuses_to_reuse_an_occupied_name` answers `ref_exists` True for a few calls **past** the limit, then False. That shape matters: with the cap deleted the probe walks on to a free name and the test fails on the missing pause, instead of hanging the suite. It pins the exact bound (`len(probed) == PRESERVE_REF_PROBE_LIMIT`), asserts `snapshot_worktree` is never reached, and asserts the dirty work is not reset away.

Ablation performed per the repo rule — cap deleted, test failed with `DID NOT RAISE _Pause`, restored from the commit.

Gates: `uv run pytest -q -n auto` 4514 passed / 26 skipped · `uv run pyright` 0 errors · `trunk check` no issues.

## Summary by CodeRabbit

* **Bug Fixes**
  * Dirty worktree recovery snapshots now use collision-safe names, adding revision suffixes when needed.
  * Recovery stops safely after 100 occupied snapshot names instead of risking reuse or indefinite probing.
  * When no safe name is available, recovery pauses, records the failure and remediation, and leaves the working tree unchanged.

* **Documentation**
  * Added changelog coverage for collision-safe snapshot naming and exhaustion handling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -152,6 +152,14 @@ whose seams had diverged enough that several ports needed a different fix, and t
 
 ### Fixed
 
+- **A re-armed escalation no longer overwrites the previous attempt's dirty snapshot (#349).**
+  `refs/attempt-preserve-dirty/*` names were keyed on `task.attempt`, which `rearm_escalation`
+  resets to 0, so a post-resolve re-drive rolling back against the same baseline recomputed the
+  earlier rollback's refname and destroyed the only copy of that attempt's work. Probe for a free
+  name instead of trusting the counter, suffixing `-r2`, `-r3`, … The scan is bounded; exhausting
+  it refuses (`attempt-worktree-preserve-failed`, then the usual pause) rather than reusing an
+  occupied name. Prune the namespace or lower `scm.preserve_keep` if it ever fires.
+
 - **The auto-sweep child refuses config a session rewrote under the run (#461).** `policy.toml` and
   `profiles/*.toml` sit in the agent-writable workspace and reach host code execution — the
   `[verify] commands` run with `shell=True`, the resolved profile plus `adapter.extra_args` decide
--- a/src/bmad_loop/recovery_flow.py
+++ b/src/bmad_loop/recovery_flow.py
@@ -455,17 +455,33 @@ class RecoveryFlow:
         # namespace is small" into an enforced invariant, and exhausting it raises
         # rather than reusing the last candidate: falling through to an occupied
         # name is the precise data loss this probe exists to prevent (#349).
+        #
+        # The probe is check-then-write, not atomic: `snapshot_worktree` finishes on a
+        # plain two-arg `update-ref`, which overwrites whatever is there rather than
+        # failing if the name were taken in between. Safe here because each name has
+        # exactly one possible writer, on two independent grounds — the control loop
+        # is sequential (nothing in this path threads, so one run never has two
+        # rollbacks in flight), and the name is keyed on `run_id`, so separate runs
+        # address disjoint namespaces. Only two processes driving the SAME run could
+        # collide, and they would already be racing on run state, worktrees and mux
+        # sessions; the remedy for that is run-level exclusion, not a compare-and-swap
+        # on this one ref.
         base_ref = f"refs/attempt-preserve-dirty/{slug}-{baseline[:8]}-{task.attempt}"
         ref = base_ref
         serial = 2
         try:
             while verify.ref_exists(workspace.root, ref):
                 if serial > PRESERVE_REF_PROBE_LIMIT:
+                    # Remedy has to hold at BOTH ends of the preserve_keep range:
+                    # "lower it" is impossible at 0, which is precisely the setting
+                    # (pruning disabled) that lets this namespace grow far enough to
+                    # exhaust the probe in the first place.
                     raise verify.PreserveRefExhaustedError(
                         f"no free snapshot refname for {base_ref}: "
                         f"{PRESERVE_REF_PROBE_LIMIT} candidates through -r{serial - 1} "
-                        f"are all taken (prune refs/attempt-preserve-dirty/* or lower "
-                        f"scm.preserve_keep)"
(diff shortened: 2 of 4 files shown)


pbean:
@codex review

pbean:
Triaged all five bot findings against the code. Two were valid and are fixed in `1b86296`; two are declined with reasons; one is a false positive against this file's own convention.

Worth noting first: **the Greptile check is a fake green.** Its review body is `pbean has reached the 50-credit limit for trial accounts` — it never looked at this diff, though the check reports `pass`.

## Fixed

**1. The exhaustion remedy was wrong at the end of the range where it fires (Codex P2, `recovery_flow.py`).** The message said to *lower* `scm.preserve_keep`. But `0` means never prune (`policy.py:1253`), and that is precisely the setting that lets `refs/attempt-preserve-dirty/*` grow far enough to exhaust a 100-candidate probe — so the one configuration most likely to produce this error is the one that cannot go lower. The message now names a remedy that holds at both ends: prune the namespace, or set `preserve_keep` to a positive value below the limit.

**CodeRabbit asked for the opposite of this, and it is worth being explicit about the conflict.** Its `verify.py:791` comment asks me to rewrite the `PreserveRefExhaustedError` docstring to say "lower `scm.preserve_keep`" so it matches the message. That would have propagated the error rather than fixed it. I aligned them the other way and added a line to the docstring saying why "lower" is specifically not the remedy.

**2. The single-writer invariant is now documented (CodeRabbit Major, `recovery_flow.py:469`).** The premise is correct: `snapshot_worktree` ends on a plain two-arg `git update-ref` (`verify.py:785`), so probe-then-write is check-then-act, not atomic.

I declined the code change — the CAS — and took CodeRabbit's own third option, "serialize recovery writers and document that invariant," because the serialization already exists. Two independent grounds, both checked:

- **No intra-run concurrency.** `snapshot_worktree` has exactly one caller (`recovery_flow.py:472`), reached from the sequential control loop. Neither `engine.py` nor `recovery_flow.py` imports `threading`, spawns a `Thread`, or gathers tasks, so a single run never has two rollbacks in flight.
- **No inter-run collision.** The name is keyed on `safe_ref_segment(run_id)`, and `new_run_id()` (`runs.py:59`) is a timestamp plus `secrets.token_hex(2)`, so separate runs address disjoint namespaces.

The only writer that could collide is a second process driving the *same* `run_id`. That process is already racing on run state, worktrees and mux sessions — the ref is not the exposure, and the remedy is run-level exclusion rather than a compare-and-swap on this one ref. Adding a CAS here would buy nothing against the reachable failure modes while narrowing `snapshot_worktree`'s contract from "set this ref" to "create this ref" for every future caller.

## Declined

**3. Imperative changelog wording (CodeRabbit, `CHANGELOG.md:162`).** Checked against the file rather than the guideline in isolation. Across the `Unreleased` section, bolded entry headlines open with `A` 48 times, `The` 22 times and `An` 6 times; none open with an imperative verb. `A re-armed escalation no longer overwrites…` is the established pattern, and "Prevent re-armed escalations…" would be the outlier. The imperative guidance in AGENTS.md governs the terse body style, not these headlines — happy to be corrected if the intent was otherwise, but a one-entry change would only create inconsistency.

**4. Assert the literal `100` instead of the constant (CodeRabbit, `tests/test_recovery_flow.py:21`).** The contract under test is "the probe stops at the bound", not the bound's numeric value, and `assert len(probed) == PRESERVE_REF_PROBE_LIMIT` pins exactly that relationship — it is not vacuous, since the ablation (cap removed) fails it, and it would also fail if the loop drifted off the constant by one. Hardcoding `100` would additionally fail on any deliberate retune of a constant that is explicitly a runaway backstop, not a contract.

Gates after the fix: `uv run pytest -q -n auto` 4514 passed / 26 skipped · `uv run pyright` 0 errors · `trunk check` no issues.

pbean:
@codex review

Issue #508, opened 2026-08-09
env_fault classifier residuals from #324: an unguarded straddler drop, an unpinned evidence clamp, anchor-drift vacuity, and stale docs


pbean:
> **Status 2026-09-20 (v0.12.0).** Main is v0.12.0 (`b755a0fa`). `src/bmad_loop/adapters/env_fault.py` had no commits in v0.11.1..v0.12.0.
>
> **Fixed (all first shipped in 0.10.0):**
> - **Item 1** (trailing-newline defeat of the straddler guard) and its sibling nit (a complete first line dropped when the seek lands on a boundary) — `70369a95` (#510, 2026-08-08). The guard now keys on the raw byte before the window, not on truncation: `src/bmad_loop/adapters/env_fault.py:213` (`if straddles and any(lines[1:])`). Pinned by `test_classify_env_fault_keeps_a_boundary_aligned_first_line` (`tests/test_generic_tmux.py:6227`) and `test_classify_env_fault_scans_a_single_oversized_terminated_line[lf|cr]` (`:6278`).
> - Timeout re-export trap (found during the same pass, not an original item) — `d0341d85` (#510, 2026-08-08). The `generic.py` re-export block is marked READ-ONLY at `src/bmad_loop/adapters/generic.py:61`.
> - **Item 4** (stale `.log`-is-the-server-sink comments) and **item 5** (stale "Seeded only for `claude`" docs, missing CHANGELOG entry) — `cb3fa7f3` (#510, 2026-08-08). The cited wording is gone from `docs/adapter-authoring-guide.md` and `docs/FEATURES.md`; `docs/adapter-authoring-guide.md:502` and `docs/FEATURES.md:76` now name the per-adapter log (`ENV_FAULT_LOG_SUFFIX`, `.server.out` for opencode-http). Module docstring `src/bmad_loop/adapters/env_fault.py:12-16`, `src/bmad_loop/adapters/profile.py:155-156` and `tests/test_env_fault_patterns.py:19-20` all name `.server.out` as the model-free sink. The 0.10.0 CHANGELOG carries the #323 entry.
>
> **Surviving on main:**
[...]

Issue #507, opened 2026-08-09
claude.toml's env_fault pattern false-positives on ordinary prose — including this repo's own CHANGELOG and docs


pbean:
Split out from #323, which covers the *miss* half (no profile classifies provider quota errors). This is the inverse defect, and it blocks the fix #323 prescribes.

Found while reviewing #324, which pins it as a strict `xfail` rather than leaving it undocumented.

## The defect

`src/bmad_loop/data/profiles/claude.toml:28` ships #194's original pattern:

```
API Error.*(Unable to connect|Connection ?(error|refused|reset|timed ?out)|ConnectionRefused|ECONNREFUSED|ETIMEDOUT|ENOTFOUND|EAI_AGAIN)
```

The unbounded `.*` between the anchor and the cause makes it match any line that merely *mentions* both. The claude adapter scans a tmux pane capture, which contains **the model's own output** — so a story that writes *about* provider errors trips it.

Compiling the shipped pattern and scanning every tracked file line-by-line yields **37 matching lines**. The non-test ones:

- `CHANGELOG.md:81` — ``connection stayed alive but idle, printing `API Error: Unable to connect …` until the session``
- `docs/FEATURES.md:70` — the `#194` bullet describing an `API Error … Connection refused`-class transport failure
- `src/bmad_loop/adapters/profile.py:107` — ``an "API Error … Connection refused" or a "usage limit``
- `src/bmad_loop/data/profiles/claude.toml:28` — **the pattern's own line matches itself**

The other 33 are in `tests/test_engine.py`, `test_generic_tmux.py`, `test_escalation.py`, `test_sweep.py` and `test_plugin_workflows.py` — i.e. exactly the text a dev session prints via pytest output, `grep`, or a diff hunk. A bmad-loop session working on bmad-loop is exposed on its own repo.

## Impact

`EnvFaultMixin._classify_env_fault` (`src/bmad_loop/adapters/env_fault.py:157`) only scans a session whose status is `timeout`/`stalled`/`crashed` with `result_json is None`. So this does **not** affect healthy completed sessions.

The real harm is the other direction: a **genuine** timeout is misread as an environment fault, which `escalation.decide_dev` maps to `Action.PAUSE` with the budget re-armed instead of charging an attempt. That masks the real timeout and halts the run for an operator. A model can also reach it deliberately — emitting `API Error: Connection refused` launders its own failure into a non-charged pause.

## Why this blocks #323

#323 prescribes:

[...]

Issue #323, opened 2026-07-27
Provider usage-limit / quota errors are not classified as environment faults — the loop burns retries and defers healthy stories


jackmcintyre:
**bmad-loop version:** 0.9.0 (repo `e955c17`)
**Adapter:** `opencode-http` (`zai-coding-plan/glm-5.2` for dev; `claude`/`opus` for review)
**OS:** macOS (darwin 25.5.0) · **opencode:** 1.18.4
**Impact:** 3 stories deferred, ~12 hours of wall-clock burned, on specs that were fine.

---

## Summary

When the dev provider returns a hard **usage-limit / quota** error, bmad-loop does not recognise it as an environment fault. It records `env_fault: false`, spends `max_dev_attempts` on the story, defers it, and moves to the next story — which hits the identical wall. In my run this converted a single 5-hour provider quota window into **three deferred stories and seven dead sessions**, each of which sat for its full `session_timeout_min` consuming **zero tokens**.

The story specs were not at fault. The first story of the same run completed and merged cleanly on the same adapter and model; it was simply the one that exhausted the quota.

## What happened

Run of 5 stories. Story 1 succeeded, burned 2,517,158 raw tokens in its dev session, and exhausted the provider's 5-hour rolling budget. Everything after it failed:

| session | status | tokens | wall-clock |
|---|---|---|---|
| story-1 dev-1 | completed | 2,517,158 | — (merged fine) |
| story-2 dev-1 | timeout | 649,118 | 90.1 min |
| story-2 dev-2 | timeout | 517,249 | 90.0 min |
| story-3 dev-1 | timeout | **0** | 103.2 min |
| story-3 dev-2 | timeout | **0** | 110.0 min |
| story-4 dev-1 | timeout | **0** | 124.4 min |
| story-4 dev-2 | timeout | **0** | 105.2 min |
| story-5 dev-1 | timeout | 26,778 | 100.2 min |

Every one of the 22 provider errors in the run logs is the same line:

```
AI_APICallError: Usage limit reached for 5 hour. Your limit will reset at 2026-07-26 22:49:46
```

Later sessions degraded to a second form — the provider dropping the connection rather than returning a clean 429:

```
AI_APICallError: Cannot connect to API: The socket connection was closed unexpectedly.
```

A representative dead session, start to finish (~6 KB of log for a 90-minute session):

```
15:45:39.732  stream providerID=zai-coding-plan modelID=glm-5.2
[...]

PR #509, opened 2026-08-09
Revert #324 to re-land it behind the @codex review gate

pbean:
Reverts merge commit `865b7597` (#324), restoring the pre-#324 state of the env-fault classifier.

**This is a process revert, not a defect revert.** #324 reviewed clean and I am not aware of anything wrong with the work. We are adopting a merge gate — a PR must clear an `@codex review` pass with **no outstanding review comments** before it lands — and that gate has to be demonstrated on an *open* PR. GitHub does not allow reopening a merged PR, so taking the change back off `main` is the only way to put #324 back into a reviewable state.

`@jackmcintyre` — apologies for the churn; this is on our process, not on your patch. The branch `fix/env-fault-quota-classify` still exists on your fork at `64efb7ef`, so the work is intact and will go back up as a fresh PR against a clean `main`.

## Verification

The revert restores the tree exactly: `git diff b3a75a8 HEAD` is empty, i.e. byte-identical to the commit `main` sat at immediately before #324 merged — a state that was already green in CI. Ten files, all from #324:

- `src/bmad_loop/adapters/env_fault.py` and `tests/test_env_fault_patterns.py` removed
- `base.py`, `generic.py`, `opencode_http.py`, `profile.py`, `opencode.toml`, `test_generic_tmux.py`, `test_opencode_http.py`, `test_profile.py` restored

## Tracker state

- #507 (claude.toml anchor false-positives on prose) is **independent of #324** and stands on its own — it describes a pre-existing defect in a file #324 never touched.
- #508 (classifier residuals) describes code that this revert removes from `main`. Leaving it open: it is the review record for the re-landing PR, and every item still applies to the branch.
- #323 stays open and unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of local environment failures from session logs, including provider connection and quota issues.
  * Added safer handling for truncated logs, ANSI formatting, incomplete lines, and oversized evidence.
  * Prevented completed or healthy sessions from being incorrectly classified as environment failures.

* **Configuration**
  * Updated profile behavior so environment-fault patterns are enabled only where supported, with clearer matching guidance.
  * Removed environment-fault classification from OpenCode server logs.


pbean:
@codex review

pbean:
## Triage of the automated review on this PR

All four CodeRabbit findings were validated empirically in two throwaway worktrees — one at this PR's head (`9c6391e8`, pre-#324 code) and one at `865b7597` (main with #324 applied). **None is actionable on this PR.** Detail below, because two of them are wrong in ways worth recording.

The governing fact: `git diff b3a75a8 9c6391e8` is **empty**. This PR restores the pre-#324 tree byte-identically and introduces zero new lines. So no finding here can have been *introduced* by it — each is either lost-relative-to-#324 or pre-existing in both. CodeRabbit's walkthrough reads the revert as a forward change ("classification moved from `EnvFaultMixin` into `GenericAdapter`... OpenCode HTTP integration removed"), which is why it is reviewing restored code as if newly written.

### F1 — `generic.py:979`, partial first tail line — declined, returns when #324 re-lands

Genuine defect, reproduced on this branch:

```
env_fault=True   evidence='JJJJ…JJJAPI Error: Connection refused'
```

But it is precisely what #324 fixed ("fix A") at `env_fault.py:185` and `:192-205`. Re-implementing it here would mean rebuilding #324 inside its own revert.

⚠️ **Do not apply the suggested diff or the Autofix.** The proposal drops `lines[1:]` unconditionally; #324 guards it with `and len(lines) > 1`. Applying CodeRabbit's version verbatim fails an existing test:

```
-        if truncated and len(lines) > 1:
+        if truncated:                      # CodeRabbit's literal proposal
→ FAILED tests/test_generic_tmux.py::test_classify_env_fault_scans_a_truncated_window_with_no_newline
```

Unguarded, a 64 KiB window containing no newline scans nothing — trading a cosmetic half-line for a missed outage. If this is ever re-litigated, #324's shape is the correct one.

(Note for the record: #324's guard has its own narrower hole — a trailing newline yields `[fragment, ""]`, so `len(lines) > 1` passes and the drop leaves nothing. That is tracked in #508, and the remedy is `any(lines[1:])`, not removing the guard.)

### F2 — `generic.py:981`, no scan-wide deadline — valid, pre-existing in both, tracked in #508

Confirmed, and it is **not** something #324 fixed — `grep -n "monotonic\|deadline\|time\."` over #324's `env_fault.py` returns nothing. Both versions pass `timeout=` per `pat.search(line, ...)` with no scan-wide bound. Measured over 20 lines of a catastrophic pattern:

| | per-line | total | exceeded 2.0 s? |
|---|---|---|---|
| this branch | 0.267 s | **5.188 s** | yes |
| #324 | 0.254 s | **5.141 s** | yes |

#324 documents the limit explicitly (`env_fault.py:57-65`: *"it is per-search, so the worst case for a scan is lines × patterns × this, not this"*). Recorded in #508 rather than fixed here — a mechanical revert is the wrong place to add a new bound.

### F3 — `opencode_http.py:369`, restore the OpenCode classifier — declined, this is the revert's intent

Verified line-for-line: `wt-revert:369` is `class OpencodeHttpAdapter(_ResultFileMixin, CodingCLIAdapter)`; #324's is `(_ResultFileMixin, EnvFaultMixin, CodingCLIAdapter)` with `ENV_FAULT_LOG_SUFFIX = ".server.out"`. The `base.py` docs, the `opencode.toml` patterns and the stale-log `unlink` are all likewise pure #324 content. This finding is a description of undoing this PR.

The revert was also checked for internal consistency: no dangling `EnvFaultMixin` references, and `opencode.toml` correctly carries no patterns.

### F4 — `test_generic_tmux.py:2495-2503`, vacuous inert test — valid, pre-existing in both, tracked in #508

CodeRabbit is right that the test does not pin its guard, and the finding is actually *stronger* than claimed — with the `not self._env_fault_patterns` guard deleted, not just that test but the **entire suite** stays green on both sides:

```
this branch, guard ablated: 4566 passed, 32 skipped
#324,        guard ablated: 4707 passed, 32 skipped, 5 xfailed
```

Not a revert artifact: the revert only changed how the empty-pattern profile is obtained, and the cited assertions are unchanged context lines — hence CodeRabbit filing it outside-diff.

But the proposed remedy does not work. The guard is **result-equivalent to no guard**: with empty patterns `_env_fault_evidence` returns `None` on its own, so `env_fault=False` with no lifecycle line either way. Its only effect is skipping one log read.

```
_env_fault_evidence() with EMPTY patterns -> None
log file opened by that call?             -> True
```

So no strengthening of the *result* assertions — including the suggested stub — can make the test pin the guard. A non-vacuous test would have to assert the log is **never opened**. It is a perf short-circuit, not a safety gate, which is why this is an issue note rather than a rewrite.

### Outcome

No code changes on this PR. F2 and F4 are appended to #508 with their corrected scope; F1 and F3 return on their own when #324 re-lands. Codex reviewed this PR at `9c6391e8` and found no issues.

Issue #419, opened 2026-08-01
Untagged sessions are weak ownership: they leak once their run dir is gone, and can be pruned by the wrong project on a run-id collision (the fallback behind #320)


dracic:
Split out of #320 (PR #418) by review, deliberately not folded into that fix: the resolution needs an intent decision, not a wider diff.

## The gap

#418 gates the psmux session project tag on `_transportable()`: a value that cannot survive psmux's CLI→server control line is refused rather than stored corrupted, leaving the option unset. That is the correct degradation, because `runs.prunable_sessions` falls back to the run dir for an untagged session — it claims our own dead runs and skips foreign ones.

**That fallback holds only while the run dir exists.** `delete_run` (`runs.py:512`) and `archive_run` (`runs.py:526`) both `rmtree` it, and neither kills an already-orphaned session. So for a project whose tag is untransportable:

1. every agent session is created untagged;
2. a crash (or any run whose session outlives its engine) leaves the session behind;
3. the operator runs `clean` / archives the run, removing the directory;
4. every later prune reads the session as untagged **with no run dir under this project**, hits `runs.py:351-352`'s `elif not is_run(run_dir): continue`, and skips it — from every project, including its own.

Result: one leaked psmux session, and its server, per run, for the life of the machine.

Note the asymmetry this exposes. A *tagged* orphan whose run dir is gone **is** prunable — that is the `orphan-1` case in `test_prunable_sessions_partitions`. The tag is the only ownership proof that survives run-dir removal, so the untagged branch is strictly weaker by design.

## Why this is not just theoretical, and not just a psmux problem

The refusal shapes look exotic until you filter them through `Path.resolve()` on Windows:

- `"` is illegal in Windows filenames — unreachable in a resolved project path.
- A mid-path `\\` and a trailing `\` are normalized away by `resolve()`.
- A standalone `;` token needs a directory literally named `;`.

The one shape that survives all of that is a **spaced UNC path** — `\\srv\share name\proj`. That is an ordinary corporate setup, not a corner case, and there every session for the project goes untagged.

Worth stating plainly: this is **not a regression** from #418. Before the gate, the same path stored a corrupted tag, which `prunable_sessions` skipped for exactly as long. #418 does not make it worse; it just makes the ceiling visible and worth naming.

## Candidate directions

[...]

Issue #335, opened 2026-07-27
awaiting-operator part 4: [operator] on_review_demotion = park (the allowlisted done -> awaiting-operator writer)


pbean:
> **Status 2026-09-20 (v0.12.0).** Parts 1–3 landed and shipped in 0.10.0 (`CHANGELOG.md:3422`): #347 `d53ba639` vocabulary — `Phase.AWAITING_OPERATOR` in `TERMINAL_PHASES` (`src/bmad_loop/model.py:44`, `:50`), the one `COMMITTING → AWAITING_OPERATOR` edge (`src/bmad_loop/statemachine.py:38`), the token before `done` in `STATUS_ORDER` and out of `ACTIONABLE_STATUSES` (`src/bmad_loop/sprintstatus.py:52`, `:64`); #348 `8a3cf914` entry — `_park_awaiting_operator` rides `COMMITTING` → `_finalize_commit_phase` (`src/bmad_loop/engine.py:3595`, `:3703`), `_operator_park_enabled` opts stories/sweep out (`src/bmad_loop/engine.py:4039`), `OperatorPolicy.enabled` default-on (`src/bmad_loop/policy.py:314`, `src/bmad_loop/data/settings/core.toml:453`); #355 `e425ac29` exit — `cmd_confirm` (`src/bmad_loop/cli.py:3805`), `operatoractions.record_park` (`src/bmad_loop/operatoractions.py:136`), four `operator.*` validate ids (`src/bmad_loop/checks.py:93-96`). Committed park records followed via #356 (PRs #361, #362).
>
> 0.12.0 hardened the park proof-of-work skip (`CHANGELOG.md:2104`, related to #676): the waiver now requires the current session's own `park_asserted` result marker (`src/bmad_loop/devcontract.py:510`, `src/bmad_loop/verify.py:8322`) and journals `park-proof-of-work-skipped`; retained frontmatter or a re-armed spec no longer inherits it.
>
[...]

PR #541, opened 2026-08-11
test: cover envvars registry + TUI hard-stop/archive paths; fix inf session-timeout override

pbean:
Test-gap backfill from an audit. Three independent gaps plus one unused dependency. No `src/` behavior changes — the only non-test edits are `pyproject.toml`/`uv.lock` (dependency removal) and `CHANGELOG.md`.

## 1. `tests/test_envvars.py` — new

`envvars.py` is the module AGENTS.md names as the one place a core `BMAD_LOOP_*` var is defined, and no test named it (`grep envvars tests/` → 0). Six tests / 13 cases pin the documented contract: `session_timeout_s()` ignores an unparseable, empty, zero or negative override; `mux_backend()` / `process_host()` pass their value back verbatim, so the downstream registry is what raises on an unknown name — a reader that swallowed a bad name would turn a loud misconfiguration into a silent auto-select.

**Coverage here is layered, not virgin.** `test_engine.py::test_session_timeout_s_env_override` already pinned the same rejection rows one layer up, through `Engine._session_timeout_s`. That is a different claim (does the *policy default survive*?) from this one (what does the *parse return*?), so both stay, and both now state their contract parity — a behavior change lands in both or records the divergence.

## 2. Env-registry scan guard — `tests/test_portability_guard.py`

AGENTS.md's second env clause — "plugin-owned env-var families stay with their plugin" — was documented but unenforced. One new scan in the file's existing single AST pass fails any `BMAD_LOOP_*` read taken from the process environment outside `envvars.py`, the two stdlib-only hook relays, and the seven Unity helper scripts.

- **Reads only.** `engine.py`, `resolve.py`, `probe.py`, `plugins/bus.py` and `unity_plugin.py` *build* a `BMAD_LOOP_*` dict to inject into a child session — the producing side, not a second source of truth. `gates.py`'s `BMAD_LOOP_NOTIFY_*` are read by osascript/PowerShell, not by us.
- **The alias arm is load-bearing.** `envvars.py` reads exclusively through named constants (`os.environ.get(MUX_BACKEND)`), so a literal-only matcher would never see the registry and its allowlist entry would be dead code.
- **No violation exists.** All 52 reads live in the 10 allowlisted files; every `os.environ` read elsewhere in `src/` targets a non-`BMAD_LOOP_*` var (`SystemRoot`, `XDG_CONFIG_HOME`, `PATHEXT`, `TMUX`, `TEXTUAL_*`). The guard's value is entirely prospective, which is what the ablations grade.

Also allowlisted, and not predicted when the work was scoped: `data/bmad_loop_probe_hook.py`, the same stdlib-only shape as `bmad_loop_hook.py`.

## 3. TUI hard-stop / archive Pilot tests — `tests/test_tui_app.py`

`x` and `A` had no Pilot coverage. Added, mirroring the graceful-stop section's helpers:

- `x` on a live run → confirm → both halves fire (`runs.stop_run` **and** `launch.kill_ctl_window`).
- `x` on a run that is not *provably* alive → warning, no modal, neither helper called. Parametrized over `dead` **and** `unknown`: the gate is `== "alive"`, deliberately stricter than `S`'s `!= "dead"`, because killing the agent window on an unidentifiable pid is not recoverable the way a control-file request is.
- `A` archives + forgets; `A` on a live run refuses. Note this gate is **asymmetric with `x`** — archive blocks only `alive`, letting `unknown` through into a confirm that carries a warning. It is also only the first of two refusals; `_archive_run_worker` catches the session-keyed `runs.LiveSessionError` (#419) behind it.
- `shift+tab` reverse-cycles the resize ring, mirroring the forward-tab test beat for beat.

`action_stop_run` also has a `_mux_missing()` pre-gate ahead of the liveness check, so these tests pin `launch.mux_available` — left unpinned they would silently depend on tmux being on the dev box's PATH.

## 4. `pytest-rerunfailures` removed

Never invoked: no `--reruns`, no `addopts`, no `flaky` marker in the repo or CI. A test-retry plugin sitting in the environment is an invitation to paper over a real flake. Removed from the `dev` group and relocked with `uv lock` (uv removed v16.3; no other lock churn).

## Ablations

Repo rule — every negative assertion was ablated against a `cp` backup (never `git checkout`), confirmed to FAIL, and restored byte-identical (md5 verified). Dated notes are recorded in the test files.

| Gate deleted | Result |
|---|---|
| `session_timeout_s` `try/except ValueError` only | 2 failed — exactly the unparseable rows |
| `session_timeout_s` `> 0` guard only | 5 failed — exactly the numeric rows |
| `session_timeout_s` `raw is None` early-out | 1 failed, on `TypeError` the `except ValueError` does *not* catch |
| `mux_backend` made "helpful" (filter unknown → None) | 1 failed, on the unregistered-name row alone |
| Scan-guard allowlist emptied | fails listing all 52 real reads (detection fires) |
| Inline read planted in a non-allowlisted core module | fails naming the file:line — verified twice, independently, in `verify.py` (`os.environ.get`) and `engine.py` (`os.getenv`) |
| Same read spelled through a module constant | fails — the alias arm is live |
| Name written into a docstring instead | stays green — prose is not a read |
| TUI `x` liveness gate | 2 failed — both `dead` and `unknown` rows, nothing else |
| TUI archive `alive` gate | 1 failed |

The two halves of `session_timeout_s`'s rejection hold **disjoint** row sets, so a combined ablation would have reddened everything and graded neither. No test was vacuous on first write.

## Verification

`uv run pytest -q -n auto` → **5003 passed, 39 skipped, 5 xfailed** · `uv run pyright` → **0 errors** · `trunk check` → **no issues**.

## Flagged, not fixed

`BMAD_LOOP_SESSION_TIMEOUT_S=inf` is **accepted** (verified: returns `inf`) — `inf > 0` is True, so no guard rejects it, and the result is an effectively unbounded session wall-clock budget. It does not violate the docstring's promise that a bad override can never silently *shorten* a run, but lengthening without bound is unguarded. Deliberately left unasserted: pinning it would freeze what may be an unintended affordance. Worth a tracker issue rather than a test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Session timeout settings now ignore invalid, non-positive, and non-finite values while accepting valid finite values.
  * Improved safeguards for environment-variable handling.

* **User Interface**
  * Hard-stop and archive controls now provide stronger confirmations and run-state safeguards.
  * Improved reverse navigation and pane resizing behavior.

* **Documentation**
  * Clarified session-timeout configuration for values such as `inf` and very large numbers.


pbean:
## Update: the `inf` gap flagged below is now fixed in this PR (`ea4c2e7`)

The PR was originally test-only and flagged `BMAD_LOOP_SESSION_TIMEOUT_S=inf` as an unasserted gap. It is now fixed here rather than deferred, so this PR carries one `src/` behavior change.

### The bug

The guard was one-sided — `value > 0` — and `float()` accepts `inf`, `1e999`, `Infinity` and `INF`, all of which pass it. Both adapters build their deadlines as `time.monotonic() + spec.timeout_s` (`generic.py:527,533`, `opencode_http.py:1007,1013`), so a non-finite budget produced a deadline that can never expire.

That budget is not one bound among several: it is the outer bound the stall-grace and wake-nudge windows explicitly defer to ("Bounded overall by `spec.timeout_s`"), and those re-arm on every event. The override silently removed the only backstop an unattended run has.

### The fix

`session_timeout_s()` now requires a finite positive value and otherwise falls back to `limits.session_timeout_min × 60` — exactly as it already did for `0`, `-1` and unparseable input.

A large **finite** value is still honoured, and `1e308` is pinned as an accepted row so that boundary is a reviewable decision rather than an accident. A duration, however unwise, is still a duration; `inf` is not one. Drawing the line anywhere else would mean inventing a ceiling, which is a policy number this module has no business choosing.

### Fix surface — why the reader is the complete fix point

`engine.py:4039` is the sole consumer, and both adapters derive their deadlines from `spec.timeout_s`, so fixing the reader covers every path.

Deliberately **not** touched: `policy.py`'s `_opt_grace` and `session_timeout_min` have the same one-sided shape, but they are policy-file input that *raises* `PolicyError` rather than silently ignoring, and they bound inner windows rather than the outer deadline. Different surface, different semantics — widening to them would be scope creep.

### Ablation — one result corrected a wrong assumption

| Gate deleted | Rows reddened |
|---|---|
| `try/except ValueError` | 2 — exactly the unparseable rows |
| `value <= 0` half | 4 — exactly the zero/negative rows |
| `not math.isfinite(value)` half | **5** — the four non-finite spellings **and `nan`** |

That last row was the surprise, and it is load-bearing. I had reasoned `nan` would be held by both halves. It is not: the old code was `return value if value > 0 else None`, where `nan > 0` is False; the new guard is spelled `value <= 0`, and `nan <= 0` is **also** False. So `nan` no longer falls out of the comparison and is now held by the finiteness check alone. Same observable result, different gate — recorded in the test so nobody "simplifies" the finiteness half away on the theory that `> 0` already covers `nan`.

Measured, not reasoned. The ablation is what caught my error.

### Also updated

- `test_engine.py`'s seam test gains the `inf`/`1e999` rows — honouring the contract-parity note added in the first commit, which says a behavior change lands in both or records the divergence.
- `README.md`'s env table stated the old contract verbatim ("A non-positive or unparseable value is ignored") and would have been left wrong.

### Verification

`uv run pytest -q -n auto` → **5008 passed**, 39 skipped, 5 xfailed (+5 cases) · `uv run pyright` → **0 errors** · `trunk check` → **no issues**.

End-to-end through the engine seam: `inf`, `1e999`, `Infinity`, `INF`, `nan`, `-inf`, `0`, `-1`, `nonsense`, `""` all fall back to the 5400s policy default; `3`, `0.5` and `1e308` win as overrides.

pbean:
## CodeRabbit review addressed (`71e451f`)

Both findings triaged against the code. One accepted in full, one in part with the other half declined on evidence.

---

### ✅ Finding 2 — NaN gate descriptions — **VALID, fixed**

Correct, and a good catch. Two annotations still described the **pre-fix** gate and contradicted the ablation block sitting a few lines below them:

| Location | Said | Actually |
|---|---|---|
| module docstring (L15) | nan is "rejected by the comparison" | true of the old `value > 0`; false now |
| parametrize row (L82) | held "by the finiteness check AND by `<= 0` being False" | incoherent — `<= 0` evaluating False is *why* that half does **not** reject it |

Both now state what was measured: `nan <= 0` is `False`, so the finiteness check rejects nan **alone**.

Worth noting why this slipped through: this is the *exact* error the single-gate ablation caught while making the fix — I had predicted 4 reddened rows and measured 5. I corrected the ablation comment and missed the two other places carrying the same stale claim. The bot's scripted reproduction found them independently.

### 🟡 Finding 1 — CHANGELOG entries — **partially valid, partially declined**

**Accepted (length).** Measured across all 122 entries under `Unreleased`: median **6** non-blank lines, range 2–16. Mine were a median of **10** — genuinely the longest cluster. Now trimmed to a median of **6**, dropping restatement and file-path detail while keeping the causal content (for the `inf` entry, *why* a non-finite budget is not merely a large one: it is the outer bound every stall-grace and wake-nudge window defers to).

**Declined (imperative headings).** The suggestion was to make the declarative headings imperative. The file's own convention is predominantly declarative:

> - **A park travels with its story's commit, so `bmad-loop confirm` works from any clone (#356).**
> - **Deferred review findings are harvested from spec frontmatter (#433).**
> - **Stories can park at `awaiting-operator` (#335, part 2 of 4).**
> - **A deferred-work entry can block a story: `gate:`.**

Rewriting these four as imperatives would make them inconsistent with every neighbouring entry. Where the guideline's literal reading and the established house form disagree, the house form wins — the guideline is a summary of the convention, not a rule the convention violates.

---

No `src/` change in this commit. `uv run pytest -q -n auto` → **5008 passed** · `uv run pyright` → **0 errors** · `trunk check` → **no issues**.

pbean:
@coderabbitai full review

pbean:
@codex review

pbean:
## Codex P2 addressed (`53bb29f`) — valid, and it found a hole the guard shipped with

> Resolve imported constants in the env-read guard

**Confirmed valid by planting both shapes into `verify.py`.** Each passed **green** before this commit, so the guard was asserting an invariant it did not enforce:

```python
from . import envvars as _ev
_LEAK1 = os.environ.get(_ev.MUX_BACKEND)     # qualified attribute -> MISSED
```
```python
from .envvars import MUX_BACKEND
_LEAK2 = os.environ.get(MUX_BACKEND)         # imported constant   -> MISSED
```

`_env_read_key` resolved only string literals and names bound to a literal in the **same** module, so borrowing the registry's constant from anywhere else bypassed detection.

The severity framing in the finding is right and worth restating: this is not an exotic shape, it is **the shape a well-meaning change actually takes**. Reusing `envvars.MUX_BACKEND` while calling `os.environ.get` yourself looks tidier than an inline literal. The two spellings the guard caught were the ones a careless author writes; the one it missed was the one a careful author writes.

### Fix

`REGISTRY_NAMES`, derived from `vars(envvars)` rather than hardcoded — register a fourth var in `envvars.py` and reads through it resolve with no edit to the test. A same-module binding still wins over a registry name it shadows.

The attribute arm matches on `.attr` alone rather than proving the object is the registry module: `import bmad_loop.envvars as ev`, `from . import envvars` and a rebound alias all spell it differently, and resolving that statically costs more than it buys. A false positive is a review prompt on a line that already reads like an env lookup — the direction a tripwire should fail in.

### Detection matrix, re-verified end to end

| Shape | Before | After |
|---|---|---|
| `os.environ.get("BMAD_LOOP_X")` | caught | caught |
| same-module constant | caught | caught |
| `os.environ.get(_ev.MUX_BACKEND)` | **missed** | **caught** |
| `os.environ.get(MUX_BACKEND)` imported | **missed** | **caught** |
| `os.getenv(...)` | caught | caught |
| `os.environ[...]` | caught | caught |
| prose mention (control) | not flagged | not flagged |

Baseline stays green, so matching by attribute name introduced no false positive anywhere in `src/`.

---

### Separately — my own finding while building that matrix

The docstring exclusion inside `_env_read_key` was **unreachable**, and the ablation I recorded for it was **vacuous**.

That helper only ever inspects a key position (a call's first arg, a subscript's slice). A docstring is a standalone `Expr` statement and can never appear there. Counted across the whole tree: **zero** key-position nodes are also docstring nodes. So the "prose stays green" row passes because a prose mention builds no node at all — not because an exclusion fired. It would have stayed green with the entire scan deleted.

Removed the dead parameter and relabelled that row a **control**, with an explicit note that it grades nothing, so nobody later cites it as evidence the way I did.

---

`uv run pytest -q -n auto` → **5008 passed** · `uv run pyright` → **0 errors** · `trunk check` → **no issues**. No `src/` change.

pbean:
@codex review

pbean:
## Codex P2 round 2 addressed (`5b00872`) — valid

> Inspect keyword environment keys

**Confirmed.** The scan required a positional arg, so a keyword key left `node.args` empty and slipped through. Planted in `verify.py`, both passed green before the fix:

```python
_KW1 = os.getenv(key="BMAD_LOOP_KW")
_KW2 = os.environ.get(key="BMAD_LOOP_KW")
```

**I checked the premise against the live interpreter first, because the obvious intuition is wrong here.** `dict.get` is positional-only in C and *would* raise on `key=`. But `os.environ` is `os._Environ`, a Python-level `MutableMapping`, so `get`/`pop`/`setdefault` are the ABC's plain-Python defs and bind `key=` happily; `os.getenv(key=...)` binds for the same reason. All four returned a value rather than raising — so all four were real holes, not syntax that cannot occur. Had the dict intuition held, the finding would have been invalid.

### Fix

`_env_call_key_node` — first positional arg, else the `key=` keyword.

| Shape | Before | After |
|---|---|---|
| `os.getenv(key=...)` | **missed** | **caught** |
| `os.environ.get(key=...)` | **missed** | **caught** |
| `os.environ.pop(key=...)` | **missed** | **caught** |
| `os.environ.setdefault(key=...)` | **missed** | **caught** |
| keyword form + imported registry constant | **missed** | **caught** |
| `get("LITERAL", default=...)` mixed | caught | caught |

Baseline stays green — no false positive anywhere in `src/`.

`uv run pytest -q -n auto` → **5008 passed** · `uv run pyright` → **0 errors** · `trunk check` → **no issues**. No `src/` change.

pbean:
@codex review

pbean:
## Codex P2 round 3 addressed (`3ec0b6d`) — valid, and I swept the rest rather than wait

> Detect membership-based environment reads

**Valid.** `"BMAD_LOOP_X" in os.environ` is an `ast.Compare`; the scan handled only calls and loaded subscripts. This is a *better* finding than round 2 — a membership test is the most natural way to spell a boolean flag, where `os.getenv(key=...)` is something nobody writes.

Rather than patch only what was reported, I swept every way a module can read the mapping. **Three were real gaps**, all now covered: `in`, `not in` (walking each pair of a chained comparison), and the `os.environb` bytes twin of every existing shape.

### Full matrix, verified in one pass by planting each into `verify.py`

**CAUGHT (12):** `get`/`pop`/`setdefault` literal · same-module alias · qualified `envvars.MUX_BACKEND` · imported registry constant · `os.getenv` · subscript · `key=` keyword · `in` · `not in` · `in` + registry constant · `environb.get` · `environb[...]`

**CORRECTLY SILENT (4):** `dict(os.environ)` · `{**os.environ}` · a non-`BMAD_LOOP_` var · a prose mention in a docstring

### Boundary, now stated in the docstring rather than left to be found

Deliberately **not** covered: rebinding the mapping (`e = os.environ; e.get(K)`) and `from os import environ`. Both obscure the *lookup* rather than the key. This guard is a tripwire for the change someone writes while trying to do the right thing — borrowing the registry's **key** is exactly that case, which is why every key spelling resolves; routing around the lookup is not, and pretending otherwise would make it a sandbox it cannot be.

---

### 🔎 A note on the review pattern, since it is the more useful signal

Three consecutive rounds each surfaced another uncovered AST shape. Every one was valid and I have fixed all three — but read together they are evidence about the **design**, not three unlucky misses: this detector is a *denylist of access forms*, so it is only ever as complete as the last sweep over them. Round 4 finding a fifth shape would not be surprising.

I have recorded that in the docstring, along with the instruction to extend the *matrix* rather than just the branch. The alternative design — flagging any `BMAD_LOOP_*` **name** in a non-allowlisted module, which is closed rather than open-ended — would also flag the ~6 modules that legitimately *build* env dicts for child sessions, so it needs a wider allowlist and a change to `src/` that is out of scope here.

Flagging that trade-off rather than silently grinding another round.

`uv run pytest -q -n auto` → **5008 passed** · `uv run pyright` → **0 errors** · `trunk check` → **no issues**. No `src/` change.

pbean:
@coderabbitai full review

pbean:
## CodeRabbit Major addressed (`aa7b036`) — the sharpest finding of the cycle

> This test only confirms that the current repository has no unallowlisted findings. If `_scan()` stops reporting `envread` findings, this assertion still passes.

**Correct, and it lands on exactly the thing I'd been getting wrong.** I had verified 12 access forms by hand-planting each into `verify.py` and reading the failures — then written the result into a **docstring** and thrown the evidence away. Prose does not fail a build. By this repo's own ablation rule, an ablation note is a *stamp*, not a *check*.

### What changed

`_scan_source(src, rel)` split out of `_scan` — behavior-preserving (same 52 `envread` findings, same counts for every other kind) — so snippets run through the **same code path** as the real scan. The matrix is now executable: **16** `ENV_READ_PROBES` that must produce a finding, **7** `ENV_READ_NON_PROBES` that must not.

The non-probes carry real weight. Without them a detector could pass every positive row by flagging *everything*, which is equally broken — a guard that cries wolf gets its allowlist widened until it means nothing. They pin the scoping the docstring argues for: bulk copies, a foreign var, prose, a **write**, and a `SessionSpec.env` lookup all stay silent.

### Measured — this is the finding's actual claim, confirmed

| Branch deleted | Probe rows red | Repo-wide test |
|---|---|---|
| membership (`ast.Compare`) | 3 | **PASSES** |
| `key=` keyword arm | 4 | **PASSES** |
| `REGISTRY_NAMES` resolution | 4 | **PASSES** |
| key resolver flags everything | 1 non-probe | fails |

Three detector regressions the pre-existing assertion **cannot see at all**. The repo-wide test is the invariant; the probes are the proof the invariant is being checked. Neither replaces the other, and the docstring now says so.

### Why this one mattered more than the three before it

Rounds 1–3 each found a *missing shape*. This found the reason those kept being findable: nothing in the suite could tell a covered shape from an uncovered one, so coverage drifted on prose alone. Adding a row to `ENV_READ_PROBES` now fails before the fix lands — which is the workflow the docstring already told future readers to follow, and previously couldn't support.

`uv run pytest -q -n auto` → **5031 passed** (+23) · `uv run pyright` → **0 errors** · `trunk check` → **no issues**. No `src/` change.

pbean:
@codex review

pbean:
## Both open findings addressed (`becbef8`) — one detector hole, one hygiene trim

---

### ✅ Codex P2 — chained membership resolves the wrong operand — **VALID, fixed**

> For a membership operator after the first comparison, the tested key is the preceding comparator, not `node.left`.

**Confirmed by running the shapes through the real `_scan_source`:**

| source | before |
|---|---|
| `X = "BMAD_LOOP_X" in os.environ` | FLAGGED |
| `X = c == "BMAD_LOOP_X" in os.environ` | **MISSED** |
| `X = c == MUX_BACKEND in os.environ` | **MISSED** |
| `X = c == "BMAD_LOOP_X" not in os.environ` | **MISSED** |

The AST is `Compare(left=c, ops=[Eq, In], comparators=[K, os.environ])`, and a chain expands **pairwise** — `c == K in os.environ` means `c == K and K in os.environ` — so the operand a membership tests is the comparator to its immediate left. The branch read `node.left` for every op, resolved `c`, and returned no key. Its own comment ("walks every (op, rhs) pair so a chained comparison cannot hide one") was false for any `in` past the first, which is the sharper half of the finding: the claim was wrong, not just the coverage.

Fix is to carry the left operand across the pairs (`left = rhs` at the loop tail). New probe row `membership-chained`, keyed by a **literal** rather than a registry constant on purpose — `membership-registry` already grades key resolution, so this row reddens for exactly one reason: chain position.

**Ablated** (`cp` backup, restored byte-identical, md5 verified):

| ablation | measured |
|---|---|
| revert the left-tracking to `node.left` | **1 failed** — `membership-chained` alone |
| delete the whole membership branch | **4 failed** — the 3 existing rows plus the new one; the repo-wide test still **PASSES** |

That second row also caught a now-stale number: the docstring said deleting the membership branch reddens *3* probe rows. It is 4. Corrected from the measurement rather than by arithmetic.

---

### ✅ CodeRabbit nitpick — dated probe records in test source — **VALID, fixed**

> Keep durable behavior facts in docstrings. Put dated planting runs, counts, and historical ablation output in git history.

Valid against this repo's own rule (`AGENTS.md`: "Never commit session notes, probe records, or run artifacts"). The calibration came from the house form: `test_engine.py` carries ~20 `Ablation target: delete X and this fails` notes and **not one date or pasted traceback**. So the ablation *target* is durable guidance and stays; the *record* goes.

**Dropped** — `Ablated 2026-08-11, six ways`, three `(verified 2026-08-11)` preambles, the pasted failure lines (`verify.py:37: _ABLATION_TIMEOUT = …`), the round-by-round narrative, and counts that go stale on contact (`52 real reads`, `12 read shapes`, `alone (of this file's 13)`).

**Kept, restated as contract** — which gate holds which rejection row; why the finiteness half alone holds `nan` (the trap: `nan <= 0` is *also* False, so "simplify it away, `> 0` covers nan" is a live regression); the prose row being a CONTROL rather than an ablation; the denylist-of-shapes property that makes this guard only as complete as its last sweep; and the claim that this assertion cannot grade the detector at all — now stated qualitatively, because `ENV_READ_PROBES` asserts it executably.

Also swept two dated notes the finding did not name — `test_tui_app.py`'s `x` and `A` ablation comments — since leaving them would have made the convention inconsistent inside one PR.

---

`uv run pytest -q -n auto` → **5032 passed**, 39 skipped, 5 xfailed (+1, the new probe row) · `uv run pyright` → **0 errors** · `trunk check` → **no issues**. No `src/` change in either commit.

pbean:
@codex review

pbean:
@coderabbitai full review

pbean:
## Round 2: three more findings, all valid (`c043cf0`, `8287792`)

---

### ✅ Codex P2 — file-wide exemptions defeat the guard's own distinction — **VALID, fixed**

> A future hook or Unity helper can read a core knob such as `BMAD_LOOP_MUX_BACKEND` directly and the generated `envread` finding will be discarded solely based on its path.

Correct, and it lands on the load-bearing part: the guard existed to separate *core* vars from *plugin-owned / session-protocol* families, and then threw that distinction away at the last step.

`ENV_READ_ALLOW` is now `rel -> allowed key families` rather than a set of paths, so the scan threads the resolved key into the `envread` finding. `envvars.py` is scoped to the names it defines — read off the live module, so a fourth registered var still needs no edit here; the hook relays to the session protocol; the Unity helpers to `BMAD_LOOP_UNITY_*` / `BMAD_LOOP_ENGINE_*` plus that protocol.

**Measured** — a core-knob read planted in `unity_ready.py` (an allowlisted file):

| allowlist semantics | result |
|---|---|
| family-scoped (this change) | **FAILS** — `unity_ready.py:54: BMAD_LOOP_MUX_BACKEND — _ABLATION_CORE = os.environ.get("BMAD_LOOP_MUX_BACKEND")` |
| file-scoped (as reported) | **PASSES** — silently accepted |

On today's tree the two semantics are indistinguishable, because every real read already sits inside its own family. That is precisely why the claim could not live in the repo-wide assertion, so it lands in `ENV_SCOPE_CASES` as rows over synthetic findings: revert the filter to a path set and exactly its three *should-be-refused* rows redden, nothing else.

**Writing the families down also refuted the allowlist's own comment.** It said the Unity helpers read the plugin's `BMAD_LOOP_UNITY_*` / `BMAD_LOOP_ENGINE_*` contract. They also read five session-protocol vars the engine injects — `RUN_DIR`, `WORKTREE`, `REPO_ROOT`, `CLEAN_TMP`, `QUIESCE_PHASE`. The rationale was accurate about the plugin half and silent about the rest; it now names both.

---

### ✅ CodeRabbit Minor — `os.environb` coverage claim was false — **VALID, fixed** (diagnosis adjusted)

> The docstring claims that `os.environb` covers the bytes twin of all access forms. The matrix probes only `.get()` and subscript access.

Valid, and the measurement moved where the hole was. The *mapping* forms were all already detected — `_is_os_environ` accepts `environb`, so pop/setdefault/keyword/membership/chained all fired unprobed. The two real misses were elsewhere:

| shape | before |
|---|---|
| `os.getenvb(b"BMAD_LOOP_X")` | **MISSED** — the arm matched the `getenv` string only |
| `K = b"BMAD_LOOP_X"` / `os.environb.get(K)` | **MISSED** — the alias table bound `str` constants alone |

`os.getenvb` is real and POSIX-only in the same way `environb` is (`os.supports_bytes_environ` is True here), so it is the bytes twin the claim promised. Both closed, and the axis is swept rather than sampled — 9 environb/getenvb probe rows now, not 2.

Ablations: dropping `getenvb` reddens its 2 rows; dropping the bytes-alias decode reddens `environb-local-const` alone.

The key-spelling half of the claim is stated precisely instead of broadened: bytes twins take a bytes literal or a bytes constant, and the registry's constants are `str`, so that half of the cross product **cannot be written** rather than being uncovered.

---

### ✅ CodeRabbit Minor — review-history narrative in docstrings — **VALID, fixed**

"Four review rounds…" and "Written after a review pointed out…" are records. Both gone; the durable claims they were evidence for stay — that a denylist of access forms is only as complete as its last sweep, that the matrix gets extended before the branch, and that prose does not fail a build.

---

`uv run pytest -q -n auto` → **5048 passed**, 39 skipped, 5 xfailed · `uv run pyright` → **0 errors** · `trunk check` → **no issues**. Still no `src/` change beyond the `envvars` fix already in this PR.

pbean:
@codex review

pbean:
@coderabbitai full review

pbean:
## Codex P2 addressed (`3464e8e`) — valid, and it is a defect I introduced last round

> When a new key extends an existing singleton name — such as `BMAD_LOOP_MUX_BACKEND_FALLBACK` — this predicate exempts it because it starts with `BMAD_LOOP_MUX_BACKEND`.

**Correct.** Scoping the allowlist by family in the previous commit, I matched *every* entry with `startswith`. That is right for `BMAD_LOOP_UNITY_` and wrong for `BMAD_LOOP_MUX_BACKEND`: an entry naming one variable exempted every longer name built on it, so an unregistered knob could ride in on a registered one's spelling — the guard waving through precisely what it exists to make someone register. Same hole on each session-protocol name.

A trailing underscore now marks a family (prefix match); every other entry is one variable (exact match).

**Ablated:** reverting to the bare prefix test reddens `registry-name-extended` and `session-name-extended` and nothing else. A real family still covers a member it has never seen (`unity-family-unseen-member`), which is what keeps it a family rather than a list.

`uv run pytest -q -n auto` → **5051 passed** · `uv run pyright` → **0 errors** · `trunk check` → **no issues**.

---

## Stepping back: this is round 7, and I am stopping here

Each finding this cycle was real and each fix was measured. But the rounds are not random — they are the same shape every time, and this last one was a defect *introduced by the previous fix*. Per AGENTS.md ("Review non-convergence is evidence about the approach, not just a defect queue — escalate rather than grind"), that is worth naming rather than grinding through an eighth.

The guard matches on **how a variable is read** — an AST denylist of access forms — and the set of Python expressions that can use a mapping is open-ended, so its completeness is unmeasurable by construction. Rounds 1, 2, 3, 5 and 6 were each the next corner of that cross-product.

Measured against `src/bmad_loop`: **106** `BMAD_LOOP_*` literals across 21 files (52 consuming side, **54 producing side**), **74** mentions of an env name, **22** of them legitimate foreign-var reads.

Filed as **#542**: replace it with a guard that matches something *closed* — a `BMAD_LOOP_*` literal may appear only in the registry, the owning plugin, or the stdlib-only relays. You cannot name a variable without a literal or an imported constant, so access shape stops mattering entirely and five of the seven rounds become unreachable rather than covered. That is also what the invariant literally says: core vars *register* in `envvars.py` — a claim about where a variable is **named**, which this guard reinterpreted as how it is **read**. Runtime observation and semgrep taint mode are written up there as considered-and-rejected, with reasons.

Deliberately **not** in this PR, which is a test backfill. The guard here is green, its boundary is documented, and its matrix is executable — a new shape fails before it is fixed.

pbean:
@codex review

Issue #498, opened 2026-08-08
The resume config-change warning baselines off state.json, inside the tree it polices


pbean:
Found while re-validating PR #495 (#461 point 4). Minor, but it makes a shipped line of code read as a security control when it is not one.

## The asymmetry

PR #495 added two consumers of the launch-time `config_digest`. Only one of them has a baseline an agent cannot reach.

**The auto-sweep gate is sound.** `_sweep_factory`'s `trusted_digest` is threaded in memory from launch — `cli.py` `cmd_run` computes it once and passes it in, and `_resume_paused_run` passes the freshly computed `new_digest`. It is never read back from disk, so a driven session has nothing to rewrite. That is the refusal path, and it holds.

**The resume warning is advisory.** It compares against `state.trusted_config_digest`:

```python
security_config_changed = bool(state.trusted_config_digest) and (
    new_digest != state.trusted_config_digest
)
```

`RunState` round-trips through `<project>/.bmad-loop/runs/<id>/state.json` (`runs.RUNS_DIR = .bmad-loop/runs`, `run_dir_for`) — inside the same agent-writable tree `config_digest`'s own docstring says driven sessions can write anywhere in. A session that rewrites `policy.toml` can also rewrite that field and the operator is never told. Blanking it is enough: the `bool(...)` guard reads `""` as "no prior pin, no warning".

So the warning catches an incidental or accidental config change. It does not catch a deliberate one, which is the case it is worded for ("re-read `.bmad-loop/policy.toml` … first if you did not make that edit").

## Scope

This is not specific to one field. Every `RunState` field has the property — `policy_snapshot`, the budget counters, `paused_reason`. `trusted_config_digest` is just the first one that reads like a security control, which is why it is worth either fixing or saying out loud.

Fixed for now by saying it out loud: the comment at the comparison site now states that it is advisory and why, so the next reader does not mistake it for the gate.

## Options

1. **Leave it advisory, documented.** What is shipped. Defensible: the human-present path only warns by design, and an adversary who can rewrite `state.json` has other levers anyway.
2. **Hold the resume baseline outside the project tree** — e.g. alongside other user-scoped state rather than in the run dir. Closes it for this field only, and puts one `RunState` field somewhere different from the rest, which is a coherence cost.
3. **Treat `state.json` integrity as its own concern** — the honest general fix, and much larger than #461 point 4. Would subsume this.

No action needed for #495 to land; the gate it adds is unaffected.

Related: #461, PR #495.

[...]

Issue #461, opened 2026-08-05
`init` vendors the hook relay into the agent-writable workspace, turning a file-write primitive into unattended persistent code execution


LuckierTrout:
## Summary

`bmad-loop init` writes `bmad_loop_hook.py` into `<project>/.bmad-loop/` and registers that in-workspace path in each CLI's hook config. Because the driven sessions have write access to that same tree, **a session with only a file-write primitive — no exec — can gain unattended code execution that persists across all future sessions in the project**, including ordinary interactive sessions running under normal permission modes.

The relay's own code is clean. This is about *where* `init` places it and what it registers.

A second, smaller issue in the relay's event write is included at the bottom.

## Environment

- `bmad-loop` 0.9.0, installed via `uv tool install "bmad-loop[tui] @ git+…"` at commit `f512288f4e7ca4f028da324b280762fd413b9f90`
- Adapters registered: `claude`, `codex` (`codex-cli` 0.146.0)
- macOS (arm64), Python 3.11
- `bmad-loop init --project <root> --cli claude --cli codex`

---

### What `init` produces

```
.claude/settings.json   hooks → python3 <project>/.bmad-loop/bmad_loop_hook.py <Event>
.codex/hooks.json       hooks → python3 <project>/.bmad-loop/bmad_loop_hook.py <Event>
```

Every element of that chain is writable by the very sessions the orchestrator drives:

1. **The relay script** lives at `<project>/.bmad-loop/bmad_loop_hook.py` — inside the worktree.
2. **The interpreter is `python3`, resolved via `PATH`.** On this machine ~8 user-writable directories precede `/usr/bin`, so the interpreter is hijackable without touching any repo file.
3. **The hook config files themselves** (`.claude/settings.json`, `.codex/hooks.json`) are in-workspace and writable.
4. **Same class, adjacent blast radius:** `.bmad-loop/policy.toml` `[verify] commands` (run by the orchestrator on the host after a clean review) and `.bmad-loop/profiles/*.toml`, which overlay the packaged profile *including `binary`* — the command the orchestrator launches.

### Why it matters

Hooks execute **outside the CLI permission system**. A prompt-injected dev/review session, a malicious dependency, or any code path with a write primitive can append a payload to the relay. Nothing runs at that moment. On the next `SessionStart` of *any* session in the project — including the operator's ordinary interactive `claude`, in a mode where every Bash call would be prompted — the payload executes with the user's full credentials and **no approval prompt**.

[...]

Issue #139, opened 2026-07-14
Worktree teardown race crashes the run after a successful merge


tjoignant:
**Version:** bmad-loop 0.8.1 (uv tool) · macOS 15 (Darwin 24.6.0) · git worktree isolation, `adapter.name = "claude"`

## Summary

When a story's session leaves a process still writing into its worktree (for us: pytest), `close_unit_workspace` fails teardown and crashes the whole run — *after* `unit-merged` has already landed the work on the target branch. No code is lost, but the run dies and leaves an orphaned worktree directory with a dangling `.git` gitfile.

Hit 2 out of 2 backend stories in a single sweep run, with an identical fingerprint.

## What I expected

A merged story tears down its worktree. If the teardown can't complete, that's housekeeping — it should be logged and reconciled, not turn a successful merge into a run crash.

## Observed

`crash.txt`:

```
File ".../bmad_loop/workspace.py", line 157, in close_unit_workspace
    verify.worktree_remove(unit.repo_root, unit.path, force=not success)
GitError: git worktree remove .../worktrees/dw-db-schema-and-gate-falsifiability failed:
error: failed to delete '.../worktrees/dw-db-schema-and-gate-falsifiability': Directory not empty

During handling of the above exception, another exception occurred:
  File ".../bmad_loop/workspace.py", line 159, in close_unit_workspace
GitError: git worktree remove ... failed:
fatal: '.../worktrees/dw-db-schema-and-gate-falsifiability' is not a working tree
```

Timeline (journal `ts` + file mtimes on the leftover worktree):

```
16:13:02  session-end            review-2
16:13:04  unit-merged            <- work is safely on the target branch
16:13:04  worktree root mtime    <- git worktree remove starts
16:13:07  backend/.pytest_cache/{CACHEDIR.TAG,README.md,.gitignore} written
16:13:09  rmdir(worktree) -> ENOTEMPTY -> run-crash
```

Those are file *content* writes (own mtimes), not git deletions — a pytest process from the just-ended session was still alive and recreated the cache dir after git had scanned past it. The second story is the same to the second (cache written `11:42:20`, crash `11:42:21`). In both leftover worktrees the only surviving entry under `backend/` is `.pytest_cache`.

## Analysis (my reading, less certain than the above)

Two separate things combine:

[...]

PR #568, opened 2026-08-13
feat(events): relocate the hook event channel to a user-scoped state root (#494)

### LINKED ISSUE #494: Windows: narrowed but non-zero check-to-write race in the hook relay's event write
## What

On Windows the hook relay's event write still has an open (but narrowed)
check-to-write window. `src/bmad_loop/data/bmad_loop_hook.py::_write_event`
refuses a redirected `events/` dir (symlink or junction), then — on the fallback
path only — re-resolves the directory **by path** for the `os.open` and the
`os.replace`.

PR #493 narrowed this: the redirect check runs again after the payload is
written and before it is published, so a junction that is swapped in and **left
in place** is refused and the temp file removed. What remains is a
swap-and-restore inside the window: replace `events/` with a junction, let the
relay create its temp file inside the attacker's directory, then restore the
real directory before the post-write check runs.

## Why it is not already fixed

POSIX closes this properly — the create+rename are anchored to a `dir_fd` opened
`O_RDONLY|O_DIRECTORY|O_NOFOLLOW`, so no later path resolution happens.

Windows has no equivalent reachable from the standard library:
`os.supports_dir_fd` is empty there (dir_fd is implemented with the POSIX `*at`
calls), and Python exposes no handle-relative open. Closing it properly needs a
directory handle from `CreateFileW` with
`FILE_FLAG_BACKUP_SEMANTICS | FILE_FLAG_OPEN_REPARSE_POINT`, and then
`NtCreateFile` with `OBJECT_ATTRIBUTES.RootDirectory` set to it — i.e. `ctypes`
against `ntdll`, inside a relay that is **stdlib-only by contract** and runs
under whatever interpreter the host happens to have. That is disproportionate to
the residual risk and carries its own hazard: a bug in it stops hook delivery on
Windows, which stalls every run — the exact harm being defended against.

## Residual risk

Low. The attacker must win a precisely-timed race against a hook that fires at
moments they do not control, and must restore the real directory immediately
after. Failing the race in either direction is safe: a persistent junction is
refused (before or after the write), and a refusal degrades to the same
`session_timeout_min` path the run would have hit anyway. No silent redirect
survives unless the restore is perfectly timed.

## Options if this is ever picked up

1. Accept and document (status quo). The narrowed window is recorded in the
   `_write_event` docstring.
2. `ctypes` + `NtCreateFile` handle-relative open on Windows. Correct, but a
   large Windows-CI-only-testable surface in a stdlib-only relay.
3. Move the relay's event writes out of the agent-writable tree entirely, which
   removes the attacker's ability to touch `events/` at all. This overlaps with
   the direction of #461 Phase 2 and is probably the cheapest real close.

Option 3 is the one worth evaluating first.

Follow-up from #461 Phase 1 (PR #493), raised by an automated reviewer on that PR.

pbean:
A run's hook-event channel — the only path by which a session reports that it stopped —
lived at `<project>/.bmad-loop/runs/<run-id>/events/`, inside the tree the driven session
is free to write. This moves it to a user-scoped state root, and adds the root itself as a
reusable place for run control-plane state that must outlive a branch switch.

## The recorded decisions

Both issues were triaged to the same shape on 2026-08-08, and the maintainer decision on
each says explicitly that they land as one relocation.

**#494** ("Windows: narrowed but non-zero check-to-write race in the hook relay's event write"):

> Maintainer decision 2026-08-08: **option 3 — move the relay's event writes out of the
> agent-writable tree**, landing in the PR #238 rework (extract #493's hardened primitive;
> the relocation and the extraction are one change). With events/ somewhere sessions cannot
> write, the check-to-write race class disappears and `_is_link_like` becomes
> belt-and-braces. The ctypes/NtCreateFile route (option 2) is declined: a large
> Windows-only ntdll surface inside the stdlib-only relay risks stalling every Windows run
> — the exact harm being defended. Shares the new out-of-tree control-plane root with
> #498's baseline relocation.

**#498** ("The resume config-change warning baselines off state.json, inside the tree it polices"):

> Maintainer decision 2026-08-08: **relocate the resume baseline out of the agent-writable
> tree** (issue option 2). `trusted_config_digest` moves to a user-scoped state root keyed
> by (project, run_id); all three stamp sites (cli.py:1249/1269, :1452, :1696) move
> together, the resume re-stamp is preserved, and the sweep gate's in-memory baseline is
> untouched. Executed as one shared relocation with #494 option 3 (the events/ move in the
> PR #238 rework) — one new out-of-tree control-plane root, two artifacts move into it.
> Full state.json sealing (option 3 of the issue) was considered and declined as
> disproportionate to an advisory warning's baseline.

PR #238 was closed unmerged; this is its rework, with the extraction (phase 1) and the
relocation (phase 3) as the one change #494 called for.

## Scope

**Moves:** the events channel, to `<state root>/<project tag>/<run id>/events/`.

**Does not move:** `RUNS_DIR` (`.bmad-loop/runs/`) and everything else under a run dir —
`state.json`, logs, bundles, feedback, worktree records — stay exactly where they are, and
so does the `tasks/` mailbox (`<run-dir>/tasks/<task-id>/`). The relocation is deliberately
two artifacts, not the run directory: an operator inspects a run dir by hand, `archive`
tars it, and moving the lot would be a much larger compatibility change than either issue
asked for.

**Follows separately:** the `trusted_config_digest` half of #498. This PR builds the shared
root and its accessors (`state_root`, `project_state_root`, `state_dir_for`) and moves the
first artifact into it; the digest's three stamp sites move onto the same root in a
follow-up PR, so #498 stays open after this merges. Only the events move is claimed here.

Closes #494

## What lands

- **`events.py`** — an importable twin of the hardened write in `data/bmad_loop_hook.py`.
  The hook ships as package *data* and is stdlib-only by contract, so neither side can
  import the other; the copy is pinned by an AST parity test over the five twinned names,
  and the half `ast.get_source_segment` cannot reach (the payload shaping the hook does
  inline in `main()`) is pinned behaviorally against the real hook subprocess.
- **`bmad-loop relay <Event>`** — a console-script hook target instead of a workspace file
  path a branch switch can take away. It honours the hook contract exactly (silent no-op
  outside a driven session, garbage stdin tolerated, `OSError` degraded, nothing ever on
  stdout, rc 0 always) and dispatches ahead of `main()`'s shared `try/except` and its
  `_configure_mux` call, whose arms print `error: …` and return 1/130 — the CLI-window
  failure the contract forbids.

  **Nothing points at it yet, deliberately.** `init` still registers the copied workspace
  relay, so this command changes no run today; retargeting those registrations is #461
  Phase 2, and that move carries its own obligation to *retarget* rather than drop the
  `hooks.relay-present` check, which the COUPLING note at `cli.py:499` reserves for it.
  Doing it here would drag that check's rework into an events-relocation PR.
- **The state root** — `BMAD_LOOP_STATE_DIR` (honoured as spelled, and required to be
  absolute: the orchestrator and the session it launches read the root from different
  working directories, so a relative value names two places and is refused rather than
  resolved against one of them), else
  `$XDG_STATE_HOME/bmad-loop` or `~/.local/state/bmad-loop`, and on win32
  `%LOCALAPPDATA%\bmad-loop\state` or `%USERPROFILE%\AppData\Local\...`. Never
  `Path.home()` on Windows: it prefers `HOMEDRIVE`+`HOMEPATH` once `USERPROFILE` is absent,
  which on a domain-joined machine can put the control plane on a network share. When no
  base answers, this raises `StateRootError` rather than degrading — it is a write path,
  and every degraded outcome here is silent. Keyed with the existing `project_tag`, so a
  symlinked or relative spelling of one project cannot open a second control plane.
- **Observation surfaces** — `validate` gains `hooks.relay-stale` (a warning that never
  moves the exit code); `diagnose`'s `events` file group now stats the state-root location
  and sums it with the still-live legacy root into the one category, payload and
  `SCHEMA_VERSION` unchanged.
- **Lifecycle** — `delete`/`archive` remove the out-of-tree counterpart on a never-raise
  tail (#139 doctrine: the run dir is already gone by then, so failing the operator's
  delete would report a removal that happened). `clean` sweeps this project's orphans —
  keyed on the run directory *existing*, not on its `state.json` parsing, so a corrupt run
  an operator is recovering keeps its control plane. It enumerates state entries **before**
  reading live run-dir names, which is a race guard rather than a style choice: a run
  creates its run dir strictly before its state dir, so reading in that order means anything
  in `entries` is certain to appear in the later `live` read — and a `clean` overlapping a
  run's startup cannot delete the control plane of a run that is starting. `clean --json` gains
  `state_dirs_swept` as an additive field on the existing schema version. `trim_run_dir` is
  deliberately untouched: a trimmed run is still resumable. All three guards hold
  `RuntimeError` alongside `OSError`, because `project_tag` resolves before digesting and
  below 3.13 `Path.resolve` reports a symlink loop that way — measured, not assumed: 3.11
  and 3.12 raise, 3.13 and 3.14 return the unresolved path.

## The version skew is the whole risk

`bmad_loop_hook.py` is **copied** into a target project by `init`. So an upgraded
orchestrator routinely drives sessions whose relay knows only the legacy location. Old
relay + new orchestrator with no fallback = no `Stop` is ever observed = every session
stalls to `session_timeout_min` — which fails no test suite and hangs an overnight run.
Three things keep that from happening, and the skew is guarded in **both** directions:

1. **Producer.** `Engine._run_session` exports `BMAD_LOOP_EVENTS_DIR`. That env dict is the
   one required site — dev/review, sweep bundles, stories and injected plugin-workflow
   sessions all dispatch through it. The non-sites fail closed without it (`resolve.py`
   sets no `BMAD_LOOP_TASK_ID`, so the relay no-ops; `probe.py` captures through its own
   probe relay; `plugins/bus.py` spawns plain shells with no task id).
2. **Relays.** Both the copied hook script and `bmad-loop relay` prefer that variable and
   fall back to `<run_dir>/events`. `or`, not a presence test — an exported-but-empty value
   names the launch cwd. The variable is deliberately **not** part of the no-op detector,
   because the mirror-image skew (older orchestrator, newer relay) must still write its
   events.
3. **Consumer.** `SignalWatcher` takes an optional `legacy_dir` and polls primary then
   legacy, ordering by the parsed `ts` across both; `_consumed` keys on `(dir, name)` so a
   name consumed in one dir cannot mask a different event in the other. Only the primary is
   created — re-creating the in-tree dir would undo the move for the operator's
   `git status` and gain nothing.

And `validate`'s new `hooks.relay-stale` names the skew out loud: `hooks.relay-present`
stats the relay but never asks which version it is, so an install predating the move read
green. The installed copy is now compared against the packaged source **as text** — which
is exactly the `read_text`/`write_text` round trip `install_into` performs; a byte compare
would call every Windows install permanently stale — and a difference is reported with the
`bmad-loop init` that fixes it. A warning, never a problem: the dual-poll fallback keeps a
stale relay working, so the exit code must not move.

## Honest caveat on what "out of tree" buys

For the #494 race class, the move is a **class elimination against in-tree write
primitives**: nothing a session does to `<run-dir>/events` — symlink, junction,
swap-and-restore — touches the channel the orchestrator actually polls. Against a
**sandboxed** session it is a hard boundary, because the state root is outside the paths
such a session is permitted to write.

It is **not** a boundary against a `--dangerously-skip-permissions` session. The env var
necessarily tells the session where the directory is; a session with unrestricted
filesystem access can read `BMAD_LOOP_EVENTS_DIR` and go there. That is exactly why the
relay keeps `_is_link_like` — the hardening from #493 stays in place as belt-and-braces
rather than being retired on the strength of the move. Claiming otherwise would be the
failure mode the issue itself warns about: a line of code reading as a security control
when it is not one.

Known limit, documented rather than fixed: a moved or renamed project re-keys under a new
`project_tag`, stranding its old subtree. `clean` sweeps orphans for the project it is run
in, so the stranded subtree is not collected by the new location.

## Verification

`uv run pytest -q` (5311 passed, 44 skipped, 5 xfailed), `uv run pyright` (0 errors),
`trunk fmt`, `trunk check --all`. Every new guard in phases 3–5 was ablated **singly**,
requiring the ablated run to exit rc 1 — hook side and watcher side separately; the
`validate` arms against a forced-True comparison, a deleted block and a hoisted gate; the
`diagnose` arms against each root dropped singly and an escaping raise; 13 rows for the
lifecycle phase. One test was found vacuous by that process and rewritten: the
unregistered-hooks gate degraded either way with no relay on disk, and now leaves the relay
installed and stale.

E2E skew coverage: the tmux and stories fakes write to `$BMAD_LOOP_EVENTS_DIR` and each
keeps a legacy-location twin, so the dual-poll path is exercised end to end rather than
only in unit rows.


pbean:
Triage of the CodeRabbit round on `6570e1e` — 5 of 6 taken, pushed as `debce19`.

**Taken**

- **`runs.py` `RuntimeError` from `Path.resolve()` (Major)** — confirmed, and reachable on two of the four supported interpreters. Measured rather than assumed: 3.11 and 3.12 raise `RuntimeError` on a symlink loop, 3.13 and 3.14 return the unresolved path. `project_tag` resolves before digesting, so all three guards leaked it — `_discard_state_dir`, and both the root and the per-entry containment guard in `reconcile_orphan_state_dirs`. Each has its own consequence and each was ablated singly requiring rc 1: the delete tail failed the operator over a removal that had already happened, the root guard took the whole `clean` down after its real work was finished, and the containment guard aborted a sweep half-done and reported none of what it had just deleted. Regression coverage added for all three, including the entry-resolution arm requested.
- **`docs/setup-guide.md` state root** — correct, and the failure is exactly backwards from harmless: with `BMAD_LOOP_STATE_DIR` set, the operator deleted a directory that was not their state root and kept the one that was. The command now resolves the same cascade the orchestrator does.
- **`SYSTEMROOT` in the hook subprocess** — taken as latent-flake removal rather than a live failure (both Windows CI jobs are green on the current env). Passed through on `nt` only; the env stays minimal otherwise, since the point of that row is that a stdlib-only relay needs nothing else.
- **`events` shadowing the module** — real, renamed.
- **Nitpick, `install.py` duplicated packaged-source spelling** — the docstring was the wrong half. It claimed the spelling "has exactly one home" while the literal also appears in `install_into`, so it has been corrected to say what is true (reader and writer stay in one module). Not extracting an accessor: `install_into` is untouched by this branch, and this repo does not refactor untouched code to satisfy a rule.

**Declined — `diagnostics.py` `_events_dir` `except Exception`**

Grounds, in order of weight:

1. This is observation, and the repo doctrine is explicit that observation degrades where repair writes raise (`AGENTS.md`). `_events_dir` feeds a file count in a dump.
2. Narrowing inverts the failure this very PR fixes. `diagnose --all` dumps are routinely read on a machine other than the one that produced them, and the derivation consults that host for a state root. Under an enumerated type set, any type not enumerated stops being an events count of 0 and becomes a crashed dump — which is strictly worse than the pre-PR behavior, where the category silently vanished.
3. The `BLE001` citation comes from a broader ruff configuration than this repo enforces; `AGENTS.md` says not to ratchet ruff past CI, and the surrounding comment already states the breadth is deliberate rather than silenced.

The cost I am accepting: an unexpected defect inside `events_dir_for` is swallowed here. It is not swallowed everywhere — `state_root`, `state_dir_for` and `project_tag` are pinned directly in `tests/test_runs.py`, so a defect there reddens at its own layer instead of only through `diagnose`.

pbean:
@codex review

pbean:
Codex round 1 on `debce19` — both findings confirmed and fixed in `5175f53`. Both were live-run kill paths, and neither was caught by the phase ablations because both are about *interaction*, not about a guard being absent.

**P1, orphan sweep races a starting run — confirmed.** I verified the premise in the code rather than taking the report: `compose_run` builds the `Journal`, whose `__init__` mkdirs the run dir (`journal.py:24`), and only then calls `make_adapters`, whose `SignalWatcher.__init__` mkdirs the events dir (`signals.py:65`). So the run dir is created strictly before the state dir, and the sweep was reading them in the order that turns that into a hazard.

Fixed by the read order rather than by a second check, because the ordering makes the race structurally impossible once the reads are the right way round: anything present in `entries` had its state dir on disk at the first read, so its run dir was on disk before that, so the later `live` read is certain to contain it. A re-check before removal would also work but would only narrow the window rather than close it. A run dir that disappears *between* the reads is the opposite case and is still swept — by then it is a genuine orphan.

**P2, relative `BMAD_LOOP_STATE_DIR` — confirmed.** The override was the one candidate `_state_base` did not filter, and `_state_base`'s own docstring already spells out the consequence for relative *derived* bases: "the state root would then move with the launch cwd, and a run whose control plane it cannot find again is a run that stalls to `session_timeout_min`". The override arm exempted exactly the value most likely to be typed by hand, and the two readers are two processes with different cwds — the engine exports the root to the session as `BMAD_LOOP_EVENTS_DIR`, the multiplexer launches that session at `spec.cwd` (a worktree under isolation), the watcher polls from the orchestrator's cwd.

It **raises** rather than absolutizing, which is the part worth stating explicitly since the docstring elsewhere says a stated override must not be silently countermanded. Raising is not that countermand — it names the variable and the fix. Absolutizing against whichever cwd this process happens to have is the guess, and it would pick one of the two candidate directories at random, making the stall harder to see rather than gone. `_state_base`'s not-the-root half is deliberately still not applied to the override: that half exists to stop a broken environment's `""` from landing a *guess* at `/`, and an override is not a guess.

Both ablated singly, requiring rc 1 — reverting the read order sweeps the mid-startup run, and dropping the `os.path.isabs` arm fails all three relative rows. The race test starts a run *inside* the `live` read so that whichever read runs second is the one that sees it, which is what a real interleaving does rather than a patch of the sweep.

Docs follow the behavior: `envvars.state_dir`'s "relative spelling included" was made false by this change and now states the reader/judge split, and the README env-var table and CHANGELOG carry the absolute requirement.

Local gates at `5175f53`: pytest 5311 passed / 44 skipped / 5 xfailed, pyright 0 errors, `trunk check --all` clean.

pbean:
@codex review

pbean:
Codex round 2 on `5175f53` — one P2, **declined on its premise**, with the durable half taken elsewhere. Pushed `f1f7554`.

**The finding:** the unconditional `events_dir=` keyword in `make_adapters` breaks a co-installed adapter class with a fixed constructor accepting "the previously documented bootstrap arguments", and the fix is to preserve the old call shape or introduce a negotiated/versioned capability.

**Why the premise does not hold.** Those arguments have never been documented to anyone outside this tree, because the seam has not shipped:

- Last tag is `v0.9.1` (2026-08-02). The adapter-registry entry (#226, PR #239) is under `## [Unreleased]` — line 42, where `## [0.9.1]` starts at line 1101. It has never appeared in a tagged release.
- `docs/adapter-authoring-guide.md` says so in as many words: *"There is no out-of-tree adapter-**class** package to copy yet."*
- The registry and `events_dir` therefore ship in the **same** release. The first author who can write against this contract sees `events_dir` in it from day one. There is no old call shape to preserve, because there is no released call shape.

A negotiated/versioned capability protocol for a seam with zero external consumers, added before its first release, is the trade this repo explicitly avoids — and versioning a contract nobody has yet consumed prices in a cost with no counterparty.

**What I did take.** Two real things sit underneath the finding, and each belongs at a different layer.

1. **The contract was implicit** — the guide told an author how to register a kind and what `dev`/`plain` means, but not that the bootstrap's keyword set grows. `f1f7554` states it: accept `**kwargs` in both variants, because `cls(**build_kwargs)` passes a description of the run that core adds to over time. It also names the bundled families as explicitly **not** the pattern to copy on this point — they carry closed signatures because they are edited in the same commit that adds the keyword, which an out-of-tree class cannot be. What they do model is the other half, accepting what you have no use for: `opencode_http` takes `events_dir` and immediately `del`s it, since that family observes over SSE and fires no hooks. Getting this stated before first release is the cheap half and the one with a deadline.

2. **The core-side failure mode is genuinely wrong, and outlives this PR** — filed as #569. A signature mismatch raises `TypeError`, which no family declares in `construct_error`, so it escapes as a bare traceback *after* `compose_run` has written the run state and pid — stranding a run directory. That is the exact hazard the `kind.load()` arm two lines above is written to prevent, and that arm was given a clean `error:` line naming the profile and the kind. The construct call has no equivalent. It is owned by #226/#239, not by #494, and #569 also flags the trap in fixing it: a `TypeError` raised from *inside* a working `__init__` is a bug in that package and must still surface as itself, so the fix has to inspect more than the exception type.

On the note that the registry tests miss this because the stub accepts `**kwargs` — the stub is right, and deliberately so: its docstring says the dev double "mirrors the real `(*args, paths, **kwargs)` dev contract". The stub matches what the guide now states explicitly. A test pinning a *closed* out-of-tree signature would pin the anti-pattern.

Local gates at `f1f7554`: pytest 5311 passed / 44 skipped / 5 xfailed, pyright 0 errors, `trunk check --all` clean.

pbean:
@codex review

pbean:
Codex round 3 on `f1f7554` — one P2, **confirmed and fixed** in `051752c`. Different area from round 2, so this is not a sibling of that disagreement.

**The finding is right about the fact.** `install._hook_command` still emits `<interpreter> <project>/.bmad-loop/bmad_loop_hook.py <Event>`, so no installed hook reaches `cmd_relay` today, while the subcommand's argparse help said `"(invoked by the installed hooks, not by hand)"`. That string was false.

**Taking the second of the two remedies codex offered — "stop presenting this as the installed hook target" — not the first.** Retargeting the registrations is #461 Phase 2, and the tree already says so in the places that matter: the COUPLING note at `cli.py:499` reads "Phase 2 **moves** the relay to the installed console script … and retires `HOOK_SCRIPT_REL`", and `events.py` calls it "the #461 Phase 2 hook target". The help string was the single place stating it as already done.

That note also names the obligation that makes the switch a separate change rather than a one-line edit: it requires `hooks.relay-present` to be **retargeted** to stat what the registration actually points at (the resolved `bmad-loop` executable), *not* dropped — because the stall it guards against survives the move, an entry point that is gone or unreadable stranding every hook event exactly like a missing script. Doing the switch in this PR would drag that check's rework into an events relocation.

**What changed:**

- The help now reads "(a hook target for machines, not a command to run by hand)" — keeps the do-not-run-by-hand guidance, drops the false claim.
- `cmd_relay`'s docstring says outright that nothing points at it yet, names what `_hook_command` actually emits, and points at the Phase 2 obligation. A console script that exists and is documented reads as the live path, and an operator debugging a lost Stop needs to know which relay actually ran.
- The CHANGELOG entry says the same, so the feature is not read as changing a run it does not touch.
- PR body updated to match.

Two observations in the finding that I am accepting as accurate rather than fixing: `cmd_relay`'s hardened stdin/error handling is indeed not exercised by installed hooks (it is exercised by tests, and it exists ahead of the switch on purpose — phase 1 built the twin and the command, Phase 2 points hooks at it), and a missing or unusable copied relay does still lose the Stop. That last one is unchanged by this PR and is exactly what `hooks.relay-present` reports.

Local gates at `051752c`: pytest 5311 passed / 44 skipped / 5 xfailed, pyright 0 errors, `trunk check --all` clean. I also grepped the tree for any other instance of the stale claim — none.

pbean:
@codex review

pbean:
CodeRabbit round 2 on `f1f7554` — one finding, **confirmed and fixed** in `95e622d`. It caught a defect I introduced in the previous commit, and the mechanism is worth recording.

The bullet I added nested a code span containing asterisks inside a bold span — `` **… `**kwargs` in both variants.** `` — which prettier cannot disambiguate. `trunk fmt` rewrote it in place, and because `trunk fmt`/`trunk check --all` both reported clean afterwards, the mangled text committed silently. What actually landed was wrong in four ways at once:

- `runsetup.make_adapters` became `runsetup.make*adapters` — a wrong identifier, not just wrong styling
- `cls(**build_kwargs)` became `` `cls(\*\*build_kwargs)` `` with the backslashes rendering literally inside the code span
- the bold and emphasis runs broke (`\_description of the run*`)
- the spaces around three code spans were eaten, joining them into their neighbours

Fixed by restructuring so no asterisk-bearing code span sits inside a bold span: the lead-in is bold and plain, and `**kwargs` moves into the sentence after it. That is how the neighbouring `(*args, paths, **kwargs)` line on the previous bullet has always been written, which is exactly why that one has never mangled — a useful tell for the next person editing this file.

I re-ran `trunk fmt` and then re-read the rendered lines rather than trusting the formatter's exit code, since a clean exit is what hid the problem the first time, and grepped the file for other escape/emphasis artifacts — none.

Local gates at `95e622d`: `trunk check --all` clean; pytest and pyright unchanged from `051752c` (docs-only commit).

Status of the codex gate: round 4 on `051752c` returned **zero findings** ("Codex Review: Didn't find any major issues"), arriving as an issue comment. This push is docs-only on top of that clean head, so I expect codex to re-review `95e622d` automatically and will report that round before treating the gate as closed.

PR #570, opened 2026-08-13
fix(resume): move the config-change baseline out of the agent-writable tree (#498)

### LINKED ISSUE #498: The resume config-change warning baselines off state.json, inside the tree it polices

pbean:
Closes #498.

The second half of the recorded 2026-08-08 decision. PR #568 built the user-scoped state root and moved the events channel onto it (#494); this moves the other artifact the decision names — the resume config-change baseline — into the same root. No new infrastructure.

## The defect

`resume` warns the operator when the host-exec config pinned at launch has changed under a run — verify commands, the resolved launch binary/args/env, the plugin allowlist (#461 point 4). The baseline it compares against round-tripped through `<project>/.bmad-loop/runs/<id>/state.json`, **inside the very tree the digest exists to police**. A session that rewrote `policy.toml` could blank `trusted_config_digest` in the same breath and the warning never fired — blanking sufficed, because the guard reads `""` as "no prior pin". The advisory caught the incidental change and never the deliberate one, which is what #498 was filed on.

## What moves

The digest is stamped at `<state root>/<project key>/<run-id>/config-digest`, beside the run's `events/`. Two new accessors in `runs.py` (`read_trusted_config_digest` / `write_trusted_config_digest`) plus a path helper; all three stamp sites move together:

| site | before | after |
| --- | --- | --- |
| `compose_run` | `RunState.trusted_config_digest` | `runs.write_trusted_config_digest` **+** the field, as a secondary |
| `compose_sweep` | same | same |
| `_resume_paused_run` re-stamp | same | same |

The read at resume prefers the state-root file and consults `RunState.trusted_config_digest` **only when that file is absent**. The field is still written — see the design point below for the two cases where it is the run's only surviving pin.

**Explicitly unchanged**, per the decision text:

- **The auto-sweep gate.** `_sweep_factory` has always compared against a baseline threaded in memory from launch, which nothing on disk can reach. Its closure carry is untouched.
- **`runsetup.config_digest`** — computation is unchanged; only storage moves.
- **The resume re-stamp** — still fires, still at the same moment, for the same reason (a human typed `resume`, which re-blesses the config; leaving the launch digest would make every later auto-sweep refuse after a legitimate resume-to-fix-a-setting).
- **`RUNS_DIR`** and the `tasks/` mailbox.

## The design point settled first

`state.json` is persisted data and a compatibility contract, so the question was whether to drop the field or keep it. **Kept and still written — as a secondary that is never preferred.** (This landed as read-only-for-one-release and was corrected in `33dc0fd` after codex round 2; see the review log.)

A run *paused under the old code* has its baseline in `state.json` and nowhere else. Reading only the state root would silently drop the pin for exactly the runs that sat idle longest — turning "this run has a pin" into "this run has none" on the first resume after the upgrade. That is the empty-means-legacy contract at `model.py` read backwards: the contract says *empty* means no prior pin, not that a *populated* field may be discarded.

The same "no file out of tree" answer arrives for a reason that has nothing to do with legacy runs, and this is what makes the field permanent rather than transitional: **the state root is keyed by the project's resolved path** (`project_tag` is `sha256(str(project.resolve()))`), so moving or renaming a project keys its runs somewhere new and orphans the subtree holding `config-digest`. `FEATURES.md` already documented that as the GC's known limit. `state.json` lives in the run dir and travels with it, so it is the copy that survives a move — and a `BMAD_LOOP_STATE_DIR` changed between launch and resume lands on the same path.

So the reader distinguishes two answers, and the distinction is load-bearing:

- `None` — no file out of tree. Means "this run predates #498, ask `state.json`".
- `""` — a baseline was stamped and it is empty. A real answer, and it must **not** reopen the agent-writable field.

Only `None` consults the field, which is what keeps rewriting the field pointless while the file is in reach — and is why the dual write costs #498 nothing. Empty keeps its meaning either way: no prior pin, no warning.

## Mechanics

- Written with `platform_util.atomic_write_text(..., follow_symlinks=False)`. The `follow_symlinks=False` is not incidental: the sibling events dir is handed to the driven session as `BMAD_LOOP_EVENTS_DIR`, so a link planted at this name must be replaced, never written through to a path of the session's choosing.
- **Read degrades, write raises.** Reading is observation feeding an advisory, and the resume is about to resolve the same state root for its events channel where the error is owned and reported; a resume that *aborted* because an advisory could not be read would be the worse failure. Writing is a repair write — a silently skipped stamp is the outcome hardest to detect later, because the next resume finds no file and decides on the in-tree copy alone, which is the tree this baseline exists to police.
- **GC is free.** The digest is a file inside the run's state dir, which `delete`/`archive`/`clean` and the orphan sweep already reclaim.
- **One ordering invariant touched, and its docstring updated.** `reconcile_orphan_state_dirs` reads state entries *before* live run-dir names on the strength of run dirs being created strictly first. The stamp is now the earliest writer into the state dir, so it is placed after `Journal` mkdirs the run dir (and before the pid lands, so a stamp that cannot be written fails the launch before an observer sees a live run). The docstring named `SignalWatcher` as the earliest creator; it now names both.

## What this buys, and what it does not

Sharpened after the codex review, which was right that the caveat is the main case rather than a corner.

**Closed: the incidental path.** The pin is no longer a project file, so nothing a session does in the ordinary course of rewriting the tree can collaterally blank it. That is precisely the case the advisory is documented to catch — "it catches the incidental change, not a deliberate one."

**Not closed: a pin at an old path key.** A project that moves away and later returns finds its old state-root subtree still there — unsweepable in between — and its stale pin is preferred over the blessing that travelled in `state.json`, for one resume, which then re-stamps and heals. Every tiebreak is unsound at equal privilege; **#572** carries the analysis.

**Not closed: the deliberate path.** Every shipped profile launches with permission bypass *by default* (`claude` `--permission-mode bypassPermissions`, `codex` `--dangerously-bypass-approvals-and-sandbox`, `gemini` `--approval-mode=yolo`, `copilot` `--allow-all-tools --allow-all-paths`, `antigravity` `--dangerously-skip-permissions`) — that is what an unattended loop is — and the session is handed `BMAD_LOOP_EVENTS_DIR`, whose parent is the state dir. A session that goes looking can *truncate* the baseline (reader answers `""`, a real "no baseline") or delete it **and** blank the in-tree secondary, and either silences the warning. The dual write raises that cost from one deletion to two edits, which is a speed bump, not a boundary — both are in reach.

There is no sound fix at equal privilege, which is why this PR states the limit instead of half-closing it: any marker distinguishing "this run should have a baseline" from "genuinely legacy run" must itself live somewhere the same session cannot reach, and at equal privilege no such place exists — in `state.json` the session edits both, in the state dir it is deleted with the file. Warning whenever *no* baseline exists instead fires a false alarm on every legacy run's first resume, on a warning whose value is that it is rare.

The direction that would work is **privilege separation on the state root** — which also covers the events channel's identical exposure (a session can delete `events/` and stall its own run) and so is worth deciding as one piece. Filed as **#571**; the limit is now stated in `runs.config_digest_path_for`'s docstring and the CHANGELOG entry, so the claim on record matches what the code delivers.

## Tests

Existing digest tests moved with the storage (4 asserts in `tests/test_cli.py` now read the state root). Added:

- **The mechanism itself** — config changed *and* `state.json`'s field blanked, exactly as one session could do it; the warning must still fire.
- **The rename path** — a run whose project is renamed between resumes still warns on a changed host-exec config, with the precondition asserted (the tag really did change and the out-of-tree read really does answer `None`), so it cannot pass for the wrong reason.
- **Both launch sites stamp both copies** — asserted on the state the launch composed, alongside the existing out-of-tree stamp assert.
- **The migration path** — a pre-#498 run compares against its in-tree field, warns, and comes out with the baseline out of tree.
- **Placement** — the digest lands at `<state dir>/config-digest` and nothing lands under `.bmad-loop/`.
- Absent-vs-empty; the read-degrades/write-raises asymmetry (3 rows: unnameable root, unresolvable project, symlink-loop `RuntimeError`); a planted symlink is replaced; the state-dir GC reclaims it.

**Every negative assertion was ablated** (harness requiring pytest rc==1 — rc=4 is a collection error, not a reddening; flags ablated singly, since a combined revert lets one coast). The harness caught two of its own:

- the first GC test used `delete_run`, which removes the run dir too, so it passed whether the digest was in or out of tree. Rewritten against `reconcile_orphan_state_dirs`, which reaches only out-of-tree state, and it reddens.
- the first precedence control for the round-2 fix ablated to `state.trusted_config_digest or read(...)`, which falls straight through on a *blanked* field and so never modelled "prefer the in-tree copy" at all — it passed while proving nothing. Re-specified as an outright pre-#498 read, and it reddens.

Round-2 rows: dropping the resume re-stamp, the `compose_run` pass-through, or the `compose_sweep` stamp each redden their own test; and the precedence is pinned in **both** directions — making the in-tree copy authoritative, or letting an empty field override a populated file, each redden #498's own attack test. That second direction is the one that matters after adding a second copy: it is what proves the original guarantee did not go vacuous.

## Verification

At `e2a6ee9`: `uv run pytest -q` 5324 passed / 44 skipped / 5 xfailed · `uv run pyright` 0 errors · `trunk fmt` + `trunk check --all --no-fix` clean.

## Review log

- **codex round 1** — one P1, "reject missing or empty external baselines". Confirmed on mechanism and materiality (bypass is the default, not an edge case), and it found a real defect in this PR: the CHANGELOG headline overstated the fix and contradicted its own caveat. Fixed in `43e92cf`, which restates both the entry and the docstring in terms of the incidental-vs-deliberate split. The suggested remedy — treat absent/empty as tampering — is declined on the stated boundary: it cannot be made sound at equal privilege, and the direction that can is #571.
- **CodeRabbit round 1** — one Major, "reject link and special-file digest reads". Confirmed: a genuine asymmetry in this PR's own work, which hardened the *write* with `follow_symlinks=False` and left the *read* on `Path.read_text` — a planted FIFO wedges `resume` indefinitely and a link to `/dev/zero` raises `MemoryError`, not the `OSError` the reader promises never to leak. Fixed in `1436ec4`, mirroring `tui.launch._read_ctl_window`: `O_RDONLY|O_NOFOLLOW|O_NONBLOCK`, `S_ISREG` on the **descriptor** (an `fstat` cannot be raced), bounded read. Unlike that twin it must not collapse empty to `None`.
- **codex round 2** — one P2, "preserve the pin when the project state key changes". Confirmed, and it is a **regression this branch introduced**, on a documented operator action with no tampering involved: renaming a project re-keys the state root and orphans the digest, and with the field no longer written the fallback was empty, so a changed host-exec config was accepted in silence. Verified against the base commit, which warns across the same rename. Fixed in `33dc0fd` by writing both copies with the read precedence unchanged; #498's attack test still passes untouched. The residual — a session deleting the file *and* blanking the field — stays on the round-1 boundary and with #571.
- **codex round 3** — one P2, "reject stale pins when a project returns to an old path". Confirmed on both harms (a false warning on an unchanged config; and silence when the config is reverted to the stale pin's own value), and **not** introduced by round 2's dual write — probed at `1436ec4`, where the stale pin is preferred and the false warning fires identically. Documented rather than fixed, in `e2a6ee9`: every tiebreak is unsound at equal privilege (preferring the session-writable in-tree copy hands back the silencing #498 closed; a generation number needs an unforgeable counterpart that does not exist; retiring the old copy needs the old key, which is either a session-supplied path feeding a delete or a cross-project delete keyed on run id). Same wall as #571, reached by re-keying rather than tampering. Filed with evidence and repro as **#572**.

## Summary by CodeRabbit

* **Security**
  * Trusted resume configuration baselines are now stored in the run’s external state directory.
  * Secure file handling protects baselines from unsafe files and symlink replacement.

* **Bug Fixes**
  * Existing runs automatically migrate from legacy project-state baselines.
  * Resume integrity checks continue working after project or state-directory moves.
  * Cleanup now removes associated control-plane data and orphaned run directories.

* **Documentation**
  * Updated disk-reclamation guidance to clarify baseline storage, fallback behavior, cleanup, and storage estimates.


pbean:
@codex review

pbean:
**P1 "Reject missing or empty external baselines" — confirmed on mechanism, and it found a real defect. Partially actioned; the suggested remedy declined on a stated boundary.**

Verified rather than taken on faith, and the premise is stronger than the finding claims: bypass is not a configuration some users pick, it is the **default** for every shipped profile — `claude.toml` `--permission-mode bypassPermissions`, `codex.toml` `--dangerously-bypass-approvals-and-sandbox`, `gemini.toml` `--approval-mode=yolo`, `copilot.toml` `--allow-all-tools --allow-all-paths`, `antigravity.toml` `--dangerously-skip-permissions`, applied by `GenericAdapter.interactive_argv` unless `[adapter] extra_args` overrides. Both silencing paths work as described: delete → reader returns `None` → legacy fallback reads `""` on any post-#570 run; truncate → `""`. Either way `bool(pinned)` is false.

**What this legitimately caught: my own overstated claim.** The CHANGELOG headline read "A session can no longer silence resume's config-change warning" and contradicted the caveat three sentences below it. That is fixed in `43e92cf`, which restates the entry and `config_digest_path_for`'s docstring in terms of the split that is actually true — the **incidental** path is closed (the pin is no longer a project file, so nothing done in the ordinary course of rewriting the tree can collaterally blank it, which is the case the advisory is documented to catch), the **deliberate** path is not.

**Why the remedy is declined here.** "Treat absent/empty/unreadable as tampering for a post-migration run, while retaining a way to identify genuinely legacy runs" requires a marker for *should this run have a baseline*, and that marker has to live somewhere the same session cannot reach. At equal privilege no such place exists: in `state.json` the session edits both; in the state dir it is deleted along with the file. The fallback shape you are pointing at is not an accident of the `None`/`""` split — it is that constraint surfacing.

The alternative of warning whenever no baseline exists at all trades a silent hole for a false alarm on every genuinely legacy run's first resume, on a warning whose entire value is that it is rare.

The direction that does work is **privilege separation on the state root** — the control plane owned by a principal the session does not run as. That also covers the events channel's identical exposure (a session can delete `events/` and stall its own run), so it wants deciding as one piece rather than as a marker bolted onto the digest. Filed with the full analysis as **#571**.

Scope note: this PR executes a recorded 2026-08-08 maintainer decision that is specifically "relocate the digest to a user-scoped state root keyed by (project, run_id)". Tampering detection is a new requirement on top, not part of that decision.

pbean:
@codex review

pbean:
@coderabbitai review

pbean:
**"Reject link and special-file digest reads" — confirmed and fixed in `1436ec4`.**

This one is right, and it caught a real asymmetry in my own work: I hardened the *write* with `follow_symlinks=False` and reasoned explicitly about planted links there, then left the *read* on `Path.read_text`. The file's parent is the `BMAD_LOOP_EVENTS_DIR` the engine hands the session, so degrading on a hostile path is not enough when the read itself is the weapon.

`read_trusted_config_digest` now establishes the shape before consuming bytes:

- **`O_NONBLOCK` + `S_ISREG` on the descriptor.** The FIFO case is the sharpest: opening one for reading blocks indefinitely, and `resume` is a foreground command a human is waiting on, so a planted FIFO wedges the terminal rather than costing a warning. The check is on the fd, not the path, so it cannot be raced.
- **`O_NOFOLLOW`** — the name is read, not wherever it points.
- **A 256-byte cap** — a link to `/dev/zero` otherwise reads to `MemoryError`, which is not the `OSError` this function promises never to leak. The cap removes the condition rather than absorbing it.

POSIX-only flags degrade to 0 on win32, where the size cap and regular-file check carry alone. This mirrors `tui.launch._read_ctl_window`, which solved the identical hazard for the ctl-window record — same shape, one idiom, and the docstring says so. It deliberately does *not* collapse empty to `None` the way that twin does: here `""` and `None` are different answers and the legacy fallback turns on the difference.

**Ablated singly**, since a combined revert would have let one flag coast: dropping `O_NONBLOCK`, `O_NOFOLLOW`, `S_ISREG`, or the cap each reddens its own row, and the absent-vs-empty control still passes under the full revert. The FIFO row is guarded by `SIGALRM` rather than a `multiprocessing` join — the first draft used `Process`, whose default `fork` start method on Linux warns on 3.12+ and risks a child deadlock, i.e. it traded a hang under ablation for a possible hang in the ordinary run.

**Scope note on the mergeability rationale.** The remaining exposure — a session *deleting* the baseline to silence the warning — is not closed by this and cannot be closed at equal privilege; that is tracked in #571 with the analysis. What `1436ec4` closes is the denial-of-service half, which is separable and fixable, and it was worth separating.

One thing your review surfaced that is **not** in this PR's diff: `signals.py`'s `SignalWatcher.poll` reads event files in the same directory with `entry.read_text(...)`, so the events channel has the identical exposure — pre-existing from #494/#568, not introduced here. Adding it to #571 rather than widening this PR.

pbean:
@codex review

pbean:
**P2 "Preserve the pin when the project state key changes" — confirmed, and it is a regression this branch introduced. Fixed in `33dc0fd`.**

This one is right, and it is a different animal from the P1 in the last round. That one needed a session to go tampering; this needs nobody to do anything wrong at all — an operator renames a directory and a security advisory goes quiet forever.

I verified the mechanism rather than reasoning about it, then verified it was a *regression* by running the same probe against the base commit:

| | out-of-tree read | in-tree field | warning on a changed host-exec config |
|---|---|---|---|
| head `1436ec4`, no move | populated | `""` | **fires** |
| head `1436ec4`, after rename | `None` | `""` | **silent** ← the defect |
| base `1878289`, after rename | *(n/a)* | populated | **fires** |

`project_tag` is `sha256(str(project.resolve()))`, so the rename keys the run somewhere new (`d7653555bc27fe3f` → `f076ba90ab0d1df6`) and orphans the subtree holding `config-digest`; `state.json` lives in the run dir and travels with it. The old subtree is still on disk, and by the same token unreachable — `FEATURES.md` already documented that as the GC's known limit, which is exactly the fact this finding turns on. `rc == 0` on the post-move resume, so the path is reachable, not theoretical. Your second scenario, a `BMAD_LOOP_STATE_DIR` that changes between launch and resume, lands on the same code path.

**The remedy is yours, and it does not cost #498 anything** — that was the thing worth checking before taking it. Both launch sites and the resume re-stamp now write the in-tree copy alongside the out-of-tree file, and the read precedence is *unchanged*: `_resume_paused_run` reads the file first and consults `state.trusted_config_digest` only when the file is absent (`None`). So the in-tree copy is never preferred while the file is in reach, which is what keeps rewriting it pointless — #498's attack (rewrite `policy.toml`, blank the field) still warns, and its test still passes untouched. In the tamper model the dual write is never worse and is better in one branch: deleting the file alone no longer silences anything, it now takes deleting the file *and* blanking the field.

The `None` vs `""` split is untouched and still load-bearing: a *truncated* file is a real "no baseline" and must not reopen the tree, so only `None` falls back.

Ablated singly, requiring `rc==1`:

- drop the resume re-stamp → the new rename test reddens
- drop the `compose_run` pass-through → the run-launch stamp assert reddens
- drop the `compose_sweep` stamp → the sweep-launch stamp assert reddens

and the precedence pinned in **both** directions, because "we wrote a second copy" is the kind of change that quietly makes the original guarantee vacuous:

- make the in-tree copy authoritative (the pre-#498 read) → #498's attack test reddens
- let an empty in-tree field override a populated file → #498's attack test reddens

My first attempt at that second control was a **fake green** and worth naming: I ablated to `state.trusted_config_digest or read(...)`, which falls straight through on the blanked field, so it never modelled "prefer the in-tree copy" at all and passed while proving nothing. The control was mis-specified, not the code — but a control that passes for the wrong reason is indistinguishable from a guarantee that holds, which is the whole hazard.

Docs corrected with the code: the CHANGELOG said "The field is no longer written", which this makes false, and `RunState.trusted_config_digest` was labelled `LEGACY, read-only, one release` — it is now the travelling secondary, permanently. `FEATURES.md`'s known-limit note now says a move does not cost the run its baseline.

Not fixed, and still #571: a session that *deletes* the out-of-tree file and blanks the in-tree copy. That needs a marker where the session cannot reach, and at equal privilege there is nowhere — unchanged from the last round.

Gates at `33dc0fd`: pytest 5324 passed / 44 skipped / 5 xfailed, pyright 0 errors, `trunk check --all --no-fix` clean.

pbean:
@codex review

pbean:
@coderabbitai review

pbean:
**P2 "Reject stale pins when a project returns to an old path" — confirmed on both harms, and not fixed here. Documented on a stated boundary in `e2a6ee9`; filed as #572.**

Both halves reproduce, so this is taken at full strength rather than talked down:

- **False warning** — A → B → A with a real resume at each stop, config untouched since B's blessing: the resume reads A's stale pin and reports the host-exec config changed. It did not.
- **Missed warning** — revert the config to the value A's stale pin holds and the resume goes *silent*, although the config did change relative to the last blessing. Your "worse" is the right word for it.

Both self-heal, because the resume re-stamps the current key — one wrong answer per return trip.

**One correction to the framing, and it is the reason this is not a fix on this branch.** The finding reads as if the round-2 dual write introduced it. It did not. Probed at `1436ec4`, before that commit, with the in-tree field unwritten:

| | out-of-tree read at A | in-tree copy | false warning |
| --- | --- | --- | --- |
| `1436ec4` (out-of-tree only) | stale `digest(C1)` | `""` | **yes** |
| `33dc0fd` (dual write) | stale `digest(C1)` | `digest(C2)` | **yes** |

Identical. The stale pin is preferred at both commits; the dual write only made a fresher copy visible beside it. What that commit *did* change on this path is the other leg — at `1436ec4` the resume at B does not warn at all, and now it does.

**Why I am not taking the remedy.** Each option is either unsound at equal privilege or worse than the bug:

- *Prefer the fresher-looking in-tree copy.* It is session-writable. Letting it beat the out-of-tree file hands every session precisely the silencing #498 exists to close — the one thing this PR must not trade away.
- *Associate the pin with a generation.* The file can carry one, but concluding "a newer blessing exists elsewhere" needs a counterpart the session cannot forge, and the only candidate is the in-tree copy. Same wall.
- *Retire stale state-root copies when re-keying.* This needs the old key. Taking it from `state.json` feeds a session-supplied path into a delete — a destructive primitive strictly worse than a spurious advisory. Taking it by enumerating the state root makes it a cross-project delete keyed on run id, which silences a live run's pin on any collision.

So it is the same wall #571 names — the identity and privilege model of the state root — reached by re-keying rather than by tampering. A pin whose location is stable across a project's lifetime, rather than derived from a path that can change and change back, has no stale sibling to arbitrate; that is a decision to take with #571, not a patch to bolt on here.

`e2a6ee9` records the limit in `read_trusted_config_digest`'s docstring next to the #571 one, and adds the caveat to the CHANGELOG so the claim on record does not overstate what the move buys — which is the defect round 1 caught, and I would rather not repeat it. Full evidence and repro on **#572**.

Gates at `e2a6ee9`: pytest 5324 passed / 44 skipped / 5 xfailed, pyright 0 errors, `trunk check --all --no-fix` clean.

pbean:
@codex review

pbean:
@coderabbitai review

pbean:
**Both actionable comments landed on real gaps in my own tests. One premise corrected, both fixed in `3454bee`.**

**1. "Bound the read, not just the result" — confirmed, and worse than filed.** `..._is_bounded` asserted only `len(got) == _MAX_DIGEST_BYTES`, and its docstring claimed it pinned the bound on the *read* while naming the exact reader that would slip through: one that slurps the file and truncates at the end. That reader returns 256 bytes too. Ablated it — **rc=0, it passed** — so the docstring was asserting something the test did not check, which is the same overstatement defect codex caught in round 1, this time inside a test. Requested counts are now captured at `os.read` and totalled; that row fails at `assert 1114112 <= 256`. The returned-length assert stays, because it is what catches a cap honoured on the read and then lost.

**2. "The FIFO test does not exercise `S_ISREG`" — premise refuted, but the variant you asked for is worth having, so it is in.** Ablating the descriptor check *does* redden the existing test:

```
E  AssertionError: assert '' is None
```

An idle FIFO opened `O_RDONLY|O_NONBLOCK` reads EOF rather than `EAGAIN`, so without the check the reader answers `""` where it owes `None` — the guard is graded today.

What the idle case does *not* model is the attack. Those bytes being empty is an accident of there being no writer; with one holding the FIFO open and feeding it, a reader that got as far as `os.read` returns **whatever the session piped in and treats it as this run's baseline** — feed the digest of the config just installed and `resume` is satisfied; feed noise and the operator is warned off a change nobody made. Neither is a hang, so the sibling's `SIGALRM` would never notice. Added as a twin, `O_RDWR` on the holder because a write-only open on a FIFO blocks until a reader arrives and would wedge the test itself.

Ablated singly, since a combined revert lets one coast:

| ablation | idle-FIFO | fed-FIFO | bounded |
| --- | --- | --- | --- |
| slurp whole file, truncate at end | — | — | **reddens** (was green) |
| uncapped `os.read` | — | — | **reddens** |
| drop `S_ISREG` | **reddens** | **reddens** | — |

Gates at `3454bee`: pytest 5325 passed / 44 skipped / 5 xfailed, pyright 0 errors, `trunk check --all --no-fix` clean.

For the record on coverage: this was your first review of this branch since `43e92cf` — the three earlier heads were rate-limited, and the summary comment is edited in place on walkthrough refreshes as well as on the limit stub, so its timestamp is not a coverage signal. The incremental pass over `43e92cf..e2a6ee9` picked up both of these, which is exactly the span that had been skipped.

pbean:
@codex review

pbean:
@coderabbitai review

PR #574, opened 2026-08-13
fix(worktree): journal undelivered wheel MODULE_SKILLS result-side

### LINKED ISSUE #464: Wheel-provided MODULE_SKILLS have no result-side completeness predicate, so a partial seed into a unit worktree is unobservable
Restated rather than fixed by PR #449 (Phase 6H of #433), and currently tracked by nothing. Filed
out of #433's close-out so it does not vanish with the tracking issue.

## What is unguarded

`provision_worktree` copies the bundled wheel skills into each CLI's skill tree
(`src/bmad_loop/worktree_flow.py`, the `for skill in MODULE_SKILLS` loop) with `skip_existing=True`
and best-effort per-entry degradation. Every *other* surface it seeds gained a result-side
predicate that re-probes disk after the copy:

| surface | predicate |
| --- | --- |
| upstream skill trees | `base_skills_seed_incomplete` → CRITICAL escalation |
| `_bmad/scripts` | `_bmad_scripts_seed_incomplete` → CRITICAL escalation |
| `_bmad/config.toml` | `_central_config_seed_incomplete` → CRITICAL escalation |
| `seed_files` / `seed_globs` / hook configs | `worktree_seed_undelivered` → `worktree-seed-dropped` |
| **wheel `MODULE_SKILLS`** | **none** |

So a wheel skill that lands partially — or not at all — in a unit worktree is invisible. The copier
degrades per entry by design (6G made it total across filesystem faults), which is exactly what
turns a partial copy into silence rather than a crash.

## Why it was deliberately excluded from the existing gate

Quoting the rationale left in the source at the copy site: story worktrees have a determinate
dev/review consumer contract for the *upstream* skills — the orchestrator dispatches a known
primitive and known review layers, so an absent contract file provably means the session writes
nothing and a CRITICAL pause is honest. The bundled operator/triage surface is a different
invocation shape: nothing in a story session is guaranteed to invoke it. Reusing the
deterministic-stall gate would therefore risk refusing healthy runs.

That reasoning is sound, which is why this is not "just add it to the gate".

## What closing it needs

A required-consumer contract for the bundled surface — i.e. deciding, per skill, what a story
session is actually entitled to invoke, and only then what absence proves. Plausible shapes, in
increasing cost:

1. **Journal-only**, mirroring `worktree-seed-dropped`: re-probe `MODULE_SKILLS` after the copy and
   report what did not arrive. Cheap, never refuses a healthy run, and makes the failure
   diagnosable instead of silent. This is probably the right first step.
2. **Gate the subset with a real consumer.** `bmad-loop-sweep` is dispatched by sweep runs, so its
   absence in a *sweep* unit worktree is a determinate stall the same way the dev primitive's is.
   That is a narrower, defensible escalation than gating the whole catalogue.
3. **Full contract**, per skill and per run mode.

## Notes for whoever picks this up

- The predicate must re-probe **disk**, never the copier's bookkeeping — the existing gates are
  written that way on purpose so a user-authored `scm.worktree_seed` entry that happens to spell a
  skill rel cannot forge a pause.
- Wheel skills come from `importlib.resources`, so the source may be a zip Traversable; the
  predicate has to answer through the same shared walk the copier uses rather than assuming
  filesystem paths.
- Related but distinct: #415 covered the `seed_files` leg and #422 the totality work. Neither
  touches this surface.

Reference: #433 (Phase 6H), PR #449 "Known gap".

pbean:
## What was unguarded

`provision_worktree`'s `MODULE_SKILLS` copy loop (`worktree_flow.py:639` on main) called
`_copy_traversable` and discarded its return. Every other seeded surface in an isolated worktree
has a result-side completeness probe; the wheel's own bundled `bmad-loop-*` skills had none, and
the code said so — a "Known gap, deliberately restated" comment sat directly above the loop.

So a wheel skill that seeded *partially*, or not at all, was silent. The realistic trigger is a
checkout that carries a non-directory where a bundled skill's directory goes: `_copy_traversable`'s
no-clobber prunes the whole subtree at its root (a file or dangling link squatting a directory
target wins, and `mkdir` must never replace it or write through it), so the loop lands nothing and
reports nothing. Same for a destination symlinked out of the worktree, or a source directory the
filesystem refuses to enumerate.

## What ships

`module_skills_seed_undelivered(worktree, trees, skills_root=None)` — a journal-only predicate,
plus a `worktree-module-skills-dropped` append in `WorktreeFlow.run_isolated`.

- **Re-probes disk, never the copier's bookkeeping.** A user-authored `scm.worktree_seed` entry
  that happens to spell a skill rel can therefore neither forge nor mask a report. Its own journal
  kind, not folded into `worktree-seed-dropped`, for the same reason.
- **Enumerates through `_walk_traversable_files`, not `rglob`.** The source is the wheel's skills
  tree, which may be a zip `Traversable` with no filesystem path.
- **Presence is the whole contract; content is NEVER compared.** Seeding is per-*file* no-clobber,
  so a checkout carrying its own divergent fork of a bundled skill keeps those bytes while its
  absent siblings are filled in. Comparing content would report that healthy shape as a drop —
  the same reason `worktree_seed_undelivered` does not compare content either.
- Delivery requires a `_is_file` destination **contained in the worktree** (the unit commits its
  own checkout; content behind a link out of it was never delivered). A directory entry is yielded
  by the walk only when `iterdir` was *refused*, so it counts as undelivered — unknown descendants
  cannot be claimed as delivered. A skill the wheel itself lacks is skipped: a broken wheel is the
  module-skills sync test's concern, not a run's.

## Why journal-only and not a gate

**No `MODULE_SKILLS` entry has a worktree-resident consumer**, so an absence here cannot prove a
stall, and a CRITICAL gate would refuse *healthy* runs:

- `bmad-loop-sweep` — the triage session that consumes it runs at `workspace.root`, the main
  checkout, never a unit worktree; `sweep.py` swaps the workspace only around bundle execution
  (`runsetup.py:291-294`).
- `bmad-loop-resolve` — interactive, dispatched at the main checkout.
- `bmad-loop-setup` — no session consumer at all.

That is the whole difference from `base_skills_seed_incomplete`, which *does* escalate: the
upstream dev/review skills are dispatched **inside** the worktree, so their absence is a determinate
stall on `Unknown command` with nothing written to disk.

**Gate extension point**, documented in the predicate's docstring: if a consumer ever becomes
worktree-resident (e.g. sweep triage moving into unit worktrees), arm the gate by routing this
predicate's result into `WorktreeFlow.escalate_unit` for that consumer's required subset. The
predicate is the reusable half; only the caller decides severity.

## Test layers

Six predicate rows in `tests/test_install.py` (real provision + disk probe; a zip `Traversable`
source; a divergent fork counting as delivered; a squatting file; a destination symlinked out of
the worktree; an unreadable source directory, modelled with an in-test stub rather than
`chmod 0o000`, which is a no-op for root on CI).

One engine row in `tests/test_engine_worktree.py` — the wiring, which the predicate rows cannot
see. It drives a full isolated run against the conftest sandbox with the mock adapter, and the
obstruction is the real copier's: a plain FILE **tracked** at the skill's destination rel rides
`git worktree add` into the checkout, where no-clobber meets it. `attach_profile` is required
rather than decorative — `MockAdapter` deliberately carries no profile, `worktree_profiles` then
yields no skill trees, and the copy loop and predicate would never run at all.

### Ablation drill (executed, not asserted)

| Ablation | Result |
| --- | --- |
| Delete the `module_skills_seed_undelivered` call in `run_isolated` | Engine row FAILS (`len(dropped) == 1` → `0 == 1`), rc 1. **The six predicate rows stay green** — they call the predicate directly, which is precisely the coverage gap this PR closes. |
| Route the result into `escalate_unit` instead of the journal | Engine row FAILS on `summary.done == 1` (`escalated=1, paused=True`), rc 1 — the "never escalated" half is load-bearing. |
| Drop the predicate's `_is_file(target)` conjunct | 3 of 6 predicate rows FAIL **and** the engine row FAILS, rc 1 — both layers bite on the predicate itself, not only on its wiring. |

`worktree_flow.py` was restored from a copy held outside the repo (never `git checkout`) and
re-verified byte-identical by checksum after each ablation.

Full suite 5332 passed / 44 skipped / 5 xfailed; `pyright` 0 errors; `trunk check --all` clean.

Closes #464

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved worktree provisioning checks for bundled module skills.
  - Missing or partially copied module skills are now recorded as informational events rather than stopping runs.
  - Existing checkout-provided skill variants are correctly recognized as available.
  - Critical missing development or review skills continue to trigger escalation.

- **Documentation**
  - Updated feature documentation and changelog details for worktree skill verification and reporting.


pbean:
@codex review

pbean:
@coderabbitai review

PR #580, opened 2026-08-13
fix(deferredwork): atomic ledger writes + surrogate-safe sanitizer (#328, #329)

### LINKED ISSUE #328: deferredwork: append_decision/append_entry bare write_text can truncate the ledger to zero bytes
## Summary

`mark_done_many` writes the ledger through `atomic_write_text` (`deferredwork.py:354`) with a
documented rationale — a bare tmp+replace resets the file mode and turns a symlinked ledger into a
regular file. Its two sibling writers do not:

- `append_decision` — `deferredwork.py:389`: `path.write_text(text, encoding="utf-8")`
- `append_entry` — `deferredwork.py:492`: `path.write_text(text + sep + block, encoding="utf-8")`

`Path.write_text` opens with mode `'w'` (truncate) and *then* encodes. Any failure between those two
points — an unencodable value, `ENOSPC`, `EIO` — leaves a **zero-byte ledger**. Every deferred-work
entry is gone.

Reproduced during review: an `append_decision` call that raised mid-write left a 35-byte ledger at
0 bytes, while the same value through `mark_done` left it intact.

## Why the asymmetry matters

`mark_done_many`'s docstring already argues the case for this exact file — "swapping a fresh inode
over the ledger otherwise resets its mode (a `0600` ledger silently becoming world-readable) and
turns a symlinked ledger into a regular file". That reasoning applies unchanged to the other two
writers; the protection just was not extended to them.

An asymmetric guard between siblings is itself the finding: a reader cannot tell whether the
difference is deliberate.

## Fix shape

Route `append_decision` and `append_entry` through `atomic_write_text` too. Note `append_entry`
creates the file when absent (`path.parent.mkdir(...)`), so the helper needs to keep working for a
not-yet-existing target.

Found during the #305 review.

### LINKED ISSUE #329: deferredwork: a lone surrogate defeats _one_line's never-raises contract and crashes the sweep
## Summary

`deferredwork._one_line` (added in #305) guards **line breaks** and documents a stronger contract
than it delivers:

> Sanitizes; never raises. The close paths call these writers bare (`sweep._close_resolved`,
> `decisions.apply_pre_answer`), so a `ValueError` here would end the sweep as crashed.

A lone surrogate defeats that without ever being a line break. Verified chain:

1. `json.loads` on a session `result.json` containing `"\ud800"` yields the surrogate.
2. `validate_triage` returns `errors == []` and carries it into `ResolvedEntry.evidence`.
3. `LINE_BREAK_RE.search` is `False`, so `_one_line` passes it through untouched.
4. The write raises `UnicodeEncodeError` — a `ValueError` subclass — at `sweep.py:942`
   (`_close_resolved`), `sweep.py:1071`, or `decisions.py:147`.

That is precisely the outcome the sanitizer exists to prevent: a bare call on a close path taking
down the run. It compounds with the bare `write_text` issue, where the same call both raises **and**
truncates the ledger to zero bytes.

Pre-existing — the writers always encoded UTF-8 — but #305 introduced the "never raises" claim, so
either the claim or the code should change.

## Fix shape

Either narrow the docstring to "never raises **on a line break**", or make it true:
`value.encode("utf-8", "replace").decode("utf-8")` before the break collapse. The latter keeps the
close paths genuinely crash-free, which is the doctrine #305 settled on.

Also worth checking: `tui/app.py:_record_decision` now catches `ValueError`, so the TUI degrades —
but the sweep close path has no such guard.

Found during the #305 review.

pbean:
Closes #328
Closes #329

## Problem

Two hazards on the same code path, and they compound.

**#328 — a failed write empties the ledger.** `Path.write_text` opens the file `'w'`
(truncate) and only *then* encodes and writes. Any failure inside that window — an
unencodable value, `ENOSPC`, `EIO` — leaves a zero-byte file where every hand-authored
deferred-work entry used to be. `mark_done_many` already wrote through
`atomic_write_text`; its siblings did not, and an asymmetric guard between siblings is
itself unreadable.

**#329 — a lone surrogate defeats the sanitizer's never-raises contract.**
`_one_line` collapses line breaks and documents "Sanitizes; never raises". A lone
surrogate is not a line break, so it sailed through untouched — and it has no UTF-8
encoding at all, so `atomic_write_text`'s strict encode raises `UnicodeEncodeError`
(a `ValueError` subclass) from close paths that call these writers bare. The revival
chain is reachable, not theoretical: `sweep._ensure_triage` caches the session's
`result.json` with `json.dumps`, whose `ensure_ascii` keeps the code point a harmless
`\ud800` escape on disk; a later cycle reloads it and `json.loads` revives the real
surrogate; `validate_triage` passes it through (`.strip()` does not remove it) into
`ResolvedEntry.evidence`; `_close_resolved` interpolates that into the `mark_done`
note, which reaches `_one_line` and then the encode.

**The coupling.** Before this branch, that same surrogate through an append writer did
both at once: the truncate had already landed when the encode raised, so the run
crashed *and* the ledger it was appending to went to zero bytes in a single call. #328
alone would keep the bytes but still crash the sweep; #329 alone would stop the crash
but leave every other failure mode (`ENOSPC`, `EIO`) able to empty the file. Fixing one
without the other leaves half the failure standing, which is why they ship together.
`tests/test_deferredwork.py::test_append_entry_encode_failure_cannot_truncate_the_ledger`
pins exactly that compounding.

## Fix

**`src/bmad_loop/platform_util.py`** — new `neutralize_surrogates(text)`: replaces every
code point in `[\ud800-\udfff]` with U+FFFD. Matched as a code-point *range*, not by an
encode round trip — a Python `str` holds astral characters as one code point, so U+1D11E
is never touched. Replace rather than strip (a vanished value reads as absent, not as
damaged) and rather than `encode("utf-8", "replace")` (that yields `?`, indistinguishable
from one the author typed). Clean text returns the identical object, so a clean write
stays byte-identical. It lands next to `atomic_write_text`, the strict encode it exists
to protect.

**`src/bmad_loop/deferredwork.py`** — `_one_line` runs `neutralize_surrogates` first, so
every free-text ledger field is covered at one chokepoint: `append_decision`'s
`label`/`detail`, `append_entry`'s `title`/`origin`/`source_spec`/`reason`/`location`,
and `_apply_done`'s `note`. Both passes keep their own fast path, so a clean value is
scanned twice and copied never. Sanitizing stays *in front of* the idempotence scan, or a
replayed defer whose origin carries a surrogate would match nothing on disk and burn a
fresh id every run. `append_decision` and `append_entry` now write through
`atomic_write_text`, matching `mark_done_many`.

**`src/bmad_loop/engine.py`** — `_defer`'s post-rollback ledger restore writes through
`atomic_write_text`. A repair write must never be the thing that empties the file it
exists to put back.

**`src/bmad_loop/sweep.py`** — `_ensure_migration`'s migration-failure ledger restore
writes through `atomic_write_text`. `_write_intent` writes `bundles/<dir>/intent.md`
through `atomic_write_text` *and* neutralizes surrogates over the whole document rather
than per field — line breaks are legitimate markdown there, so `_one_line`'s collapse
would be damage. The verbatim ledger blocks it re-attaches come from a strict-UTF-8 read
and pass through byte-unchanged.

Call sites stay bare — repair writes raise, never degrade.

## Scope note

#328 names the `append_decision`/`append_entry` asymmetry against `mark_done_many`. The
two engine/sweep restore writes and the `intent.md` write were folded in with maintainer
approval: same hazard class, same one-line remedy, and leaving them out would have
recreated the asymmetry the issue is about one layer over. Five write sites in total,
enumerated in the Fix section. Nothing else in the diff.

Deliberately **not** touched: the `json.dumps` writes in `sweep.py` (manifest,
`migrate-result.json`, cached `result.json`, decision answers). Those are ASCII-only by
`ensure_ascii` and are machine-minted caches, not operator-curated state — that is the
very property that lets a surrogate hide on disk as an escape in the first place.

## Visible behavior diffs

- **A ledger created from nothing by `append_entry` now lands mode `0600`** instead of
  `0644 & ~umask`. `atomic_write_text` has no prior mode to carry over for a fresh
  target, so it keeps `mkstemp`'s private mode — deliberate, per that helper's docstring,
  and the safer end for a file that can carry unreleased-finding text. Verified on this
  box under `umask 022`: `oct(stat.S_IMODE(...)) == '0o600'`.
- **A surrogate now renders as `�` (`REPLACEMENT CHARACTER`) in ledger fields and in
  `intent.md`** rather than crashing the write. A surrogate-only title therefore stays
  truthy, so `append_entry`'s `(untitled DW-<n>)` substitution deliberately does *not*
  fire for it — the field says *something unencodable was here* instead of going silently
  blank.
- Clean values are unchanged byte-for-byte: both sanitizer passes have a fast path, and
  `atomic_write_text` keeps `write_text`'s translating-newline default, so line endings
  do not shift on Windows.

Scope of the durability claim: **the ledger writers are now crash-free on text content.**
That covers `append_decision`, `append_entry`, and the `mark_done`/`mark_done_many`/
`mark_done_many_reopenable` family — every caller-supplied free-text field they
interpolate routes through `_one_line` — plus `mark_open`, which interpolates no caller
text at all (it writes back a status line it decoded from hex under a strict UTF-8
decode). It is **not** a claim about all close paths or all writes in the orchestrator:
the JSON caches above and every writer outside `deferredwork.py` are untouched and
unaudited here.

## Testing

Full suite green locally (`uv run pytest -q -n auto`: 5388 passed, 44 skipped, 5 xfailed),
`uv run pyright` 0 errors, `trunk check --all` clean.

Every guard below was ablated — the fix reverted in place, the target run, the file
restored from a `cp` backup (never `git checkout`, which would have taken the rest of the
branch with it). One line of evidence each; the tree was verified pristine afterwards.

**#328 — atomic writes**

| Test | Ablation evidence |
| --- | --- |
| `test_deferredwork.py::test_append_decision_write_failure_raises_and_keeps_the_ledger` | `append_decision` → bare `write_text`: `Failed: DID NOT RAISE OSError` (the write no longer routes through the patched helper). |
| `test_deferredwork.py::test_append_entry_write_failure_raises_and_keeps_the_ledger` | `append_entry` → bare `write_text`: `Failed: DID NOT RAISE OSError`. |
| `test_deferredwork.py::test_append_entry_encode_failure_cannot_truncate_the_ledger` | Same ablation: `assert b'' == b'# Deferred ...pec-2-3.md)\n'` — the raise still happens, the *bytes* are what reddens. |
| `test_engine.py::test_defer_ledger_restore_write_failure_propagates_and_keeps_the_ledger` | `_defer` restore → bare `write_text`: `assert (False) … crashed=False` — the failure never reaches the run. |
| `test_sweep.py::test_migration_restore_write_failure_propagates_and_keeps_the_ledger` | `_ensure_migration` restore → bare `write_text`: `assert (False) … crashed=False`. Compares the restored ledger as **text**, not bytes — the fixture reaches disk via `write_text`, so a byte assertion would read CRLF and redden on Windows only. |

**#329 — surrogate neutralization.** Wiring and predicate are separate ablation axes and
were ablated separately.

| Test | Ablation evidence |
| --- | --- |
| `test_platform_util.py::test_neutralize_surrogates_replaces_only_lone_surrogates` (10 rows) | *Predicate* — helper → identity: 5 surrogate rows fail with `UnicodeEncodeError`; `clean`, `empty`, `astral-untouched`, `mixed-non-ascii`, `range-boundaries` stay green, which is what proves the range is not over-broad. |
| `test_platform_util.py::test_neutralize_surrogates_returns_clean_text_untouched` | Pins the `is`-identity fast path, so a clean ledger write stays byte-identical. |
| `test_platform_util.py::test_neutralize_surrogates_makes_atomic_write_text_survive_a_surrogate` | *Predicate* — helper → identity: `UnicodeEncodeError … surrogates not allowed`. |
| `test_deferredwork.py::test_append_entry_writes_a_finding_carrying_a_lone_surrogate` | *Wiring* — drop the `neutralize_surrogates` call from `_one_line`: `UnicodeEncodeError: 'utf-8' codec can't encode character '\ud800'`. No monkeypatching in the test — the value survives a real writer, a real strict encode and a real strict read back. |
| `test_deferredwork.py::test_append_decision_writes_a_label_and_detail_carrying_a_lone_surrogate` | Same ablation, same `UnicodeEncodeError`. |
| `test_deferredwork.py::test_mark_done_writes_a_note_carrying_a_lone_surrogate` | Same ablation, same `UnicodeEncodeError`. Uses the exact string `sweep._close_resolved` builds. |
| `test_deferredwork.py::test_append_entry_idempotence_survives_sanitizing` (3 rows) | Same ablation: the two surrogate rows redden, the `line-break` row stays green — sanitizing must stay ahead of the idempotence scan for both passes, not just the older one. |
| `test_sweep.py::test_write_intent_neutralizes_surrogates_and_keeps_the_markdown` | *Wiring* — drop `neutralize_surrogates` from `_write_intent`: `UnicodeEncodeError`. Also asserts `keep�this\n\nsecond para` — the paragraph break either side survives, so the document-wide pass is not collapsing markdown. |

No POSIX-only assertions were added. The `0600` mode diff is asserted nowhere in the
suite (mode semantics differ on Windows); it is documented above from a local probe and
in `atomic_write_text`'s docstring.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented failed deferred-work and bundle updates from truncating existing ledger data.
  * Improved recovery when ledger restoration fails, preserving the original data.
  * Sanitized unsupported characters in ledger entries, decisions, notes, and bundle intents while preserving formatting.
  * Newly created ledger files now use private permissions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -331,6 +331,15 @@ whose seams had diverged enough that several ports needed a different fix, and t
   creates from nothing now lands `0600` (the helper's private temp mode) rather than the umask
   default.
 
+- **A lone surrogate in triage text no longer crashes the sweep (#329).** A cached triage
+  `result.json` stores `\ud800` as an escape; the reload's `json.loads` revives the real code point
+  into an entry's evidence and on into the ledger note. It is not a line break, so the sanitizer
+  passed it through untouched, and it has no UTF-8 encoding at all — the strict encode raised
+  `UnicodeEncodeError` from a close path that calls the writers bare. Every free-text ledger field
+  is now neutralized at that same chokepoint, and the bundle `intent.md` gets the same pass over the
+  whole document (its line breaks are legitimate markdown, so only surrogates are touched). Each
+  becomes U+FFFD `�`, so the text stays visible rather than vanishing.
+
 - **A `#` inside a quoted sprint-status value is no longer rewritten into a comment (#366).** The
   board writer split value from inline comment with one fused pattern, so `3-2-x: "a # b"` advanced
   to `3-2-x: done # b"` — scalar text promoted into a comment the board never had. The split is now
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -322,6 +322,15 @@ whose seams had diverged enough that several ports needed a different fix, and t
 
 ### Fixed
 
+- **A failed ledger write can no longer empty the deferred-work ledger (#328).** `Path.write_text`
+  truncates the file and only then encodes, so any failure in that window — an unencodable value,
+  `ENOSPC`, `EIO` — left a zero-byte ledger with every entry gone. `append_decision`,
+  `append_entry`, the post-rollback and migration-failure restores, and the bundle intent write now
+  go through `atomic_write_text` like their `mark_done_many` sibling: the replacement is built
+  beside the target, so a failure raises with the original untouched. A ledger `append_entry`
+  creates from nothing now lands `0600` (the helper's private temp mode) rather than the umask
+  default.
+
 - **A `#` inside a quoted sprint-status value is no longer rewritten into a comment (#366).** The
   board writer split value from inline comment with one fused pattern, so `3-2-x: "a # b"` advanced
   to `3-2-x: done # b"` — scalar text promoted into a comment the board never had. The split is now
--- a/src/bmad_loop/deferredwork.py
+++ b/src/bmad_loop/deferredwork.py
@@ -24,7 +24,7 @@ from pathlib import Path
 
 from . import sprintstatus
 from .fences import fenced_spans
-from .platform_util import atomic_write_text
+from .platform_util import atomic_write_text, neutralize_surrogates
 
 HEADING_RE = re.compile(r"^### (DW-\d+): (.+?)\s*$", re.MULTILINE)
 ANY_HEADING_RE = re.compile(r"^#{1,6} ", re.MULTILINE)
@@ -618,11 +618,24 @@ def _one_line(value: str) -> str:
     line anyway — so this is the fix, and the skill docs are guidance that
     reduces occurrences without gating on them.
 
-    A value with no break is returned **untouched**, so an existing ledger is
-    never reformatted and a clean write is byte-identical to before the guard.
-    The trailing `.strip()` removes all surrounding whitespace, not merely the
-    space a leading or trailing break left behind — which is why it must stay on
-    the far side of that fast path.
+    That contract covers one hazard more than the break collapse alone, which is
+    what the `neutralize_surrogates` pass in front of it buys (#329). A lone
+    surrogate is not a line break, so it sailed through untouched — but it has
+    no UTF-8 encoding, and `atomic_write_text`'s strict encode raises
+    `UnicodeEncodeError` (a `ValueError` subclass) on it, from inside those same
+    bare close-path calls. It arrives the way the break did: a triage
+    `result.json` is cached with `json.dumps`, whose `ensure_ascii` keeps the
+    code point a harmless `\\ud800` escape, and the reload's `json.loads` revives
+    the real thing into `ResolvedEntry.evidence` and on into the `mark_done`
+    note. Refusing it upstream would only move the stoppage again — same
+    doctrine, same answer.
+
+    A value with neither a break nor a surrogate is returned **untouched**, so an
+    existing ledger is never reformatted and a clean write is byte-identical to
+    before the guard; each pass keeps its own fast path, so the common value is
+    scanned twice and copied never. The trailing `.strip()` removes all
+    surrounding whitespace, not merely the space a leading or trailing break left
+    behind — which is why it must stay on the far side of that fast path.
 
     A break-only value therefore sanitizes to `""`. Keeping it non-empty *here*
     could only yield bare whitespace, which trades an unfindable entry for an
@@ -634,7 +647,13 @@ def _one_line(value: str) -> str:
     empty detail rather than promising one that is not there. Its `label` needs
     neither: every member of :data:`LINE_BREAK_RE` is `str.isspace()`, and
     `validate_triage` builds each `DecisionOption` with `.strip() or key`, so a
-    break-only label has already become the option key before it arrives."""
+    break-only label has already become the option key before it arrives.
+
+    A surrogate-only value, by contrast, sanitizes to a truthy `"�"`, so neither
+    caller's empty-handling fires for it. That is the point of replacing rather
+    than stripping: a title reading `�` still says *something unencodable was
+    here*, where a vanished one would silently become `(untitled DW-<n>)`."""
+    value = neutralize_surrogates(value)
     if not LINE_BREAK_RE.search(value):
         return value
     return LINE_BREAK_RE.sub(" ", value).strip()
--- a/src/bmad_loop/deferredwork.py
+++ b/src/bmad_loop/deferredwork.py
@@ -875,7 +875,13 @@ def append_decision(path: Path, dw_id: str, date: str, label: str, detail: str) 
 
     Precondition: `date` is ISO `YYYY-MM-DD`; anything else raises `ValueError`,
     checked before the ``is_file`` short-circuit so an absent ledger cannot hide
-    the bug."""
+    the bug.
+
+    The write goes through :func:`~bmad_loop.platform_util.atomic_write_text` for
+    the reasons documented on :func:`mark_done_many`, plus one this sibling shares
+    with it: a bare ``Path.write_text`` truncates *before* it encodes, so any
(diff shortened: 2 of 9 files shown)


pbean:
CI note, so the red-then-green history on this PR is not mistaken for a fixed defect
here: the first run's `test (windows, py3.11)` failed on
`tests/test_tui_app.py::test_poll_skips_while_another_holds_the_lock` with
`textual.worker.WorkerCancelled` — the only failure in a `5258 passed` job.

That test is unrelated to this diff (this branch touches no TUI code; `git diff` over
`src/bmad_loop/tui/` and `tests/test_tui_app.py` is empty for the branch), and
`test (windows, py3.14)` passed it in the same run on the same commit. Re-running that
one job on the same commit passed. Diagnosis and remedy directions are filed as #581,
which this PR leaves open — the diff was not widened to reach it.

All 10 checks are green as of `2bce331`.

pbean:
@codex review

pbean:
@codex review

PR #582, opened 2026-08-13
fix(release,ci): promote `## [Unreleased]` at release time instead of authoring beside it

pbean:
## Problem

`CHANGELOG.md:3-6` declares Keep a Changelog 1.1.0 and `AGENTS.md` tells every dev session to file entries under `## [Unreleased]`. The release driver did the opposite: `.claude/skills/bmad-release/SKILL.md` step 3 told the curator to **author a brand-new `## [X.Y.Z]` section** from `release.py commits` (git log). Nothing ever promoted Unreleased into it.

`git grep -n "Unreleased"` returned exactly three tracked hits — `AGENTS.md:67`, the heading at `CHANGELOG.md:8`, and the compare link at `CHANGELOG.md:3086`. **Nothing in `scripts/release.py` mentioned `Unreleased` at all.** Consequences:

- `## [Unreleased]` grew to 1192 lines / 149 bullets, and `## [0.9.1]` duplicates much of it under different issue numbers.
- `ensure_link_ref` matched only *numeric* `[X.Y.Z]:` refs, so `[Unreleased]:` had no writer and its compare base went stale at every bump.
- `cmd_check`'s help string "local mirror of the CI release guards" was false — CI's `version-sync` job ran only `sync_version.py --check` and never looked at the CHANGELOG.

## What changed

**`prepare` refuses an unpromoted changelog.** Reuses `has_curated_section(text, "Unreleased")` — `section_re` already escapes its argument, so no new section regex. Paired with the existing "`## [X.Y.Z]` must be non-empty" precondition, the two together are what prove a *promotion* happened: content left Unreleased and arrived under the version heading.

**`ensure_link_ref` re-points `[Unreleased]:` at `compare/v<new>...HEAD`.** The version-ref insert and the Unreleased rewrite are now independent — the old shared early return meant a present `[X.Y.Z]:` ref would skip the rewrite, which is exactly what a re-run of `prepare` looks like. The rewrite is deliberately shape-blind so a hand-mangled line is repaired in place rather than duplicated by an insert.

**`check` holds the promote-and-reopen *result*, and CI runs it.** It calls `sync_version.check()` in-process rather than spawning `uv run python`, so it stays stdlib-only and the `version-sync` job can invoke it under `--no-project` without syncing the project. The job name is unchanged (branch protection); only the command moved.

**`AGENTS.md`** states the contract in tracked text. The skill that drives the release lives in gitignored `.claude/`, so this is what stops it drifting again.

### Design note: `check` does *not* mirror the emptiness guard

Canonical `__version__` is `0.9.0` and Unreleased currently holds 1192 lines, so a literal mirror would go red on this PR and stay red until the changelog is drained. It would also be semantically wrong: **between releases a populated `## [Unreleased]` is the correct state.** `check` therefore asserts the always-true results of promote-and-reopen — the heading was reopened, and its compare link tracks the current version — while emptiness stays a `prepare`-only precondition. Verified green on today's tree:

```
$ uv run --no-project python scripts/release.py check
version-sync:
ok: every version field agrees on 0.9.0 and the root mirror matches
changelog: `## [0.9.0]` section present
changelog: `## [Unreleased]` heading present
changelog: `[Unreleased]:` compares against v0.9.0
rc=0
```

## Ablation

Per the AGENTS.md testing rule, every new guard was deleted and the matching row confirmed to redden — six axes, each ablated **singly**, with `__pycache__` cleared between rows and the file restored from a `cp` backup (not `git checkout`). The harness aborted if an ablation patch matched nothing, so no row could pass as a fake green.

| Ablated | Reddens |
| --- | --- |
| `prepare` Unreleased-emptiness guard | `test_prepare_refuses_a_still_populated_unreleased` |
| `ensure_link_ref` Unreleased rewrite | the three `..._repoints_/_inserts_unreleased_...` rows |
| `check` heading probe | `test_check_flags_a_consumed_unreleased_heading` |
| `check` missing-compare-ref arm | `test_check_flags_a_missing_unreleased_compare_ref` |
| `check` stale-compare-base arm | `test_check_flags_a_stale_unreleased_compare_base` |
| shape-blind rewrite → compare-shaped | `test_ensure_link_ref_repairs_a_malformed_unreleased_ref_in_place` |

Each `check` fixture derives from a promoted baseline by breaking exactly one thing, so a shared `rc == 1` cannot pass for the wrong reason; the rows assert the specific message. `test_prepare_accepts_a_promoted_changelog` is the positive control proving the guard is not simply always-failing.

## Not in this PR

- **The `bmad-release` skill edit is invisible in this diff.** `.claude/` is gitignored (`.gitignore:8`), and `bmad-release` is dev-only with no canonical copy under `src/bmad_loop/data/skills/`, so editing it in place is correct and `seed_skills.py` was not run. Its step 3 now says: rename `## [Unreleased]` to `## [X.Y.Z] — <ISO date>` in place, reopen an empty `## [Unreleased]` above it, and use `release.py commits` only as a *gap cross-check* after curating.
- **`CHANGELOG.md` content is untouched** — draining Unreleased is a follow-up. No edit was needed for CI to pass: canonical is `0.9.0` and the compare link already reads `compare/v0.9.0...HEAD`.
- **No linter for the six subsections.** History contains `### Docs` (2×) and `### Migration` (1×); enforcement would fail on it. AGENTS.md states the forward contract only.
## Also in this PR (requested)

The lint job's checkout gains `persist-credentials: false` — it was the only one of six in the file leaving the credential helper configured, and it does so with `fetch-depth: 0` (`zizmor/artipacked`). trunk-action only reads local history, so the token is not needed. Pre-existing on `main`; included here at the maintainer's request rather than split out.

## Verification

- `uv run pytest -q -n auto` → 5403 passed, 44 skipped, 5 xfailed
- `uv run pyright` → 0 errors
- `trunk check --no-fix` (no path filter) → No issues
- `uv run --no-project python scripts/release.py check` → rc 0

## Summary by CodeRabbit

* **Release Management**
  * Strengthened release validation for version consistency and changelog structure.
  * Release preparation now promotes unreleased changes into dated version sections and refreshes comparison links automatically.
  * Invalid or incomplete changelog states are rejected before release.

* **Documentation**
  * Updated release and testing guidance to describe the enhanced validation workflow.

* **Tests**
  * Added coverage for changelog promotion, comparison links, version mismatches, missing sections, and invalid release states.


pbean:
@codex review

pbean:
## Review round 1

**Codex P1 — reopened `## [Unreleased]` heading: ACCEPTED, fixed in `e873e35`.**

Confirmed on both legs. `has_curated_section` reports `False` identically for a correctly emptied section and for one that was renamed and never reopened, so the guard could not tell them apart. And `release.yml` fires on `push: [main, release/*]` with **no dependency on ci.yml** — it can create the tag and GitHub release while `version-sync` is still running, so `check` catching the missing heading afterwards is too late to prevent the irreversible step. `prepare` is the last gate that runs before it.

Split the two failures apart using `extract_section`'s `None`, and took the ordering half too — a reopened Unreleased must sit above the section it was promoted into. Both are ablated: collapsing missing-into-empty reddens `test_prepare_refuses_a_never_reopened_unreleased`, and disabling the ordering compare reddens `test_prepare_refuses_an_unreleased_reopened_below_the_release`.

**CodeRabbit — compare-link repository validation: DECLINED.**

The observation is accurate (`\S+` accepts any URL prefix), but comparing against `repo_url()` costs more than it buys:

1. **It reintroduces a git subprocess into a check that is deliberately subprocess-free.** `cmd_check` calls `sync_version.check()` in-process precisely so CI can run it under `--no-project` without syncing the project. `repo_url()` shells out to `git remote get-url origin`.
2. **It false-reds a forked local clone.** A contributor whose `origin` is their own fork would fail `release.py check` against a CHANGELOG that correctly names the upstream.
3. **This file's own data contradicts the invariant.** Link refs at `≤0.7.12` legitimately name `bmad-code-org/bmad-auto` and `≥0.8.0` name `bmad-loop`, from the historical rename — "the ref's repo equals today's origin" is not a property this CHANGELOG holds.

The write side already guarantees correctness: `prepare` always emits `repo_url()` through `ensure_link_ref`. What `check` exists to catch is a **stale base**, which it does. The residual threat is a hand-edit pointing at another org — not a failure mode with a path into this repo, and self-correcting on the next `prepare`.

Happy to revisit if either of you sees a way to bind the repo without the subprocess and the fork false-positive.

@codex review

pbean:
## Review round 2

**Codex P2 — leftover second `## [Unreleased]`: ACCEPTED, fixed in the pushed head.**

Confirmed. `extract_section` is built on `search`, so both new guards answered for the first heading only. Insert a fresh empty `## [Unreleased]` above the release but forget to rename the original, and the emptiness check reads the empty one, the ordering check compares that same match, and `prepare` succeeds while the real entries sit below — unshipped, and duplicated into a later release. That is precisely the Unreleased-vs-versioned-section drift this PR exists to close, so it landing through my own guard is worth having caught.

Fixed by scanning `section_re("Unreleased").finditer(...)` rather than searching, and refusing anything other than exactly one heading. Uniqueness stays prepare-only for the same reason emptiness does: it describes the *promotion*, not the state a released CHANGELOG sits in between releases.

Ablated: replacing the `len(unreleased) > 1` arm reddens `test_prepare_refuses_a_leftover_second_unreleased` and nothing else.

Gates: 5406 passed / 44 skipped / 5 xfailed, pyright 0 errors, `trunk check --no-fix` (no path filter) clean, `release.py check` rc 0.

@codex review

pbean:
## Review round 3

**Codex P2 — promoted heading without an ISO date: ACCEPTED, fixed in `0053872`.**

Correct on the mechanism: `section_re` matches `## [X.Y.Z]` followed by `[^\n]*`, so a heading renamed without its date — or with arbitrary trailing text — reads as a clean promotion to every other guard, and `prepare` commits.

Taking it because it is this PR's own thesis rather than adjacent scope: the root cause here was a **documented-but-unenforced** contract, and I had just written `## [X.Y.Z] — <ISO date>` into AGENTS.md as tracked text. Leaving that half unenforced would reproduce the exact failure the PR exists to close.

Checked the enforcement against real data before adding it — every versioned heading in CHANGELOG.md is `## [X.Y.Z] — YYYY-MM-DD`, the sole exception being the pre-dating `## [0.1.0]` at line 3080, which no release can re-prepare (`version_gt` rejects it). So the gate matches the repo's own history rather than inventing a rule it violates.

Scoped deliberately:
- **`prepare` only.** `check` reads the *canonical* version's section and describes the post-release state; the date is a promotion-time property, consistent with where the emptiness and uniqueness guards live.
- **Conditioned on the heading existing**, so a missing section is still reported once rather than twice for one cause.

Ablated: neutralizing the condition reddens `test_prepare_refuses_a_release_heading_without_a_date` and nothing else.

Gates: 5407 passed / 44 skipped / 5 xfailed, pyright 0 errors, `trunk check --no-fix` (no path filter) clean, `release.py check` rc 0.

@codex review

pbean:
## Review round 4

**Codex P2 — impossible ISO dates: ACCEPTED, fixed in `5ec2f59`.**

Correct: `RELEASE_HEADING_RE` pins digit widths only, so `2026-02-31`, `2026-99-99` and `2026-13-05` all shape correctly and the round-3 guard took them.

Taken because the fix is a *subtraction*, not another layer: instead of growing the regex toward a calendar, the captured value now goes to `date.fromisoformat`, which is the thing that actually decides whether a string is an ISO date. Still stdlib-only, so `check` remains runnable under `--no-project`.

Ablated: dropping the `fromisoformat` call while keeping the digit regex reddens all three `test_prepare_refuses_an_impossible_release_date` rows and leaves the round-3 undated row green — the two axes are disjoint, which is what shows the calendar check is doing its own work rather than riding on the shape check.

Gates: 5410 passed / 44 skipped / 5 xfailed, pyright 0 errors, `trunk check --no-fix` (no path filter) clean, `release.py check` rc 0.

---

**On convergence.** Four rounds, four accepted findings, and I want to name the trajectory rather than keep grinding: rounds 1–2 were real data-loss defects (entries silently lost before an irreversible publish; entries stranded behind a duplicate heading). Rounds 3–4 have been successive refinements of the heading-format check, at cosmetic severity.

That is a converging severity curve, and this round closes the date axis completely — `fromisoformat` admits exactly the valid calendar days, so there is no stricter version of this check to ask for next.

If round 5 raises another refinement on the same axis (a future-dated release, a date disagreeing with the commit, ordering between versioned headings), I'm going to hand it to the maintainer as a follow-up issue rather than widen `prepare` further in this PR. Genuine defects elsewhere in the diff are of course still welcome.

@codex review

pbean:
**Correction to my previous comment — disregard the deferral policy.**

The maintainer has directed that findings be fixed here rather than split into follow-up issues. So strike the last section of my round-4 comment: there is no "round 5 goes to an issue" line. Every valid finding gets fixed in this PR until the gate is clean, refinements on the heading axis included.

The severity observation still stands as context, but it is not a reason to defer anything.

pbean:
## CodeRabbit finding — decline REVERSED, fixed in `2eee7e1`

Two things changed my position.

**1. The green was not a green.** The CodeRabbit check reports `pass`, but its comment carries `<!-- rate limited by coderabbit.ai -->` — *"You've used all free OSS reviews for now"* — and the verdict it edited in place still reads:

> The release check can accept an Unreleased compare link for a different repository… **merge should wait for canonical-repository validation or explicit owner acceptance.**

So the finding was never withdrawn; a passing check simply meant no review ran.

**2. Two of my three objections do not survive checking.**

- *"It puts a subprocess into a subprocess-free check."* `--no-project` governs Python **project syncing**, not subprocesses. Reading `origin` installs nothing and does not weaken that flag. The claim was wrong as stated.
- *"It false-reds a fork PR in CI."* `actions/checkout` sets `origin` to the **base** repo even for a pull request from a fork, so CI cannot false-red. The residue is a local clone whose `origin` is a fork — and there the arm names the fork, which is *accurate*, not spurious.
- *"The CHANGELOG's own history breaks the invariant."* This one was never load-bearing: the historical `bmad-auto` refs are untouched, since only the `[Unreleased]` line is bound.

Implemented as proposed: capture the repo in `UNRELEASED_COMPARE_RE` and compare it with `repo_url()`, plus the test that was asked for — canonical version, different repository, `cmd_check` fails.

Ablated: neutralizing the repo comparison reddens `test_check_flags_an_unreleased_compare_link_to_another_repo` and leaves `test_check_flags_a_stale_unreleased_compare_base` green. The two arms are disjoint, so the repo check is doing its own work rather than riding on the base check.

Verified against the live tree — `origin` and the CHANGELOG's `[Unreleased]` link agree, so `release.py check` is still rc 0 on `main`'s content.

Gates: 5411 passed / 44 skipped / 5 xfailed, pyright 0 errors, `trunk check --no-fix` (no path filter) clean, `release.py check` rc 0.

Every finding from both bots is now fixed; nothing is deferred and nothing is declined.

@codex review

PR #585, opened 2026-08-14
docs: name the obligations CI enforces, fix accuracy defects

pbean:
## What

Adds the contributor obligations CI actually enforces but the docs never stated, and fixes a set of accuracy defects found in an audit.

## Why

A contributor following CONTRIBUTING's verify step — "`trunk check` and `uv run pytest -q` both pass" — pushes code that fails CI, which runs a dedicated `pyright` job. CONTRIBUTING had zero mentions of `pyright` and zero of the CHANGELOG, whose `[Unreleased]` contract AGENTS.md requires.

## How

- **CONTRIBUTING**: adds `uv run pyright` to dev setup and the verify step; a CHANGELOG section pointing at the `[Unreleased]` contract; Windows `PYTHONUTF8=1` and the Python 3.11 floor; corrects the extras gloss (the extras are `tui`, `non-linux`, `opencode` — it claimed only `tui`); names `scripts/release.py` and its two-phase `prepare`/`publish` flow.
- **PR template**: adds a Changelog section.
- **README**: adds `bmad-loop adapters` to the command table; links `docs/README.md` as the full index plus the five guides the list omitted; documents the `[dev]` policy section.
- **FEATURES**: lists all 15 policy sections (it showed 12, omitting `[stories]`, `[dev]`, `[operator]`, `[mux]`) and documents `[dev] skill`.
- **ROADMAP**: the native-Windows entry was `planned` — its own legend's "agreed, not started" — while the body said psmux had shipped in-tree. Restated as `in progress`, with the legend extended to define it.
- **AGENTS**: leaf-module count `~25` → 28 (42 top-level `.py` − 12 named in the table − 2 dunders).

`bmad-loop relay` is deliberately **not** added to the command table: its help calls it "a hook target for machines, not a command to run by hand," and its docstring says nothing points at it yet — `init` still registers the copied workspace relay. A note under the table states the omission and why.

## Testing

`trunk check --all` clean across 254 files; suite 5411 passed / 44 skipped / 5 xfailed; `uv run pyright` 0 errors. Every relative link and anchor added was verified to resolve.

## Changelog

Entry added under `## [Unreleased]` → `Fixed`.

## Summary by CodeRabbit

- **Documentation**
  - Added guidance for the `bmad-loop adapters` command, adapter profiles, load warnings, and the internal `relay` hook.
  - Documented development-skill configuration, supported policy sections, dependency setup, validation, Windows configuration, and release procedures.
[...]

PR #586, opened 2026-08-14
fix(sync-version): refuse a marketplace.json with no usable plugins list, and correct four doc claims

pbean:
Merged `main` in (PR #585 landed while this was open) and folded in the round-3 review fixes that missed that merge. Two independent changes, kept in separate commits.

## 1. `sync_version.check()` could pass without checking anything

The marketplace loop was `for i, plugin in enumerate(market.get("plugins", []))`. A renamed, deleted or emptied `plugins` key makes that iterate zero times, so the `version-sync` gate goes green on exactly the corruption it exists to catch. AGENTS.md names `sync_version.py` the sole version write path, and this job is what holds it before a release.

Measured by ablation against the pre-change code:

| marketplace.json | before | after |
| --- | --- | --- |
| valid, agreeing | 0 | 0 |
| valid, stale version (positive control) | 1 | 1 |
| `plugins` renamed to `plugin` (+ stale version behind it) | **0** | 1 |
| `plugins` key deleted | **0** | 1 |
| `plugins: []` | **0** | 1 |
| `plugins: {}` | **0** | 1 |
| `plugins: null` | RAISED TypeError | 1 |
| `plugins: ["str"]` | RAISED AttributeError | 1 |

The four `0` rows are silent vacuous passes — the renamed-key row had a genuinely stale version behind it and still went green. The two raises were fail-closed already, but surfaced as a crashed job rather than a drift report.

New `tests/test_sync_version.py` (there was none; `test_release.py` stubs `check` out). Every negative case is paired with a positive control, and I removed the guard to confirm the five vacuous-pass rows go **red** before trusting them.

## 2. Four documentation claims that were false

From the round-3 automated review on #585, validated before acting:

- **`trunk check` scope.** AGENTS.md and CONTRIBUTING both said "full lint, no path filter". Bare `trunk check` lints **7 of 254** files here (changed-vs-upstream); `--all` lints 254. CI's `trunk-action` defaults to changed-files too — ci.yml's own checkout comment says `fetch-depth: 0` is "for diff-aware checks". Both files now name the real scope and point at `--all`.
- **`release.py check` enforcement.** CONTRIBUTING claimed it enforces the no-hand-authored-version-section rule. It does not — injecting a populated `## [0.9.1]` above a preserved `## [Unreleased]` returns **rc=0**. Scoped to what it actually holds, matching AGENTS.md's wording.
- **`prepare` does not promote.** Nothing told a maintainer the rename-and-reopen is a manual pre-step; `release.py:382-435` only refuses until it is already done.
- **The verify step** listed 3 of 6 CI jobs as if complete. Qualified with a pointer to docs/testing.md rather than enumerating jobs — Windows and the 3.11–3.14 matrix aren't locally reproducible anyway.

One review finding is **declined**: CodeRabbit asked to move the "Uninstall procedure" entry from `[0.5.1]` to `[Unreleased]`. That entry is not new — it exists on `main` at CHANGELOG.md:2742 under `[0.5.1]` → `### Docs`, and #585 only moved it into that same released version's `### Added`. Following the advice would republish a v0.5.1 change (shipped 2026-06-20) as new in the next release, and `release.py check` returns 0 either way, so no gate would catch the misattribution.

## Testing

Suite 5420 passed / 44 skipped / 5 xfailed (9 new), `uv run pyright` 0 errors, `trunk check --all` clean across 255 files, `release.py check` green. Merge verified entry-preserving: 374 pre-#585 changelog entries + #585's 2 + 1 new = 377, none lost. All relative links and anchors resolve.

## Changelog

Two entries under `## [Unreleased]` → `Fixed`.


pbean:
@codex review

PR #587, opened 2026-08-14
fix: validate limits policy scalar types

### LINKED ISSUE #278: Strict scalar-type validation for limits.* policy fields
### Context

Surfaced during the adversarial review of #277 (finding #6, deferred there as a repo-wide consistency issue rather than a one-off).

`policy.py` loads every `limits.*` scalar with a bare coercion — `int(...)`, `float(...)`, `str(...)`, `bool(...)` — then range/membership-validates a few (e.g. `session_budget_mode`). There is **no** strict per-field *type* validation. `bool(...)` in particular silently coerces a truthy string:

```toml
# policy.py:720
dev_contract_nudge = bool(limits_d.get("dev_contract_nudge", LimitsPolicy.dev_contract_nudge))
```

so `dev_contract_nudge = "false"` (a quoted string — a plausible user error) coerces to `True` and **enables** the nudge. A well-formed TOML boolean (`dev_contract_nudge = false`) already parses correctly; only the quoted form misbehaves.

### Why not fix it narrowly

Fixing only `dev_contract_nudge` would be an inconsistent island: every sibling `limits.*` field uses the same coerce-then-value-validate convention. The right fix is a small cross-cutting sweep.

### Proposal

Add strict scalar-type validation across `limits.*` fields (bool/int/float/enum) — reject or warn when a value's TOML type doesn't match the field's declared type — following the existing strict precedent in the plugin-settings path (`_validate_plugin_settings`, `policy.py:613`), which already raises `PolicyError` on a non-`bool` where a bool is expected. Add invalid-type tests per kind.

### Scope

- `src/bmad_loop/policy.py` (the `limits` load path)
- `tests/test_policy.py`

Low priority / hardening.

shin4141:
## What

Reject TOML scalar-type mismatches across every `limits.*` policy field before constructing `LimitsPolicy`.

## Why

Bare `int()`, `float()`, `str()`, and `bool()` coercions could silently change policy meaning. In particular, `dev_contract_nudge = "false"` evaluated to true.

Fixes #278

## How

- Route integer, number, boolean, and string limits through strict type validators.
- Preserve the existing range and option checks after type validation.
- Cover every integer field with TOML boolean, float, and quoted-integer cases, plus the number, boolean, and string validators.
- Record single-gate ablation evidence for all four scalar validators.

## Testing

Local candidate `66faf2ca44973e4d03854a0f0d0d5484072e5d8b`:

- `uv sync --all-extras --locked`
- `uv run pytest tests/test_policy.py tests/test_settings_schema.py -q` — 193 passed
- `uv run pytest -q -n auto --durations=15`
- `uv run pyright`
- `trunk check --all`
- `uv run python scripts/release.py check`
- `git diff --check`

[Exact-head CI run 32089380448](https://github.com/bmad-code-org/bmad-loop/actions/runs/32089380448) passed on the same SHA:

- Ubuntu Python 3.11, 3.12, 3.13, and 3.14
- Native Windows Python 3.11 and 3.14
- Pyright, trunk lint, version-sync, and packaging
- Both Windows jobs used `PYTHONUTF8=1` and completed the full deterministic suite with 5,799 passed, 200 skipped, and zero failures
- CodeRabbit completed successfully with no new actionable comments

No macOS CI result is claimed.

## Changelog

Added a Fixed entry under Unreleased.

AI assistance was used; I reviewed the resulting diff and tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Policy limit settings now reject mismatched TOML value types instead of silently converting them.
  * Invalid integer, numeric, boolean, and string values consistently report clear policy errors.
  * Existing defaults and range validation remain unchanged.

* **Documentation**
  * Added changelog coverage for stricter policy type validation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@ breaking changes may land in a minor release.
 
 ### Fixed
 
+- **Reject mismatched TOML scalar types for every `limits.*` policy field (#278).**
+  Quoted booleans and other coercible values now raise `PolicyError` instead of silently changing
+  the configured limit or enabling a disabled behavior.
+
 - **Attempt-owned spec-only retries no longer demand a false manual rollback (#123).** A
   bound plain attempt whose only residue is its lifecycle flip is restored to its pre-attempt
   lifecycle status and proven Git-clean before retry. A resolved re-drive may instead retain an
--- a/src/bmad_loop/policy.py
+++ b/src/bmad_loop/policy.py
@@ -666,6 +666,34 @@ def _validate_plugin_settings(name: str, raw: dict[str, Any], specs: Any) -> Non
             )
 
 
+def _limit_int(raw: dict[str, Any], key: str, default: int) -> int:
+    value = raw.get(key, default)
+    if isinstance(value, bool) or not isinstance(value, int):
+        raise PolicyError(f"limits.{key} must be an integer: got {value!r}")
+    return value
+
+
+def _limit_float(raw: dict[str, Any], key: str, default: float) -> float:
+    value = raw.get(key, default)
+    if isinstance(value, bool) or not isinstance(value, (int, float)):
+        raise PolicyError(f"limits.{key} must be a number: got {value!r}")
+    return float(value)
+
+
+def _limit_bool(raw: dict[str, Any], key: str, default: bool) -> bool:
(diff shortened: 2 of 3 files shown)


pbean:
@codex review

PR #598, opened 2026-08-14
ci: speed up the test jobs (temp on D:, logical workers, slim sandbox template)

pbean:
## Baseline

Measured on 2026-08-14 against live CI logs:

| | mean | max | timeout |
| --- | --- | --- | --- |
| `test-windows` | **699 s** | **888 s** | 900 s |
| `test` (Linux) | **215 s** | — | 900 s |

The "Run tests" step is 95%+ of the Windows job, and the overhead is spread evenly across the suite (~0.635 s/test on Windows vs ~0.176 s/test on Linux) rather than concentrated in a few slow tests — so this is per-test constant cost, not a handful of bad tests.

## Levers

Each is its own commit.

| # | Lever | Expected effect |
| --- | --- | --- |
| 1 | **Pin `TMP`/`TEMP` to `runner.temp`** on the Windows "Run tests" step | pytest roots `tmp_path` at `tempfile.gettempdir()`, which follows `TMP`/`TEMP` on Windows. The user-profile temp has historically sat on the slow C: volume while the workspace and `RUNNER_TEMP` sit on D:, so every per-test sandbox copytree crossed to the slower disk. pip measured −28% from the same redirect ([pypa/pip#13129](https://github.com/pypa/pip/issues/13129)). No-op if the default temp is already on D:. |
| 2 | **`-n auto` → `-n logical`** in *both* test jobs | `psutil` is in the `non-linux` extra with no environment marker (pyproject.toml:34) and CI installs `--all-extras`, so xdist's `auto` resolves to `psutil.cpu_count(logical=False)` — the *physical* count, ~2 on the 4-vCPU hosted runners. `logical` asks for the vCPU count and self-adapts if runner shape changes. Applies to Linux too: same situation there. |
| 3 | **Drop git's 14 dead `*.sample` hooks** from the session-scoped sandbox template | ~1,500 tests each copytree the template; all 14 sample files were replicated per test and nothing reads them. |
| 4 | **Windows `timeout-minutes: 15 → 20`** | The max observed run was 888 s against a 900 s cap — one slow runner away from a timeout that reads as a test failure. Headroom for the tail while the levers above pull the mean down. Windows only. |

### Notes on what was deliberately *not* done

- The `git init --template=` shortcut is wrong here: an empty template also removes `.git/hooks/` and `.git/info/exclude`, and tests depend on both — three sites write `.git/hooks/pre-commit` with no `mkdir` (tests/test_engine.py:940, tests/test_engine.py:3002, tests/test_verify.py:2600) and tests/test_install.py reads `.git/info/exclude`. Only the `*.sample` files are deleted.
- Lever 1 is set at **step** level, not job level. The `runner` context is unavailable in a job `env:` block — actionlint rejects it, and the expression would expand to empty, which Python's `tempfile` then skips. Job-level would have been silently inert.
- The stale 2025 advice circulating for this problem (remove the D: drive, Defender exclusion tricks) does not apply: today's image is windows-2025-vs2026, the workspace / `RUNNER_TEMP` / `UV_CACHE_DIR` already live on `D:\a`, and Defender is already disabled on hosted runners.

## Diagnostics commit — will be stripped

The last commit adds a temporary `Diagnostics (STRIP BEFORE MERGE)` step to both test jobs, printing the runner's default tempdir, `RUNNER_TEMP`, and psutil's physical/logical CPU counts. It carries no `TMP`/`TEMP` override on purpose, so it shows both where the temp defaulted *and* where the pin moves it — with the override it could only echo the pinned value back.

**This commit will be reverted once the first runs confirm the tempdir location and the worker counts.** The measurements land in a PR comment below.

## Local verification

- `uv run pytest -q -n logical` — 5461 passed, 44 skipped, 5 xfailed
- `uv run pytest -q tests/test_engine.py tests/test_verify.py tests/test_install.py` — 970 passed, 23 skipped (the tests that touch `.git/hooks` and `.git/info/exclude`)
- `trunk check` — clean (actionlint + zizmor cover the workflow edit)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test reliability across Linux and Windows environments.
  * Ensured temporary test files remain in the appropriate workspace location on Windows.
  * Improved handling of platform-specific user directories and repository test fixtures.

* **Documentation**
  * Updated testing guidance to recommend predictable logical worker allocation for parallel test runs.
  * Clarified hosted-runner behavior and development test commands.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
         run: uv sync --locked --all-extras
 
       - name: Run tests
-        run: uv run pytest -q -n auto --durations=15
+        run: uv run pytest -q -n logical --durations=15
 
   test-windows:
     # Real-Windows runner for the deterministic suite (live psmux e2e stays a manual gate).
@@ -92,7 +92,7 @@ jobs:
         env:
           TMP: ${{ runner.temp }}
           TEMP: ${{ runner.temp }}
-        run: uv run pytest -q -n auto --durations=15
+        run: uv run pytest -q -n logical --durations=15
 
   version-sync:
     name: version-sync
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,6 +82,16 @@ jobs:
         run: uv sync --locked --all-extras
 
       - name: Run tests
+        # pytest roots `tmp_path` at `tempfile.gettempdir()`, which follows
+        # TMP/TEMP on Windows. The user-profile temp has historically sat on the
+        # slow C: volume while the workspace and RUNNER_TEMP sit on D:, so every
+        # per-test sandbox copy crossed to the slower disk; pin both to the
+        # workspace volume. A no-op if the default temp is already there.
+        # Step-level, not job-level: the `runner` context is unavailable in a
+        # job `env:` block, where this would silently expand to empty.
+        env:
+          TMP: ${{ runner.temp }}
+          TEMP: ${{ runner.temp }}
         run: uv run pytest -q -n auto --durations=15
 
   version-sync:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,10 +55,11 @@ jobs:
     # Real-Windows runner for the deterministic suite (live psmux e2e stays a manual gate).
     name: test (windows, py${{ matrix.python-version }})
     runs-on: windows-latest
-    # 20, not the 15 the Linux test job uses: the slowest observed Windows run
+    # 20, not the 15 the Linux test job uses: the slowest pre-lever Windows run
     # took 888s against a 900s cap, so one slow runner was a bad minute away from
-    # a timeout that reads as a test failure. The levers above should pull the
-    # mean well down; this keeps the tail from failing the build first.
+    # a timeout that reads as a test failure. The levers below cut the measured
+    # mean to ~316s, but the leg still spreads 220-348s run to run; the extra
+    # headroom keeps that tail from failing the build first.
     timeout-minutes: 20
     # The suite reads/writes UTF-8 files; Windows' cp1252 default would break plain
     # text I/O. UTF-8 mode (PEP 686's Python 3.15 default) makes the runner match the
@@ -87,10 +88,10 @@ jobs:
 
       - name: Run tests
         # pytest roots `tmp_path` at `tempfile.gettempdir()`, which follows
-        # TMP/TEMP on Windows. The user-profile temp has historically sat on the
-        # slow C: volume while the workspace and RUNNER_TEMP sit on D:, so every
-        # per-test sandbox copy crossed to the slower disk; pin both to the
-        # workspace volume. A no-op if the default temp is already there.
+        # TMP/TEMP on Windows. Measured on the windows-2025 image (2026-08-14):
+        # the default is C:\Users\RUNNER~1\AppData\Local\Temp while the workspace
+        # and RUNNER_TEMP sit on D:\a — so every per-test sandbox copy crossed to
+        # the slower volume. Pinning both to RUNNER_TEMP keeps them on D:.
         # Step-level, not job-level: the `runner` context is unavailable in a
         # job `env:` block, where this would silently expand to empty.
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,11 @@ jobs:
     # Real-Windows runner for the deterministic suite (live psmux e2e stays a manual gate).
     name: test (windows, py${{ matrix.python-version }})
     runs-on: windows-latest
-    timeout-minutes: 15
+    # 20, not the 15 the Linux test job uses: the slowest observed Windows run
+    # took 888s against a 900s cap, so one slow runner was a bad minute away from
+    # a timeout that reads as a test failure. The levers above should pull the
+    # mean well down; this keeps the tail from failing the build first.
+    timeout-minutes: 20
     # The suite reads/writes UTF-8 files; Windows' cp1252 default would break plain
     # text I/O. UTF-8 mode (PEP 686's Python 3.15 default) makes the runner match the
     # files under test; the conftest guard enforces the same for local win32 runs.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -6,7 +6,7 @@ bmad-loop is a deterministic Python orchestrator that drives unattended BMAD-met
 
 ```bash
 uv sync --all-extras          # setup (extras: tui, non-linux, opencode)
-uv run pytest -q              # test suite (-n auto to parallelize)
+uv run pytest -q              # test suite (-n logical to parallelize)
 uv run pytest tests/test_engine.py -q   # single file
 uv run pyright                # typecheck — same pinned version CI runs
 trunk fmt                     # format changed files
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -19,10 +19,13 @@ and every negative assertion proven by ablation.**
 - Configuration is two settings under `[tool.pytest.ini_options]` in `pyproject.toml`:
   `testpaths = ["tests"]` and `asyncio_mode = "auto"`. There is no `addopts`, no marker
   registry, no `filterwarnings` (the last is under triage — #548).
-- `uv run pytest -q` runs everything runnable on the host; `-n auto` (pytest-xdist) is
+- `uv run pytest -q` runs everything runnable on the host; `-n logical` (pytest-xdist) is
   supported because every test builds its state from per-test fixtures (the template repo is
-  per-worker). A test that fails only under xdist load is a flake, which is a bug — #360 is
-  the open instance. Live/E2E modules skip themselves when their host requirements are
+  per-worker). Prefer `logical` over `auto`: `psutil` is installed on every platform here
+  (the `non-linux` extra carries no environment marker, and CI syncs `--all-extras`), so
+  xdist's `auto` resolves to the _physical_ core count — half the vCPUs on the hosted
+  runners. CI passes `-n logical` in both test jobs for that reason. A test that fails only
+  under xdist load is a flake, which is a bug — #360 is the open instance. Live/E2E modules skip themselves when their host requirements are
   absent — selection is in-file, never in config.
 - Only builtin pytest marks appear: `parametrize`, `skipif`, `usefixtures`, and exactly one
   `xfail(strict=True)` (a pinned known defect, see [Ablation records](#ablation-records)).
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -326,6 +326,13 @@ def _project_template(
     (root / "src.txt").write_text("original\n")
     (root / ".gitignore").write_text(".bmad-loop/runs/\n")  # as `bmad-loop init` would
     git(root, "init", "-q", "-b", "main")
+    # `git init` seeds 14 dead `*.sample` hooks nothing here reads, and every test
+    # replicated all 14 through `project`'s copytree. Drop the files only — NOT via
+    # `git init --template=` (an empty template also drops `.git/hooks/` itself and
+    # `.git/info/exclude`, which tests do depend on: three sites write
+    # `.git/hooks/pre-commit` with no mkdir, and the install tests read info/exclude).
+    for sample in (root / ".git" / "hooks").glob("*.sample"):
+        sample.unlink()
     # Local config: copies (and their worktrees) inherit it via the copied .git/config.
     git(root, "config", "user.email", "test@test")
     git(root, "config", "user.name", "test")
--- a/tests/test_conftest.py
+++ b/tests/test_conftest.py
@@ -1,0 +1,33 @@
+"""Contract tests for the sandbox fixtures in `tests/conftest.py`.
+
+`project` hands every test a copytree clone of a session-scoped template repo, so
+the template's shape is a shared dependency of most of the suite. What is pinned
+here is the part of that shape other modules rely on without asserting it.
+"""
+
+from __future__ import annotations
+
+
+def test_template_drops_sample_hooks_but_keeps_hooks_dir_and_exclude(project):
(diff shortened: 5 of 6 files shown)


pbean:
## First run: both hypotheses confirmed, and one real test bug surfaced

Run `31849660788` (head `5ca730a`). Ancillary jobs (`build`, `lint`, `typecheck`, `version-sync`) all green.

### Diagnostic output

Identical on both Windows jobs and both Python versions:

```
default tempdir: C:\Users\RUNNER~1\AppData\Local\Temp
RUNNER_TEMP:     D:\a\_temp
physical: 2
logical:  4
```

Linux:

```
default tempdir: /tmp
RUNNER_TEMP:     /home/runner/work/_temp
physical: 2
logical:  4
```

Both premises hold:

- **Temp really was on C:.** pytest's `tmp_path` defaulted to the user-profile temp on the slow C: volume while the workspace sits on `D:\a` — exactly the gap lever 1 targets. (Linux `/tmp` is fine as-is; lever 1 is Windows-only for that reason.)
- **`auto` really was halving the pool.** `physical: 2` vs `logical: 4` on *both* OSes, so `-n auto` was running 2 workers on a 4-vCPU runner. Confirms applying lever 2 to Linux as well, not just Windows.

### "Run tests" step duration vs baseline

| job | before | after | Δ |
| --- | --- | --- | --- |
| `test-windows` (job wall-clock, mean) | **699 s** | **352 s** | **−50%** |
| `test-windows` (job wall-clock, max) | **888 s** | **362 s** | **−59%** |
| `test` Linux (job wall-clock, mean) | **215 s** | **164 s** | **−24%** |

Per-job "Run tests" step: windows py3.11 321 s, py3.14 306 s; linux py3.11 143 s, py3.12 157 s, py3.13 142 s, py3.14 132 s.

All three levers landed at once, so the split between them isn't separable from this run. But Linux — which got only levers 2 and 3 — moved −24%, while Windows, which also got lever 1, moved −50%. That gap is consistent with the temp redirect carrying roughly half of the Windows win.

The Windows max is now 362 s against the old 900 s cap, so the lever-4 bump to 20 min is headroom rather than necessity. Worth keeping anyway — the tail is what the cap has to survive, not the mean.

### One failure, and it is a real bug — not a flake

`tests/test_probe.py::test_discover_location_redacts_username` failed on **both** Windows jobs (py3.11 on `gw3`, py3.14 on `gw1`) with an identical assertion. Same test, both versions, different workers: deterministic. Linux is green on all four versions.

**Mechanism.** The test does:

```python
monkeypatch.setenv("HOME", str(tmp_path))
...
assert found.location.startswith("~/")
```

`probe._redact_location` resolves the home it collapses to `~` via `os.path.expanduser("~")` (src/bmad_loop/probe.py:228). On Windows that is `ntpath.expanduser`, which reads **`USERPROFILE`, then `HOMEDRIVE`/`HOMEPATH` — and never `HOME`** (`ntpath.py:362-368`). So on win32 the `monkeypatch.setenv("HOME", ...)` is **inert**: `expanduser("~")` still returns `C:\Users\runneradmin`.

The test nevertheless passed on Windows until now, for an accidental reason: pytest resolves its basetemp, so `tmp_path` was `C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\...` — genuinely *under* the real home. `path.relative_to(home)` therefore succeeded and the `~` branch ran, regardless of the inert monkeypatch.

Lever 1 moves `tmp_path` to `D:\a\_temp\...`, which is no longer under `C:\Users\runneradmin`. `relative_to` raises `ValueError`, `_redact_location` takes its absolute-path branch, and the result starts with `/` instead of `~/`:

```
/<redacted>/a/<redacted>/<redacted>/pytest-0/popen-gw1/test_discover_location_redacts0/<redacted>/abc-123.jsonl
```

Reproduced locally against the real function — both branches match CI exactly:

```
outside-home  -> /a/<redacted>/pytest-of-runneradmin/.../<redacted>/abc-123.jsonl
under-home    -> ~/AppData/Local/Temp/pytest-of-runneradmin/.../<redacted>/abc-123.jsonl
```

**This is a test bug, not a product bug.** `_redact_location` is correct in both branches and the redaction guarantee still holds in the failing output — `.secret-home-dir` became `<redacted>` and the username-bearing `pytest-of-runneradmin` became `<redacted>`. Nothing leaked. The test's other two assertions still pass. What broke is only `startswith("~/")`, which was silently asserting a branch the test never deterministically put itself in on Windows.

**Fix shape.** Set the variable the code actually reads on that platform — i.e. also `monkeypatch.setenv("USERPROFILE", str(tmp_path))` on win32 — so the test exercises the `~` branch deterministically on both OSes, independent of which volume `tmp_path` lands on. That is strictly more honest than what it does today, where the Windows leg has been passing on a coincidence.

Holding for a decision on whether that fix belongs in this PR or a separate one before doing anything further. Not merging.

pbean:
## Second run: fully green (10/10)

Run `31850389639`, head `2f6516a` — the `USERPROFILE` fix in `test_probe.py`. Both Windows jobs now pass, which is the platform the fix targets; Linux stayed green throughout.

### Windows timings across both runs

| run | py3.11 | py3.14 |
| --- | --- | --- |
| 1 (`5ca730a`) | 362 s | 341 s |
| 2 (`2f6516a`) | 250 s | 374 s |

Pooling all four Windows jobs: **mean 332 s, range 250–374 s**, against a **699 s mean / 888 s max** baseline — **−53% on the mean, −58% on the worst observed**.

Linux, pooling both runs (8 jobs): **mean 167 s** vs a **215 s** baseline, **−22%**.

Note the Windows spread: 250 s to 374 s across four jobs on identical code. Runner-to-runner variance is a real component here, so treat any single job's number as noisy and the pooled mean as the signal.

### Where that leaves lever 4

The Windows job now peaks at 374 s — 42% of even the *old* 900 s cap. The `timeout-minutes: 20` bump is no longer preventing anything, and on this evidence 15 would also be safe. Keeping it is cheap insurance against the tail (the cap only ever has to survive the worst runner, and the worst runner here was 50% slower than the best), but it is now a judgment call rather than a fix — happy to drop it back to 15 for symmetry with the Linux job if preferred.

### Remaining before merge

Commit `5ca730a` (the diagnostic step) has confirmed everything it was added for and should come out — `git revert 5ca730a` applies cleanly, since it only adds steps. Leaving that to the PR owner rather than doing it unprompted.

Not merging.

pbean:
@codex review

pbean:
## Measurement: 3 completed runs on the PR head

Run `31850389639`, attempts **1, 2 and 3** on head `2f6516a` — three full green matrices (10/10 jobs each). All figures below are the **"Run tests" step** duration pulled from `actions/runs/<id>/attempts/<n>/jobs`, not the job duration; the step is ~92-96% of the job either way, but mixing the two would flatter the result.

### Baseline

Rebuilt from the **8 most recent successful `main` runs**, same step-level basis. Note this lands *below* the 699 s / 888 s in the PR description — that sample and this one aren't the same 8 runs, and this one excludes cancelled/failed runs. The lower baseline makes every delta below **conservative**.

`main` runs the full 4-version Windows spread (push event); PR runs only the `3.11`/`3.14` boundaries. Comparisons are therefore per-leg, like for like.

### Old vs new — "Run tests" seconds

| leg | baseline mean (n=8) | new mean (n=3) | new samples | Δ |
| --- | --- | --- | --- | --- |
| **windows py3.11** | 651 | **298** | 220, 326, 348 | **−54%** |
| **windows py3.14** | 641 | **334** | 328, 332, 342 | **−48%** |
| windows py3.12 | 625 | *(not run on PRs)* | — | — |
| windows py3.13 | 655 | *(not run on PRs)* | — | — |
| **linux py3.11** | 194 | **144** | 137, 143, 153 | **−26%** |
| **linux py3.12** | 198 | **152** | 142, 145, 168 | **−23%** |
| **linux py3.13** | 184 | **143** | 137, 144, 149 | **−22%** |
| **linux py3.14** | 181 | **144** | 136, 148, 149 | **−20%** |

Pooled over the two comparable Windows legs: **646 s → 316 s, −51%**. Pooled Linux: **189 s → 146 s, −23%**.

No new red on either OS across the three attempts. No retries were added.

### Which levers actually fired

The diagnostic step printed the same values on both Windows legs and both Python versions:

```
default tempdir: C:\Users\RUNNER~1\AppData\Local\Temp
RUNNER_TEMP:     D:\a\_temp
physical: 2
logical:  4
```

| # | Lever | Fired? |
| --- | --- | --- |
| 1 | `TMP`/`TEMP` → `runner.temp` | **Yes.** The default temp was on **C:**, not D: — so this was *not* the no-op the comment hedged about. Every per-test sandbox copy was crossing volumes. |
| 2 | `-n auto` → `-n logical` | **Yes, on both OSes.** `physical: 2` vs `logical: 4` — `auto` was resolving to 2 workers, `logical` gives 4. Linux reports the identical 2/4, which is why Linux moved too. |
| 3 | Drop the 14 `*.sample` hooks | Mechanically yes (the files are gone), but **its individual share is not isolated** — see the caveat below. |
| 4 | `timeout-minutes` 15 → 20 | N/A — a headroom change, it moves no timing. Still earning its keep: the py3.11 leg spreads 220-348 s run to run. |

### Caveat on attribution

All four levers landed together, so this measures **the bundle, not each lever**. I can state which levers changed the runner's actual conditions (above) but not a per-lever share of the 51% without ablation runs. I'm not claiming one.

### `--durations=15` cast stability

Stable, and stable in the way that matters:

- **Identical totals everywhere** — baseline and all three attempts report `5319 passed, 186 skipped, 5 xfailed` on Windows. Collection, skips and xfails are untouched by the `conftest.py` hook change.
- **The two sleep-bound controls are pinned**: `test_wait_for_completion_transcriptless_stop_is_terminal_without_flag` at 15.03 s and `test_stop_run_signals_live_process` at ~10.2 s in every single cast, before and after. A harness distortion would have moved these.

The cast *membership* did shift, in a direction that corroborates rather than contradicts the change:

- Filesystem-heavy tests **left** the cast — `test_a_gitignored_board_story_finished_by_one_run_is_not_re_picked_by_the_next` (8.88 s), `test_sentinel_on_disk_pauses` (6.22 s), `test_a_parked_story_confirms_on_a_fresh_clone` (3.42 s), `test_repeat_two_cycles_then_no_open` (4.77 s). These are the copytree/clone-bound ones — exactly what levers 1 and 3 target.
- CPU-bound render tests **rose within** it — `test_tui_settings.py` entries now dominate, and several got individually *slower* (e.g. `test_settings_screen_enable_toggle_roundtrip` 7.34 s → 9.55 s). Expected: 4 workers on 2 physical cores is more contention per test, so per-test wall-clock rises while total falls.

No unexpected test family entered the cast. I read this as no regression signal.

### Diagnostic commit

Stripped by rebase (`git rebase --onto`), not reverted, so it leaves no trace in history. Head is now `b654ef6` and `ci.yml` contains no `STRIP BEFORE MERGE` step.

Two follow-ups went in with the strip, both turning conjecture into the measured fact:

- `docs:` — `AGENTS.md` and `docs/testing.md` still told readers to parallelize with `-n auto`, which went stale the moment CI switched. They now say `-n logical` and record *why* `auto` isn't a synonym here (psutil is installed on every platform, so `auto` resolves to the physical count).
- `ci:` — the `TMP`/`TEMP` comment hedged with "has historically sat on the slow C: volume" and "a no-op if the default temp is already there"; the timeout comment guessed the levers "should pull the mean well down". Both now carry the measurement.

pbean:
@codex review

pbean:
## Review round: both findings accepted and fixed in `f8f653a`

Codex is clean at zero findings on `f8f653a203`. CodeRabbit's two inline findings (filed against `b654ef6`) were both valid and are addressed below. Every ablation claimed in a docstring was actually executed, not asserted.

### 1. `tests/conftest.py` — regression test for the template cleanup

Accepted. The constraint that made `git init --template=` the wrong tool — `.git/hooks/` and `.git/info/exclude` have to survive — existed only as a comment, so nothing would have caught a later "simplification" back to it. New `tests/test_conftest.py` pins it.

Ablated on **both** axes, because the two halves of the invariant pull against each other and I wanted to know they were independent rather than one implying the other:

| ablation | assertion 1 (`no *.sample`) | assertion 2 (`hooks.is_dir()`) | assertion 3 (`info/exclude`) |
| --- | --- | --- | --- |
| remove the `sample.unlink()` loop | **FAILS** (14 files) | passes | passes |
| empty `git init --template=` | passes | **FAILS** | **FAILS** |

The failure sets are **disjoint** — that disjointness is the actual proof. A single-axis ablation would have shown the test reddening and told me nothing about whether the second half was load-bearing.

Assertion 3 isn't observable in the axis-2 run (pytest stops at assertion 2), so rather than let the docstring claim something I hadn't seen, I verified it separately: `git init --template=<empty>` produces neither `.git/hooks/` nor `.git/info/exclude`.

### 2. `tests/test_probe.py` — ablation record for the negative assertion

Accepted, and the timing is pointed: this is the same test whose win32 leg was — until the commit immediately before — asserting the `~` branch while merely *happening* to be in it, because `tmp_path` sat under the real profile dir. It broke the moment `TMP` moved volumes. That is precisely the vacuity the ablation rule exists to catch, so the record was overdue.

Executed: dropping the `not sanitize.looks_like_identifier(c)` leg from `_redact_location`'s `comp()` reddens it alone, on `".secret-home-dir" not in found.location`:

```
E  AssertionError: assert '.secret-home-dir' not in '~/.secret-h...bc-123.jsonl'
```

That leg is the only one that redacts this component — the name is not identifier-shaped, and the `embeds_current_username` leg beside it does not match it — so the assertion answers about the gate it names rather than about some unrelated reason the string could be absent. The record is undated, matching the existing `Ablation target:` records.

Restores were done from a `cp` snapshot of the **fixed** state, not `git checkout` — a checkout-based restore here would have silently reverted the `USERPROFILE` fix this PR just landed.

### Verification

- `uv run pytest -q -n logical` — **5462 passed**, 44 skipped, 5 xfailed (5461 + the new test)
- `trunk check --all` — clean, 255 files
- Post-strip CI on `b654ef6` — green, 10/10

Issue #603, opened 2026-08-15
TUI never surfaces sweeps_refused: a refused auto-sweep looks identical to one that ran


pbean:
#501 added `RunState.sweeps_refused` (trigger → `not-started` / `failed` / `dirty`) so that a run
whose deferred-work sweep was refused stops looking exactly like one that swept. It is surfaced on
four paths:

- the end-of-run summary (`Engine.summary()` / `RunSummary.render()`, which also reaches the
  ATTENTION file and the desktop notification),
- `documents.status_document`, hence `bmad-loop status --json`,
- the human `bmad-loop status` print,
- both `bmad-loop diagnose` renders.

The TUI is not one of them. `src/bmad_loop/tui/` references neither `sweeps_refused` nor
`sweeps_triggered`, and `tui/data.py` reads `state.json` directly (through `journal.load_state`)
rather than going through `status_document`, so it picks up neither field for free.

An operator watching a run in the dashboard therefore still sees a run that swept and a run whose
sweep was refused as identical — the exact gap #501 closed on every other surface. Under
`[sweep] auto = "run-end"` there is one trigger per run and nothing re-asks it, so the refusal is
not something a later screen will reveal.

Real work rather than a rename: the field has to be read in `tui/data.py` and given a surface (the
run detail pane is the obvious home, beside the other terminal-state fields).

Follow-up to #501.

pbean:
Moved to the local deferred-work ledger as DW-366; tracking it there instead of GitHub.

PR #604, opened 2026-08-15
fix: record an auto-sweep trigger only once its child has started (#501)

### LINKED ISSUE #501: A pre-launch auto-sweep failure burns its trigger for the life of the run
`_maybe_auto_sweep` latches the trigger into `sweeps_triggered` and persists it **before** it calls the sweep factory, so a child sweep that never launched anything still consumes its one chance.

## The ordering

`engine.py:5080-5114`:

- `5086-5087` — early return when `trigger in self.state.sweeps_triggered`
- `5097-5098` — `self.state.sweeps_triggered.append(trigger)` then `self._save()`
- `5110` — `self.sweep_factory(trigger)`
- `5112-5114` — `except Exception` → journal `sweep-auto-failed` + `gates.notify`, parent continues

`sweeps_triggered` is persisted RunState (`model.py:508-510`, serialized `model.py:578`, rehydrated `model.py:607`) and nothing in `src/` ever removes from it — `RunState.clear_pause()` (`model.py:529-535`) does not touch it. So the burn survives a resume, which `cli.py:212-215` already documents as intended.

## Why it matters

At-most-once is right for a child that *ran* — re-firing it could double-commit. But the factory has several arms that refuse or die **before** launching anything:

- the config-integrity refusal (`cli.py:1497-1502`, #461 point 4)
- `ProfileError` → `SystemExit` on an unresolvable `[adapter] name`
- transport/OS faults on the way to spawning

For those, nothing was executed and nothing is at risk of running twice, yet the trigger is spent. Under `sweep.auto = "run-end"` there is exactly one trigger per run (`engine.py:774`), so a single pre-launch failure means the run's deferred-work sweep never happens automatically at all — and the run still records `finished` + `run-complete` (`engine.py:438-441`). The only signals are a journal entry and a best-effort notify (`gates.notify` swallows its own errors, `gates.py:97-105`, `gates.py:120-125`).

This has now been the cost driver behind two separate review findings on #495: it is what makes an over-refusal expensive rather than merely noisy, and it is why the digest's over-coverage had to be argued rather than shrugged off.

## Shape of a fix

Distinguish "refused/failed before launching" from "launched and then failed" — e.g. record the trigger on entry to the factory's launch step rather than before the call, or have the factory signal *not-started* so the caller can un-latch. The `graceful-stop` guard at `engine.py:5088-5096` already returns with the trigger unrecorded, so the un-latched path is not a new concept in this function.

Whatever the mechanism, keep at-most-once for a child that actually started.

Adjacent, not necessarily the same fix: a refused auto-sweep leaves no trace in the run's terminal status or `--json` read model, only in the journal.

Found while validating a reviewer finding on #495 (that the digest hashes `bypass_args` even for a role whose `extra_args` overrides it). That over-coverage is deliberate and documented there; this issue is the reason it is not free.

### LINKED ISSUE #600: A child auto-sweep's SystemExit kills the parent run: exit 1, neither finished nor crashed, orphaned session
`Engine._maybe_auto_sweep` guards the child sweep with `except Exception`, but the contract it is
written against is "a paused or failed child never interrupts this run". `SystemExit` is a failed
child that is *not* an `Exception`, so it was missed here, by every arm of `_run_inner`, and by
`cli.main`.

## Reachable, not theoretical

`runsetup.make_adapters` raises `SystemExit` at five sites — unresolvable profile, unknown adapter
kind, a kind that fails to load, a construction failure, an unusable multiplexer. The last one sits
behind `mux_usable`, a live uncached `shutil.which` re-run on every call, so a mux binary that
disappears (or a `PATH` that changes) mid-run is enough to reach it.

## What the operator is left with

The `SystemExit` unwinds through `_run_inner`'s `finally` — which persists the state, including the
trigger latch the pre-#501 code had already burned — and ends the process at exit 1. The parent run
is left:

- neither `finished` nor `crashed` in `state.json`,
- with no `run-complete` journal entry,
- with its agent session orphaned in the multiplexer.

Nothing reconciles that shape, so the run also lingers in `bmad-loop list` looking resumable.

## Fix

Widen the guard to `except (Exception, SystemExit)`. Deliberately **not** `BaseException`:
`KeyboardInterrupt` must keep escaping, because the nested-child re-raise in `_run_inner` depends on
it reaching the owner.

Found while fixing #501 and fixed on that branch (`bc784e3`), since #501's own fix moves code inside
this guard.

### LINKED ISSUE #601: A swallowed RunStopped from a child auto-sweep leaves the parent run unstoppable
The mirror image of the `SystemExit` hole: `RunStopped` subclasses `Exception` but is not a failed
child at all, so `Engine._maybe_auto_sweep`'s `except Exception` swallowed the operator's stop and
journaled it as `sweep-auto-failed`.

## Two consequences, and the second is the bad one

1. **The parent runs on to `finished`.** A stop the child honoured is recorded as a completed run.

2. **The parent becomes unstoppable.** Only the outermost engine in the process owns the signal
   handlers — `_install_stop_signals` returns early when `Engine._stop_signals_owner` is already
   set, so a nested auto-sweep run deliberately lets the exception propagate up to the owner. The
   SIGTERM that lands during a child sweep is therefore handled by the **parent's** handler, which
   sets `self._stopping = True` *before* raising `RunStopped`. Once that raise is swallowed, every
   later signal returns early at the `if self._stopping: return` latch. The operator's second,
   third and fourth `bmad-loop stop` do nothing at all, and the run continues to completion.

## Fix

An `except RunStopped: raise` arm ahead of the failure arm — the child's hard-stop arm re-raises it
precisely so the owner records the stop.

It is also not a *refusal*, and the arm has to sit ahead of the recording arms to keep that true: by
the time a child can reach its own stop arm it is well past `compose_sweep`'s success boundary, so
the trigger is legitimately spent and no `sweeps_refused` entry is written.

Found while fixing #501 and fixed on that branch (`bc784e3`), since #501's own fix moves code inside
this guard.

### LINKED ISSUE #602: A launch that aborts standing up its adapters strands an empty finished=False run dir
`runsetup.make_adapters` raises `SystemExit` at five sites, every one of them *after* `save_state`
has published a run directory carrying `finished=False` / `crashed=False` and no `run-start`.

Nothing reconciles that shape: `runs.reconcile_stale_worktrees` is finished-ONLY by design (a
stopped run is still resumable, so its worktree is left for `resume`/`clean` rather than stranded
out from under the operator). So the directory lingers in `bmad-loop list` looking resumable — for a
run that never started and has nothing to resume.

Both composers are affected (`compose_run` and `compose_sweep`), and the escape need not be a
`SystemExit`: an adapter family's lazy load thunk is first imported here, and it is deliberately
never invoked by `validate` or `bmad-loop adapters`, so an `ImportError` from an optional dependency
has had no earlier gate and strands the run the same way.

## Fix

Make composition atomic from the first published artifact: on any escape, remove the run directory
and its out-of-tree state, then re-raise the original error unchanged.

Two things the fix has to get right, both of which only surfaced under review:

- **`runs.delete_run`, not a bare `rmtree`.** `delete_run` also drops the run's out-of-tree state
  dir, which is what covers the config-digest stamp the composers write between the state and the
  pid.

- **The run id must be _claimed_, not probed.** `delete_run`'s `force=False` guard is
  liveness-scoped — it refuses a *live* session, and the dangerous case is precisely the run that is
  not live (paused, stopped, finished). Since `Journal.__init__` mkdirs with `exist_ok=True`, the
  hidden `--run-id` flag (on both `run` and `sweep`) pointed at an existing run silently adopted
  that run's directory and published its own `state.json` over it. Adding an unwind on top of that
  would have deleted a pre-existing run outright — journal, logs and tasks. The id is now claimed
  with `mkdir(parents=True, exist_ok=False)`, which makes creation and collision-refusal one atomic
  step, and the claim sits *outside* the composers' `try` so a refusal can never reach the unwind
  arm and delete the very run it protects.

A removal that itself fails is reported — stderr `warning:` plus a `composition-unwind-failed`
journal entry — rather than passing silently. It stays best-effort (`force=False`, under
`suppress(Exception)`) so a cleanup failure never *replaces* the launch error, but not raising is
not the same as not reporting.

Found while fixing #501 and fixed on that branch (`9c7a284`, `aab2b19`, `8e35247`): #501 moves the
sweep trigger's latch onto `compose_sweep`'s success boundary, which is only a meaningful boundary
once a failed composition leaves nothing behind.

pbean:
Closes #501.
Closes #600.
Closes #601.
Closes #602.

Seven commits on the auto-sweep parent/child boundary. `6d240c6` is the #501 fix proper; the two
before it are prerequisites it depends on and each stands alone (now filed as #600/#601 and #602),
`aab2b19`/`8e35247` are review fallout from the second, and `b88ad82` is the visibility half.

## What #501 actually buys — read this before the issue text

The issue frames the cost as a lost *retry*: the trigger is spent, so "the run's deferred-work sweep
never happens automatically at all." Both call sites were traced before the fix was written, and
**that framing is misleading — an un-latched trigger is essentially never re-asked.** Stating it
here so nobody reviews this change against a promise it does not make:

- **`run-end`** fires from `_loop`, whose return lands directly on `state.finished = True` in
  `_run_inner`. `cli._resume_paused_run` refuses a finished run outright. There is no later ask.
- **`per-epic`** fires from `_epic_boundary`, and the boundary *is* `state.current_epic != story.epic`
  — a field that advances within the same frame, either before the gate's `RunPaused` or on return
  into `_loop`. Nothing in between can pause the run: `gates.notify` never raises (it swallows its
  own `OSError`), and `_emit` isolates both hook kinds (`_HookError` for a declarative hook,
  `except Exception` for an in-process one) and *returns* vetoes rather than raising them — which
  `_epic_boundary` does not even resolve.

So the retry survives only in a **crash window**: a process death or a `BaseException` across a few
statements. No code, comment, test or changelog line in this PR promises a retry, and the stale
comments that asserted "burns the trigger for the life of the run" were *corrected* rather than
deleted — a wrong config pin still refuses every trigger, it just no longer spends them.

What the change buys on every other run is **record honesty**:

1. `sweeps_triggered` stops claiming sweeps that never ran. It is durable `RunState`, nothing in
   `src/` ever removes from it, and `bmad-loop diagnose` renders it.
2. A transient `git status` timeout stops permanently spending the run's only trigger. This is not a
   config-error path at all: `verify.worktree_clean` fails closed on `GitError`, and `_run_git`
   reports a `subprocess.TimeoutExpired` as exactly that — so a slow filesystem, not a dirty tree,
   silently consumed the sweep. Under `sweep.auto = "run-end"` there is exactly one trigger per run.
3. A refusal now leaves a trace outside the journal (below).

## The #501 fix — `6d240c6`

- Hoist the worktree check **above** the record, and split its journal line with a `reason` so a git
  fault reads differently from real local changes (`sweep-auto-skipped-dirty` now carries
  `reason="git-error"` or `reason="dirty"`).
- Define a `SweepFactory` Protocol beside `RunPaused`/`RunStopped` (no import cycle: `engine` imports
  neither `cli`, `runsetup` nor `sweep`) carrying a **required** keyword-only `started` thunk — no
  default, because a defaulted no-op would let an un-updated implementation claim "started" for a
  child that never launched, the one direction that costs a duplicate sweep. Threaded through
  `cli._start_sweep` as `on_started` and fired from `runsetup.compose_sweep` once composition has
  succeeded.
- A plain return from the factory latches too; the thunk's only job is to classify a **raise**.
  `latch()` is idempotent and sets its in-memory flag before the write.
- A raise that beat composition journals `sweep-auto-not-started` and keeps its own `gates.notify`:
  the #461 config-integrity refusal is a security event and must not go quiet just because it
  stopped being permanent. A child that fails *after* its run dir exists still spends the trigger —
  that run is resumable, so re-firing would duplicate it.

At-most-once for a child that actually started is preserved, which is the one constraint the issue
asks for unconditionally.

### Why `on_started` sits where it does

At `compose_sweep`'s success boundary rather than at `save_state`, and only because #602's fix (in
this PR) made a failed composition unwind its own run dir. From that boundary the child owns a
published run dir, `sweep.json` and a live pid file, so a later failure leaves something
`bmad-loop resume` can pick up.

The docstring records the exception rather than asserting the guarantee flatly: `_unwind_composition`
is best-effort, so a refused unwind *can* leave a resumable child behind. At-most-once holds anyway,
because the flag is set **before** the write — the unwind decides only what a refusal costs.

## The prerequisites

- **#600 / `bc784e3` — a child `SystemExit` killed the parent.** The guard was written over
  `Exception` while the contract was stated over "a paused or failed child", and the two sets differ
  in *both* directions. `SystemExit` is a failure that is not an `Exception` (`make_adapters` raises
  it at five sites, reachably — the unusable-mux refusal sits behind a live `shutil.which`), so it
  unwound through the `finally` and ended the process at exit 1, leaving the parent neither
  `finished` nor `crashed` with an orphaned session.
- **#601 / `bc784e3` — a swallowed `RunStopped` left the parent unstoppable.** The reverse: an
  `Exception` that is not a failure at all. Only the outermost engine owns the signal handlers, so
  the SIGTERM landing during a child sweep is the **parent's** handler setting `self._stopping = True`
  before raising. Swallowing that raise let the parent run on to `finished` *and* made every later
  `bmad-loop stop` return early at the latch. Deliberately still not `BaseException`:
  `KeyboardInterrupt` must keep escaping for the nested-child re-raise in `_run_inner`.
- **#602 / `9c7a284` + `aab2b19` + `8e35247` — a failed composition stranded an empty run.**
  `make_adapters`' five `SystemExit`s all fire *after* `save_state` published a run dir carrying
  `finished=False` / `crashed=False` and no `run-start`; `runs.reconcile_stale_worktrees` is
  finished-ONLY, so nothing reconciled it. Both composers now unwind. `runs.delete_run` rather than a
  bare `rmtree`, because it also drops the out-of-tree state dir carrying the config-digest stamp.

  Two corrections came out of the review gate on that commit, and both are load-bearing:

  - **`aab2b19`** — `delete_run`'s `force=False` guard is **liveness-scoped**, not
    collision-scoped: it refuses a *live* session, and the dangerous case is precisely the run that
    is not live. Because `Journal.__init__` mkdirs with `exist_ok=True`, a hidden `--run-id` pointing
    at an existing run adopted that run's directory — so the new unwind would have deleted a
    pre-existing run outright. The id is now **claimed** (`mkdir(exist_ok=False)`, creation and
    refusal in one atomic step) *outside* the composers' `try`, so a refusal can never reach the
    unwind arm.
  - **`8e35247`** — the unwind's `suppress(Exception)` was defended as "a cleanup failure must not
    replace the launch error". True, and silent about *surfacing* it. It now reports: stderr
    `warning:` plus a `composition-unwind-failed` journal entry. The suppress around the journal
    write itself is deliberate — `Journal.append` opens with `"a"` and does not mkdir, so when the
    run dir is what failed to go away it raises `FileNotFoundError`, which unsuppressed would
    replace the launch error after all.

## The visibility half — `b88ad82`

#501's closing paragraph notes that a refused auto-sweep leaves no trace in the run's terminal status
or the `--json` read model, only in the journal. Since the un-latched trigger is never re-asked, that
trace *is* the remedy, so it is in scope here rather than deferred.

`RunState.sweeps_refused` (trigger → reason) is surfaced by the end-of-run summary, `bmad-loop status`,
`status --json` and both `bmad-loop diagnose` renders. `_run_inner` already pipes `summary.render()`
into `gates.notify` on both the graceful-stop and finish branches, so the new line reaches the
ATTENTION file and the desktop toast with no new call site.

Three design constraints worth naming for review:

- **The reason is a closed vocabulary of slugs** — `not-started` / `failed` / `dirty` — never an
  exception message. `sanitize.guard` **raises** `LeakDetected` on a home-path hit (only
  pseudonymizer strays are repaired by substitution; the hard rules fail closed), so a free-form
  `str(e)` would make `bmad-loop diagnose` fail outright on exactly the runs worth dumping.
- **It is a sibling of `sweeps_triggered`, not a widening of it.** That field is the re-fire latch
  the #461 gate depends on, and `diagnostics.py` filters it through a per-*element*
  `looks_like_identifier` loop — a dict would silently degrade that to keys-only. A trigger
  legitimately appears in both (`failed` = a child that started, then failed).
- **`sweep-auto-suppressed` is deliberately excluded.** That arm returns *ahead* of the latch, so a
  resume can still fire the trigger and an entry there would go stale. Pinned by an inverse ablation.

Additive per `machine.py`, so `STATUS_SCHEMA_VERSION` stays 1; no exit code moved. Both dirty arms
collapse to one `dirty` slug — the journal keeps `git-error` vs `dirty` apart for forensics, but the
operator's next action is identical. The render names a **clean worktree**, not just
`bmad-loop sweep`, because `cmd_sweep` hard-refuses an unclean tree and a follow-up that omitted it
would walk the operator into a second refusal.

## Docs corrected — `c5ae67f`

- `docs/FEATURES.md` said flatly that "a failed/paused child sweep never interrupts the parent run",
  which was false in one direction before this branch (#600) and is *deliberately* false in another
  after it (a stop or Ctrl-C delivered through the child propagates, by #601's fix). It now says
  both, and documents `sweeps_refused`.
- `runsetup.resolve_profiles` argued a lost race round costs the writer nothing because "the next
  auto-sweep trigger deals again". The repeat is the writer's, never the orchestrator's — and under
  `[sweep] auto = "run-end"` a run has exactly one trigger. It is `per-epic` that hands out further
  rounds.
- `_maybe_auto_sweep`'s lead sentence keeps the contract but now names its two exceptions, so a
  reader who stops at the lead does not carry away the claim the four paragraphs below it spend their
  length qualifying.

Four further sites that asserted the trigger burn were corrected in earlier phases of this branch
(`runsetup.py`'s digest-completeness comment and adapter-load comment, two in `cli.py`, and the
`test_engine.py` suppression docstring).

## Scope

Surfacing `sweeps_refused` in the **TUI** is left out: `src/bmad_loop/tui/` references neither
`sweeps_refused` nor `sweeps_triggered`, and `tui/data.py` reads `state.json` directly rather than
going through `status_document`, so it is real work rather than a rename. Follow-up filed as #603.

## Verification

- `uv run pytest -q -n logical`: **5486 passed**, 44 skipped, 5 xfailed
- `uv run pyright`: 0 errors
- `trunk check --all --no-fix`: no issues

Every gate was ablated **singly** against `cp` backups (never `git checkout`), each required to give
`rc == 1` — `rc == 4` is a collection error, i.e. a fake green. 17 axes on the visibility half alone,
because "the record is written" and "the record is written *from this arm*" are different claims:
deleting `_record_sweep_refusal` from each of the four refusal arms reddens one test each, all four
disjoint.

Four results worth carrying into review rather than just asserting:

- **Dropping the `looks_like_identifier` key-half and value-half filters reddens the same single
  test**, so the exit code cannot distinguish them — only the assertion diff can (pytest's "Omitting
  2 identical items" plus the one differing entry). A test whose docstring claims *which element* an
  ablation breaks is graded by reading the diff, not the rc.
- **A refusal test can stay green while the destructive behavior happens.** Moving the run-dir claim
  inside the composers' `try` leaves the `SystemExit` and its message unchanged, so
  `pytest.raises(..., match="already exists")` still passes; the only failing line is the
  prior-run-untouched assertion. The positive control that carries the weight is
  `probe.published == {}` — proof the refusal beat `make_adapters`.
- **`test_auto_sweep_no_refire_on_resume` stays green with the latch removed entirely**, because
  `current_epic` also closes that boundary. Recorded as a green ablation in the test itself: any
  "the trigger is not re-fired" assertion driven through a whole `run()` measures
  `finished`/`current_epic`, not the latch.
- **Aliasing the state dict instead of snapshotting it reddens eight tests, not one** — six in
  `render()` with `ValueError: too many values to unpack`, since iterating a dict yields bare keys.
  `RunSummary` is `frozen=True`, so the field is a `tuple[tuple[str, str], ...]`: a `dict` there
  silently removes hashability, and nothing hashes a `RunSummary` today, which is exactly why it
  would have shipped.

Several ablation records were first written up as "fails alone" and were wrong; the sets above are
the measured ones.

## Summary by CodeRabbit

- **New Features**
  - Auto-sweep status and diagnostics now show refused triggers with actionable reasons.
  - Refused sweeps remain available for retry instead of being marked completed.
  - Text and JSON status outputs include refusal details.

- **Bug Fixes**
  - Improved handling of failed or interrupted auto-sweeps, including stop and Ctrl-C behavior.
  - Prevented duplicate run IDs from overwriting existing runs.
  - Failed run setup now cleans up partial artifacts.
  - Writes are safer when replacing files or following symlinks.


pbean:
@codex review

pbean:
@codex review

pbean:
**Codex P3 (`runsetup.py:960`, extend the unwind guard to the post-claim setup): valid, fixed in `6d96b36`.**

Confirmed the mechanism: `_claim_run_dir` creates the directory, and the guard opened two statements
later at `save_state`, so an abort at `Journal(...)` / `build_run_state(...)` (and `policy.to_dict()`
in `compose_sweep`) left it behind. The claim publishes the *first* artifact — the directory is
exactly what a later `--run-id` collides with — so the "atomic from the first published artifact"
comment sitting above the `try` was describing a window two statements wider than the code's.

Reachability, graded rather than assumed: `Journal.__init__` mkdirs `exist_ok=True` over a directory
this frame just created, and `build_run_state` is a pure constructor (no I/O, no raise), so neither
fails on its own. Your "most plausibly Ctrl-C" is the honest characterization, and it is sufficient
justification — the arm is `BaseException` precisely because a signal lands between arbitrary
statements.

Severity is narrower than the sibling case this PR already fixed, and worth stating so the fix is not
oversold: the stranded dir has no `state.json`, and `runs.list_run_dirs` is state.json-gated, so it
never appears in `bmad-loop list` and never looks resumable. The only way to meet it is a later launch
with that explicit `--run-id` being refused by a directory holding nothing.

**One thing the suggestion did not mention, and it is the part that could have gone wrong:**
`_unwind_composition(project, run_dir, journal)` requires a `Journal`. Moving the `try` above
`journal = Journal(run_dir)` means the except arm can now run with `journal` unbound — an
`UnboundLocalError` that would *replace* the launch error, which is the one outcome that arm forbids.
So the signature is now `Journal | None` with an explicit `if journal is not None:` around the
`composition-unwind-failed` append. Explicit rather than relying on the surrounding
`suppress(Exception)`: an `AttributeError` on `None` **is** an `Exception` and would have been
swallowed, so the code would have worked by accident while reading as though a journal were
guaranteed. The stderr `warning:` report — the half that matters — is unaffected either way.

The collision claim stays outside the guard, as you noted it must.

Tests: two, one per composer, and they redden **disjointly** under separate ablations (moving each
`try` back down fails exactly one test each). Each carries a positive control recording that the
directory existed at the moment of the raise — an "is it gone" assertion passes just as happily for a
directory that was never created. The `compose_sweep` one drives `Journal` itself so it exercises the
`journal is None` path; its docstring records that the `None` guard is a **green** ablation (the
suppress absorbs the AttributeError), so that guard is graded by the annotation and the docstring,
not by an exit code.

Gates on `6d96b36`: 5488 passed / 44 skipped / 5 xfailed, pyright 0 errors, `trunk check --all
--no-fix` clean.

pbean:
@codex review

Issue #608, opened 2026-08-15
De-duplicate `needs_strict_codec` between `tests/conftest.py` and `tests/test_verify.py`


pbean:
`needs_strict_codec` now exists twice in the test tree: the shared copy in `tests/conftest.py` and a local one in `tests/test_verify.py`.

## State

The marker skips a test when the host codec can decode byte `0xff` — every ISO-8859-x and cp125x codec maps all 256 byte values, so under such a locale a strict-decode test goes vacuously green rather than exercising the fault it is named for.

- `tests/conftest.py:65` — the shared copy, added by the #380/#383 bundle so phase 2's tests could import it. Used by `tests/test_multiplexer.py`, `tests/test_hook_bus.py` and `tests/test_probe.py`.
- `tests/test_verify.py:59` — a local copy predating it, with its own `_codec_rejects_bad_byte` helper. Used at `:296`, `:1090` and `:1114`.

The two are byte-identical in condition and skip reason.

## Why the duplication was left in place

`test_verify.py` is otherwise untouched by the #380/#383 fix. De-duplicating it would have dragged a large, unrelated file into a bundled fix whose whole discipline was to claim exactly what it fixed and no more. Deliberate, and named in that PR's "Out of scope" section — filed here so it does not simply rot.

## Asked for

Delete the local copy and its `_codec_rejects_bad_byte` helper from `tests/test_verify.py`; import the shared one from `conftest`, as the other three files do.

Worth carrying over rather than dropping: the comment above the local copy (`tests/test_verify.py:56-58`) records that CI is always on the exercising side of this skip — the Linux legs run UTF-8 and the Windows legs set `PYTHONUTF8=1` (`.github/workflows/ci.yml`). That fact belongs with the shared definition, which currently does not state it.

## Verification

Mechanical, but confirm the tests still **run** rather than newly skipping — a marker refactor that silently flips three tests to skipped is indistinguishable from green. `-rs` on the affected file, or an explicit assertion on the collected outcome, is enough.

pbean:
Closing as not worth tracking: the two copies are identical; pure test de-duplication.

Issue #607, opened 2026-08-15
The unity plugin scripts decode child output strictly behind `TimeoutExpired`/`OSError` guards


pbean:
The shipped unity plugin scripts capture child output with `text=True` behind guards that name only `TimeoutExpired`/`OSError` — the same narrow-net defect #380 and #383 closed in orchestrator core, still open in plugin data.

## The defect genus

`subprocess.run(..., text=True)` decodes with the **locale** codec at `errors="strict"`. The fault that raises is `UnicodeDecodeError` — a `UnicodeError`, so a `ValueError`. It is neither an `OSError` nor a `SubprocessError`, so a guard naming only those does not catch it, and the decode happens *after* the child has exited, so a child that succeeded takes its exit code down with the raise.

That is the same shape as #383 (`plugins/bus.py`, `probe.py`) and the decode half of #380 (`adapters/tmux_base.py`). Those were fixed by adding `errors="replace"` — stop raising, rather than catch more.

## Sites

Under `src/bmad_loop/data/plugins/unity/`:

- `unity_ready.py:113` (`wait-for-ready`) and `unity_ready.py:129` (`run-tool`) — both guarded by `except subprocess.TimeoutExpired` only. Output is sliced (`[-500:]`) into hook advisory text.
- `unity_quiesce.py:111`
- `unity_plugin.py:408`
- `unity_setup.py:182`, `:324`, and `:402` — `:402` (`setup-mcp`) has no `try` at all; it writes `proc.stdout + proc.stderr` straight to stderr.

The decoded bytes are output from the Unity CLI and the Editor bridge — an operator's own toolchain, not a path, so nothing constrains them to the run's encoding. A project path, asset name or C# diagnostic carrying a non-UTF-8 byte under a non-UTF-8 locale is enough.

## Why it was not folded into the #380/#383 fix

These are **shipped plugin data**, not orchestrator core, and neither issue names them. Folding them in would have widened a bundled fix past the two issues it closes. Filed separately so the decision is recorded rather than implied.

## Asked for

- Add `errors="replace"` to each capture, matching the fix used in `bus.py`/`probe.py`. `"replace"` rather than `surrogateescape` for the same reason as there: this text feeds forward into advisory strings and journals, never back to a filesystem API, and `surrogateescape` would seed it with lone surrogates that cannot be encoded to UTF-8 — converting a decode crash into the encode crash #380 fixed.
[...]

Issue #606, opened 2026-08-15
Record why `_write_heartbeat`'s narrow `except OSError` is inert, so the next codec sweep does not re-find it


pbean:
`_write_heartbeat` is byte-for-byte the shape of the encode bug #380 just fixed, but is currently unreachable — record why, so the next sweep does not re-find it and a future payload change is caught.

## What it looks like

`GenericTmuxAdapter._write_heartbeat` (`src/bmad_loop/adapters/generic.py:367`) is the same construction that made `_append_diag_jsonl` crash a run:

```python
try:
    (self.tasks_dir / task_id / "heartbeat.json").write_text(
        json.dumps(payload, ensure_ascii=False), encoding="utf-8"
    )
except OSError:
    pass
```

`ensure_ascii=False` means `json.dumps` returns a `str` that may still hold a lone surrogate; the fault fires later, at the UTF-8 encode inside `write_text`, as `UnicodeEncodeError` — a `UnicodeError`, so a `ValueError`, which `except OSError` does not catch. That is exactly the defect PR for #380 widened to `(OSError, UnicodeError)` in the sibling writer.

## Why it is inert today

Both call sites pass scalars only, so the encoded string is surrogate-free by construction:

- `src/bmad_loop/adapters/generic.py:634-641`
- `src/bmad_loop/adapters/opencode_http.py:1080-1087`

Both build the identical payload — `ts` (float), `remaining_s` (float), `stall_armed` (bool), `stall_nudges_sent` (int). No path-derived or otherwise operator-controlled string reaches it. There is no input that can produce the crash, so widening the guard here would add an arm that nothing could ablate into failing — a guard with no test that can prove it.

## Asked for

Not a behavior change. A comment on `_write_heartbeat` recording:

1. that the narrow `except OSError` is deliberate, and why — the payload is scalar-only at every call site;
2. that adding a path-derived or otherwise arbitrary string to that payload requires widening the guard to `(OSError, UnicodeError)` at the same time.

The value is twofold: a future codec sweep stops re-finding this site and re-litigating it, and the condition that would make it live is written down next to the code that would break.

## Context

Named and declined in the #380/#383 bundle PR's "Out of scope" section. Deliberately left alone there to keep that diff to the reachable faults.

pbean:
Closing as not worth tracking: the guard is unreachable today; the ask was a comment only.

Issue #435, opened 2026-08-02
cleanup --json can report a surviving ctl window as removed


dracic:
Follow-up to #313 (PR #434).

Since #313, the psmux backend's `kill_window` verifies whether the kill actually landed (a liveness listing gates the key cleanup) — but the seam signature is `kill_window(target) -> None`, so that knowledge dies inside the adapter. `cleanup`'s read model (`documents.py`, `ctl_windows.removed`) lists every window a kill was *attempted* on, and the human-readable summary (`cli.py`, "removed N session(s), M ctl window(s)") counts the same way. A kill that fails — window still in the liveness listing — is reported as removed.

Sketch: let the seam return whether the window is verifiably gone (`bool`, tri-state if "unverifiable" should stay distinct from "survived"), thread it through the prune fan-out, and report survivors separately in the `--json` payload — which is a schema-versioned contract, so the shape change needs a version bump.

Low urgency: the failure mode is a too-optimistic report, not a wrong action, and the next cleanup pass retries the survivor.

PR #609, opened 2026-08-15
fix: close the codec except-family holes in the diag writer, tmux captures, and plugin hooks (#380, #383)

### LINKED ISSUE #380: Codec except-tuple asymmetry in _append_diag_jsonl (and POSIX-only strictness in tmux_base)
Found while sweeping for copies of the encode-half defect fixed in #376 (the tail of `_worktree_local_exclude` named `UnicodeDecodeError`, so the `write_text` encode fault escaped; widened to `UnicodeError`).

Exactly one other real instance of the asymmetry exists.

## `adapters/generic.py:308` `_append_diag_jsonl`

```python
json.dumps(payload, ensure_ascii=False)
...
except OSError:
```

The `ensure_ascii=False` is load-bearing: with the **default** `ensure_ascii=True`, a lone surrogate is escaped to ASCII and writes fine; with `False` it reaches `fh.write` as a **`UnicodeEncodeError`**, which is a `ValueError`, not an `OSError`.

Reachable the same way as #376: payloads carry `spec=str(path)` derived from `impl_artifacts.glob("*.md")`, so a POSIX filename with a non-UTF-8 byte arrives surrogate-escaped.

It escapes the outer guard at `generic.py:1819` too, defeating two stated contracts:

- `:300` — "an unwritable run dir must never break the completion loop"
- `:1826` — "a best-effort rescue must never escalate a clean stall/timeout into an exception, which the engine does not contain per-task"

Minimal fix: widen `:308` to `(OSError, UnicodeError)`, matching what #376 landed in `install.py`.

## Deliberately not filed as findings

Three structural look-alikes do read+write under a decode-only tuple but are **unreachable**, not real: `engine.py:2258`, `runs.py:857`, `runs.py:922` all write strictly-decoded file text plus ASCII literals (a git sha, a status token), so the written string is surrogate-free by construction. `journal.py:43`, `resolve.py:154`, `install.py:682`, `worktree_flow.py:281` are safe because `ensure_ascii` defaults to `True`.

`psmux_backend.py:841` is the mirror done right and is a good model — write-only block, encode-only guard, and its subprocess leg cannot raise a decode fault because `_ERRORS = "backslashreplace"`.

## Related

`adapters/tmux_base.py:88` `_run` documents this same hazard and mitigates it on **Windows only** — `_ERRORS = None` (strict) on POSIX, `backslashreplace` on the Windows leaf, whose comment explains the byte "degrades visibly instead of raising mid-capture". Guards at `:169`, `:357`, `:394` catch `(SubprocessError, OSError)`, so a POSIX decode fault escapes. `@bmad_project` stores `str(project.resolve())` (`runs.py:303`), a raw filesystem path. `:351` happens to survive only because it includes `ValueError` for an unrelated `int()` parse. Same one-sided-mitigation shape as #379; folding it in here rather than filing a fourth issue.

All pre-existing.

### LINKED ISSUE #383: plugins/bus.py _run_subprocess decodes hook output strictly; a non-UTF-8 byte escapes _HookError and crashes the run
Found while fixing #378 (PR #382). Same genus, same fix shape, but a different call site and a different guard — and it is untracked by #377–#381, so filed separately.

`plugins.bus._run_subprocess` is the default transport for declarative plugin hooks. It shells out to an operator-authored command and decodes the child's output strictly:

```python
    try:
        proc = subprocess.run(  # nosec B602 - operator-authored plugin command
            cmd,
            shell=True,
            cwd=cwd,
            env=env,
            capture_output=True,
            text=True,
            timeout=timeout,
        )
    except subprocess.TimeoutExpired as e:
        raise _HookError(f"timed out after {timeout}s") from e
    except OSError as e:
        raise _HookError(str(e)) from e
    return proc.returncode, proc.stdout + proc.stderr
```

`text=True` decodes with the locale codec at `errors="strict"`. `TimeoutExpired` and `OSError` are the whole net; `UnicodeDecodeError` is in neither.

## Why it fires

Exactly the reason #378 does: the decoded bytes are arbitrary output from an operator's own tool, not a path, so nothing constrains them to the run's encoding. A hook that shells to a linter, a formatter, a test runner or a git wrapper can emit a latin-1 filename in a diagnostic, a `bytes` repr, or any non-UTF-8 byte in a non-UTF-8 locale.

## Consequence

`_HookError` is the designed channel for "the hook could not run to completion": `_dispatch` catches it, journals `plugin-hook-error`, and either fails open or (blocking + `fail_closed`) adds a `defer` veto. A `UnicodeDecodeError` bypasses that channel entirely and escapes the bus into the engine — so a plugin hook, whose entire contract is that a transport failure is classified rather than fatal, crashes the run instead. Worse than the verify case in one respect: the hook's own exit code is lost too, so a hook that actually *passed* takes the run down.

## Suggested fix

`errors="replace"`, mirroring the #378 fix in `verify.run_verify_commands` (PR #382). Same reasoning applies verbatim: the captured output is fed to `_apply_stdout` and the journal, nothing round-trips to a filesystem API (so no surrogate escaping needed), and locale decode is right because these are arbitrary host tools rather than our own UTF-8 CLI. Worth a real-child regression test for the same reason #378 and #374 needed one — a monkeypatched `subprocess.run` never runs the stdlib decode, so it passes identically with the bug restored.

## Lower-priority same-genus sightings

Both catch `(OSError, subprocess.SubprocessError)`, which also excludes `UnicodeDecodeError` (a `ValueError`), around a `text=True` capture:

- `probe.py:185` `_run_capture` — documented "Never raises" (`run_version_help`), but a `--version`/`--help` banner with a non-UTF-8 byte does raise. Lower priority: argv-based, and the output is a probe banner, but the never-raises claim is the contract being broken.
- `tui/app.py:808` `_commit_subject` — a commit subject is author-supplied text; a non-UTF-8 subject in a repo with a non-UTF-8 `i18n.commitEncoding` crashes a TUI cell render that otherwise degrades to `""`.

Both are the same one-word fix if this is taken as a sweep.

Related: #378 / PR #382 (the verify-commands original), #377 (`_run_git`, a different guard and fix).

pbean:
Closes #380
Closes #383

## Problem

Both issues are the same defect genus: **an exception net that names one direction of the Unicode error hierarchy while the other escapes it.**

`UnicodeEncodeError` and `UnicodeDecodeError` are both `UnicodeError` ⊂ `ValueError`. Neither is an `OSError`, and neither is a `SubprocessError`. So:

- `except OSError` catches **neither** direction;
- the post-kill rescue's `except (OSError, UnicodeDecodeError)` catches the decode half while the **encode** half escapes.

Every site in this PR is a best-effort observation path whose stated contract is that it must never escalate — a diagnostic breadcrumb, a liveness capture, a plugin-hook transport, a `--version` banner. Every one of them escalated to a run-killing crash.

**#380, encode half.** `_append_diag_jsonl` (the shared writer behind `_note_resultless_stop` and `_note_lifecycle`) serializes with `json.dumps(..., ensure_ascii=False)` and writes through a handle pinned to `encoding="utf-8"`. `ensure_ascii=False` is what makes the bug exist: `json.dumps` does **not** raise on a lone surrogate — it returns a `str` that still holds it, and the fault fires later, at the encode inside `fh.write`. A POSIX filename holding a non-UTF-8 byte arrives surrogate-escaped, and six `_note_resultless_stop` call sites interpolate such a path straight into `detail`. That defeated two stated contracts: "an unwritable run dir must never break the completion loop", and the rescue that "must never escalate a clean stall/timeout into an exception, which the engine does not contain per-task".

**#380, decode half.** `tmux_base._ERRORS` was `None` — the strict handler — on POSIX, while the Windows leaf already set `"backslashreplace"`. A tmux capture holding an undecodable byte therefore raised out of `_run`, the one place tmux is spawned.

**#383.** `subprocess.run(..., text=True)` decodes with the **locale** codec at `errors="strict"`. `plugins/bus.py` guarded `(TimeoutExpired, OSError)`; `probe.py` guarded `(OSError, SubprocessError)`. Neither names the fault. The bus half is the run-crashing one: `_HookError` is the designed channel for "the hook could not run to completion" — `_dispatch` catches it, journals `plugin-hook-error`, and either fails open or adds a defer veto — and failure isolation is the bus's stated contract per its module docstring. A `UnicodeDecodeError` bypassed that channel entirely and escaped into the engine. Worse, the decode happens inside `run`, *after* the child has exited, so **a hook that actually passed took its own exit code down with the run.** The decoded bytes are arbitrary output from an operator's own tool — not a path, so nothing constrains them to the run's encoding.

## Fix

| File | Change |
| --- | --- |
| `adapters/generic.py` | `_append_diag_jsonl`'s guard widened `except OSError` → `except (OSError, UnicodeError)` — covers both directions at once, the form #374/#375 already landed in `install.py`. |
| `adapters/tmux_base.py` | `_ERRORS = "backslashreplace"` on the base class. |
| `plugins/bus.py` | `errors="replace"` on `_run_subprocess`, the declarative-hook transport. |
| `probe.py` | `errors="replace"` on `_run_capture`, the version/help capture. |

**Why `tmux_base` is fixed at the source rather than at the catch sites.** There are fourteen catch sites still naming only `(SubprocessError, OSError)`. Widening them would have been *actively worse*: most of those methods return a **sentinel** on exception (`[]`, `None`, `{}`), so catching a decode fault there converts a crash into a **wrong answer** — the file argues this itself, that "a sentinel `[]` would falsely read as 'window dead → session crashed'". Setting `_ERRORS` stops the raise at the single place tmux is spawned, so no sentinel is ever reached for this cause. `_ENCODING` deliberately stays `None`: POSIX keeps the locale codec and only stops being strict (`errors=` is honored when `encoding=None`).

**Why `"replace"` and not `surrogateescape` in #383.** Both captures feed text **forward** and never back to a filesystem API — bus output goes to `_apply_stdout` and the journal; probe output goes to `sanitize.scrub_text` and into the probe document, rendered with `json.dumps(..., ensure_ascii=False)` and written as UTF-8. `surrogateescape` would seed that string with lone surrogates, which cannot be encoded to UTF-8 — converting this bug into **the exact encode-side crash #380 fixes in the other half of this PR.** `"replace"` yields U+FFFD, an ordinary character that encodes cleanly.

The `except` tuples in #383 are deliberately **not** widened. The fix is to stop raising, not to catch more; a `UnicodeDecodeError` can no longer occur at either site, so a widened tuple would be a guard nothing could ablate into failing.

## Scope note

- **`tui/app.py` `_commit_subject` — the third sighting listed in #383 — was already fixed by #390 / PR #544 and is untouched here. This PR does not claim it.** It routes through the `verify.git_bytes` chokepoint and takes **bytes**, decoding with `errors="replace"` itself (`tui/app.py:855`). Its regression test `test_commit_subject_routes_the_chokepoint_and_replaces_undecodable_bytes` was re-run green as proof.
- **Two `tmux_base` guards were already widened to `UnicodeError` by earlier unrelated work** — `0fccbef` (#435) for `list_window_ids`, `c07d906` for `version()`. Both are **kept**: they still guard a leaf that overrides `_ERRORS` back to a strict handler. But the comments justifying them with "POSIX is strict" were corrected, because that premise no longer holds after this change. Three such stale comments in total (`list_window_ids`, `version()`, and two test docstrings).
- **`tests/test_verify.py` keeps its own local copy of the `needs_strict_codec` marker** rather than importing the new shared one from `conftest.py`. De-duplicating it would drag an otherwise-untouched file into this PR. Deliberate; follow-up filed as #608.
- **Nothing else was folded in.** Named-and-declined sites are listed under *Out of scope* below.

## Visible behavior diffs

Both fixes swap a crash for degraded output. Three user-visible changes, stated plainly:

1. **POSIX tmux captures containing an undecodable byte now return `\xNN` escapes instead of raising.** A session name or window title carrying such a byte reads back **escaped**. This matches what the Windows backend already did.
2. **Plugin-hook output and `--version`/`--help` probe banners containing bytes the locale codec cannot decode now show `�` (U+FFFD)** — in journal entries, hook advisory text and probe documents — where the run previously crashed.
3. **A diagnostic breadcrumb whose payload holds a lone surrogate is now silently DROPPED** rather than crashing the run. That is inside `_append_diag_jsonl`'s documented best-effort contract, but it is a real change and reviewers should see it stated rather than inferred.

## Out of scope

Each of these was found, examined, and **declined** as a code change here; a follow-up issue is filed for each.

- **`adapters/generic.py` `_write_heartbeat` — #606.** Byte-for-byte the same shape as the fixed `_append_diag_jsonl` (`json.dumps(..., ensure_ascii=False)` under `except OSError: pass`) but **unreachable**: both call sites (`generic.py:634-641`, `opencode_http.py:1080-1087`) pass scalars only — `ts`, `remaining_s`, `stall_armed`, `stall_nudges_sent` — so the encoded string is surrogate-free by construction and a widened guard would be one nothing could ablate. Left alone deliberately. The follow-up adds a comment recording *why* it is inert, so the next codec sweep does not re-find and re-litigate it, and so a future payload gaining a path-derived string is caught.
- **The unity plugin scripts under `src/bmad_loop/data/plugins/unity/` — #607.** Same narrow guard around `text=True` captures: `unity_ready.py:113` and `:129` (guarded by `except subprocess.TimeoutExpired` alone), `unity_quiesce.py:111`, `unity_plugin.py:408`, `unity_setup.py:182`, `:324` and `:402` (no `try` at all). Shipped plugin **data**, not orchestrator core, and neither issue names them — folding them in would widen a bundled fix past the two issues it closes.
- **De-duplicating `needs_strict_codec` between `tests/conftest.py` and `tests/test_verify.py` — #608.**

## Testing

Four new tests, one per guard:

| Test | Guard |
| --- | --- |
| `tests/test_generic_tmux.py::test_resultless_stop_breadcrumb_surrogate_detail_is_swallowed` | `generic.py` encode |
| `tests/test_multiplexer.py::test_run_decodes_an_undecodable_byte_instead_of_raising` | `tmux_base.py` decode |
| `tests/test_hook_bus.py::test_real_subprocess_runner_replaces_undecodable_output` | `bus.py` decode |
| `tests/test_probe.py::test_run_capture_replaces_undecodable_banner_bytes` | `probe.py` decode |

One pre-existing test updated: `test_run_posix_default_passes_no_encoding_and_no_env` asserted `errors is None`. `encoding` and `env` stay `None`, so its name still holds — it pins the kwarg, while the new test pins the decode it actually produces.

### Ablation evidence — one line per new guard

Each ablation was run **singly**, against a `cp` backup (never `git checkout`), then restored byte-identical and re-confirmed green. Actual exception types below, not paraphrases:

- **`generic.py` encode** — guard reverted to `except OSError:`; the encode test failed **alone** with `UnicodeEncodeError: 'utf-8' codec can't encode character '\udcff' in position 69: surrogates not allowed`, raised at `src/bmad_loop/adapters/generic.py:340` on `fh.write(line + "\n")`.
- **`tmux_base.py` decode** — `_ERRORS` reverted to `None`; the decode test failed **alone** with `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 3: invalid start byte` at `subprocess.py:1099`. The traceback's own frame read `data = b'ok-\xff-tail', encoding = 'UTF-8', errors = 'strict'` — a real child's decode, not an injected raise.
- **`bus.py` decode** — `errors="replace"` deleted; `test_real_subprocess_runner_replaces_undecodable_output` failed with `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 7: invalid start byte`, raised from `subprocess.py:1099 _translate_newlines` (`encoding = 'UTF-8', errors = 'strict'`). The traceback's `Popen` repr read `returncode: 7` — making the lost-verdict consequence directly visible: the child had already exited 7 when the decode killed the call.
- **`probe.py` decode** — `errors="replace"` deleted; `test_run_capture_replaces_undecodable_banner_bytes` failed with the same `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 7: invalid start byte` — a **raised** exception, not a `None` return, confirming the `(OSError, SubprocessError)` guard does not catch it and that nothing in the test swallows the fault.

### Real children, not fakes

**Both #383 tests and the `tmux_base` decode test drive REAL child processes** (`_BINARY = sys.executable`), never a monkeypatched `subprocess.run`. This is load-bearing, not incidental: a faked seam **never executes the stdlib decode**, so such a test passes **identically with the bug restored** — the exact fake-green this repo was bitten by in #378.

### Skip guard

The three codec-conditional tests carry `conftest`'s shared `needs_strict_codec` skipif. Byte `0xff` is undecodable only under UTF-8/ASCII — every ISO-8859-x and cp125x codec maps all 256 byte values, so under such a locale the strict decode never raises and the test would go **vacuously green**. CI is always on the exercising side: the Linux legs run UTF-8, and the Windows legs set `PYTHONUTF8: '1'` (`.github/workflows/ci.yml`), so `needs_strict_codec` evaluates true on both and the tests **run** rather than skip.

The encode test is deliberately **not** codec-conditional: that fault is an encode *to* UTF-8, so a lone surrogate is unencodable on every host whatever the locale.

### Gates

Full suite green — **5486 passed, 50 skipped, 5 xfailed**. `uv run pyright` clean (0 errors, 0 warnings). `trunk fmt` clean. `trunk check --all --no-fix` clean across 256 files. bandit still passes: the `# nosec B602` on `bus.py`'s `subprocess.run` survived the edit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of non-UTF-8 output from hooks, version checks, and terminal sessions.
  * Prevented invalid diagnostic data from interrupting runs; affected breadcrumbs are safely skipped.
  * Preserved process results and error handling when command output contains undecodable characters.
  * Terminal output now displays problematic bytes using readable replacement characters or escapes.

* **Tests**
  * Added regression coverage for Unicode and undecodable subprocess output scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -163,6 +163,36 @@ breaking changes may land in a minor release.
   name. When the readable temp name cannot fit, the helpers now fall back to a short digest of it
   rather than failing — the OS decides, so there is no guess at a per-filesystem byte limit.
 
+- **A best-effort diagnostic write no longer takes the run down, and POSIX tmux captures no longer
+  raise on an undecodable byte (#380).** Both halves are one defect: an exception net that names one
+  direction of the Unicode hierarchy while the other escapes it. `_append_diag_jsonl` serializes
+  with `ensure_ascii=False` and writes through a UTF-8 handle, so a story spec whose filename holds
+  a byte the filesystem codec could not decode arrived surrogate-escaped and failed at the encode
+  with `UnicodeEncodeError` — a `UnicodeError`, so a `ValueError`, not an `OSError` — straight past
+  a guard whose stated contract is that an unwritable run dir must never break the completion loop.
+  The guard is now `(OSError, UnicodeError)` and the breadcrumb is dropped instead of crashing the
+  run. Separately and visibly: `tmux_base`'s `_ERRORS` was the strict handler on POSIX while the
+  Windows backend already set `"backslashreplace"`, so a capture holding such a byte raised out of
+  the one place tmux is spawned; it now comes back with that byte rendered as a `\xNN` escape, and a
+  session name or window title carrying one reads back escaped. Fixing the decode at its source
+  rather than at the fourteen catch sites is deliberate — most of them return a sentinel, so
+  catching there would have converted a crash into a wrong answer (`[]` reads as "window dead").
+  `_ENCODING` stays `None`, so POSIX keeps the locale codec and only stops being strict. This closes
+  the diagnostic writer and the capture decode; it does not sweep the wider `except`-tuple family.
+
+- **A plugin hook or version probe whose child emits a non-UTF-8 byte no longer crashes the run
+  (#383).** `subprocess.run(..., text=True)` decodes with the locale codec at `errors="strict"`, and
+  the `UnicodeDecodeError` that raises is neither an `OSError` nor a `SubprocessError`, so it
+  escaped both captures' guards. The declarative-hook transport is the run-crashing half: `_HookError`
+  is the bus's designed channel for "the hook could not run to completion", and the decode happens
+  after the child has exited — so a hook that actually **passed** took its exit code down with the
+  run instead of being classified. `probe.run_version_help` documents "Never raises", and a
+  `--version`/`--help` banner carrying such a byte made that false. Both now decode with
+  `errors="replace"`, so the hook's verdict survives and a fault is classified through the normal
+  channel. Visible change: those bytes now show as `�` (U+FFFD) in journal entries, hook advisory
+  text and probe documents. The `except` tuples are deliberately not widened — the fix is to stop
+  raising, and a widened tuple would guard a fault nothing could ablate into existence.
+
 ## [0.10.0] — 2026-08-14
 
 Much of this section is the `release/0.9.x` hotfix line brought forward onto `main` (#433).
--- a/src/bmad_loop/adapters/generic.py
+++ b/src/bmad_loop/adapters/generic.py
@@ -326,14 +326,19 @@ class _ResultFileMixin:
     def _append_diag_jsonl(self, task_id: str, filename: str, payload: dict) -> None:
         """Append ``payload`` as one JSON line to ``tasks/<task_id>/<filename>``.
         Pure observability, best-effort: an unwritable run dir must never break
-        the completion loop."""
+        the completion loop. ``ensure_ascii=False`` is why the guard names more
+        than OSError: it leaves a lone surrogate — what a POSIX filename holding
+        a non-UTF-8 byte becomes, surrogate-escaped — in the dumped str, which
+        then hits the UTF-8 encode inside ``fh.write`` as a UnicodeEncodeError.
+        That is a ValueError, not an OSError; ``UnicodeError`` covers it and the
+        decode direction both (#380)."""
         try:
             path = self.tasks_dir / task_id / filename
             path.parent.mkdir(parents=True, exist_ok=True)
             line = json.dumps(payload, ensure_ascii=False)
             with path.open("a", encoding="utf-8") as fh:
                 fh.write(line + "\n")
-        except OSError:
+        except (OSError, UnicodeError):
             pass
 
     def _note_resultless_stop(self, task_id: str, verdict: str, detail: str = "") -> None:
--- a/src/bmad_loop/adapters/tmux_base.py
+++ b/src/bmad_loop/adapters/tmux_base.py
@@ -59,10 +59,12 @@ class BaseTmuxBackend(TerminalMultiplexer):
     #: Output decoding for captured tmux text. ``None`` (POSIX) = locale default,
     #: byte-identical to a bare ``text=True``; a Windows leaf sets ``"utf-8"``.
     _ENCODING: str | None = None
-    #: Decode error handling to pair with :attr:`_ENCODING`. ``None`` (POSIX) =
-    #: the default strict handler; a Windows leaf sets ``"backslashreplace"`` so
-    #: a stray non-UTF-8 byte degrades visibly instead of raising mid-capture.
-    _ERRORS: str | None = None
+    #: Decode error handling to pair with :attr:`_ENCODING`. ``"backslashreplace"``
+    #: on every platform (#380): a stray byte the codec cannot decode degrades
+    #: visibly to a ``\xNN`` escape in the captured text instead of raising
+    #: mid-capture. Honored even where :attr:`_ENCODING` is None, so POSIX keeps
+    #: the locale codec and only stops being strict. A leaf may still override.
+    _ERRORS: str | None = "backslashreplace"
     #: Diagnostic from the last :meth:`version` probe (see
     #: :meth:`TerminalMultiplexer.version_error`). A class-level default so an
     #: instance that never probed answers None instead of AttributeError.
@@ -322,12 +324,14 @@ class BaseTmuxBackend(TerminalMultiplexer):
         # answer to "is it alive?" is "unknowable" -> MultiplexerError. A real dead
         # window still returns [] via the returncode != 0 path below (no exception).
         #
-        # UnicodeError is a transport failure too: _run decodes with the strict
-        # POSIX handler (see version()), so an undecodable byte in the capture
-        # raises a ValueError-family error that neither exception arm above names.
-        # It must not escape raw — prune_ctl_windows takes its post-kill verdict
-        # from this probe and only a MultiplexerError lands the candidates in
-        # `unverifiable` rather than aborting cleanup mid-receipt (#435).
+        # UnicodeError is a transport failure too. _run no longer decodes strictly
+        # on any platform (_ERRORS is backslashreplace, #380), so this arm is now
+        # defence for a leaf that overrides _ERRORS back to a strict handler: such
+        # a decode raises a ValueError-family error that neither exception arm
+        # above names. It stays because the seam-honesty contract above is what
+        # callers rely on — prune_ctl_windows takes its post-kill verdict from this
+        # probe and only a MultiplexerError lands the candidates in `unverifiable`
+        # rather than aborting cleanup mid-receipt (#435).
         try:
             probe = self._run(
                 ["list-windows", "-t", f"={session}", "-F", "#{window_id}"], check=False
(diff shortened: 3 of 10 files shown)


pbean:
@codex review

Issue #340, opened 2026-07-27
Decide whether a failed worktree snapshot should block the rollback reset


pbean:
Split out of #338 review (see [this thread](https://github.com/bmad-code-org/bmad-loop/pull/338#discussion_r3660585086)). #338 fixed the *claim*; this issue decides whether the *policy* should change.

## Current behavior

`RecoveryFlow.rollback_or_pause` runs: clear `preserve_ref` → `preserve_attempt_commits` → `preserve_attempt_worktree` → `safe_reset`.

`preserve_attempt_worktree` is best-effort by contract: on `verify.GitError` it journals `attempt-worktree-preserve-failed` and returns. `safe_reset` at the end of the arm is unconditional and outside any `try`, so it runs anyway — `git reset --hard <baseline>` plus deletion of run-created untracked files. Those are exactly the two categories the failed snapshot existed to capture.

`verify.snapshot_worktree` raises `GitError` at seven points (`read-tree`, `add -u`, `add -- <untracked>`, `write-tree`, `rev-parse HEAD^{tree}`, `commit-tree`, `update-ref`), and `_run_git` converts a `GIT_TIMEOUT_S` (120s) timeout into one. A slow `write-tree` on a large dirty tree, a full disk, or an unreadable file during `git add` all land there.

Net: the attempt's uncommitted work is destroyed, and the only record is one journal line.

## What #338 already did

`StoryTask.preserve_partial` is latched in that `except` arm, and the defer notice downgrades its claim to committed-work-only instead of offering a `merge --ff-only` that implies the whole attempt is recoverable. That makes the operator-facing message honest. It does not save the work.

## The question

Should a *worktree*-snapshot failure pause the way a *commits*-preserve failure does (`preserve_attempt_commits` calls `pause_for_manual_recovery(..., preserve_failed=True)` when `allow_pause`)?

Arguments against, which is why this was not done in #338:

- `preserve_attempt_worktree`'s docstring states the best-effort contract explicitly: "a capture failure is journaled but never blocks the … reset — the recovery ref is a safety net, not a gate."
- It collides with #161's invariant that an in-worktree rollback never pauses. A mounted unit worktree is disposable by design, and pausing there emits main-checkout recovery instructions for a tree the operator never works in.
- A re-drive is contractually pause-free (`allow_pause=not redrive`), so any new pause would need the same carve-out.

[...]

PR #613, opened 2026-08-15
fix(verify): split the value from the diagnostic in every git output probe (#442)

### LINKED ISSUE #442: `verify._git` merges stderr into stdout for every caller, so a git that warns at rc 0 corrupts every output-reading probe
`verify._git` (`verify.py:145-147`) returns `(rc, (stdout + stderr).strip())`. Every caller that reads the *text* rather than only the *returncode* therefore reads git's diagnostics as data. Git writes advisories to stderr while still exiting 0, so this is not a failure path — it is the normal path on a host whose git config the orchestrator does not control.

PR #441 (#423) fixed the two most dangerous instances — `worktree_clean` and the new `path_tracked`, both emptiness reads whose inversion is silent and load-bearing — by reading `proc.stdout` alone through `_run_git`. It deliberately did **not** widen to the rest of the family. This issue is that rest.

### Reproduction (git 2.55.0)

```console
$ git init -q r && cd r && git commit -qm init --allow-empty
$ git config core.fsyncMethod bogusvalue      # any unknown value; not an error
$ git rev-parse HEAD; echo "rc=$?"
77943fe624cd0aa1540d68f4571a18d8cf64f996
rc=0
# stderr, same invocation:
warning: ignoring unknown core.fsyncMethod value 'bogusvalue'
```

`git ls-files --others --exclude-standard` and `git rev-list HEAD..HEAD` both behave identically: empty stdout, rc 0, the warning on stderr. Other real-world producers of the same shape: a `core.fsmonitor` hook that cannot exec, a stale-index advisory, `core.hooksPath` pointing at a missing dir, and every `warning: ignoring unknown <key> value` for a config key a newer git dropped.

### The affected readers

| site | reads | corruption under a warning |
|---|---|---|
| `rev_parse_head` (`verify.py:189`) | `return out` | returns `"<sha>\nwarning: …"` — not a sha. Flows into `same_commit()` comparisons and into persisted baselines in run state, so a resume compares a warning-suffixed string against a clean one and reads "moved". |
| `untracked_files` (`verify.py:393`) | splits lines into a set | a **phantom untracked path** named `warning: …` on a pristine tree. Its contract is "what `git clean -fd` would remove", so the phantom becomes a rollback candidate. |
| `commits_above` (`verify.py:468`) | splits lines into a list | a **phantom commit sha**, handed to `preserve_commits`. Docstring says "Empty when HEAD is at or behind baseline" — under a warning it is never empty. |
| `last_commit_for` (`verify.py:207`) | `return out` | same as `rev_parse_head`; backs operator-park provenance. |
| `current_branch` (`verify.py:846`) | branch name | a branch name with a warning appended. |
| `worktree list --porcelain` (`verify.py:939`) | record parsing | an unparseable extra record. |
| `commit_paths`' `status --porcelain` leg (`verify.py:2240`) | `if not out` | reads non-empty on an unchanged path set, so it proceeds to `git commit` instead of returning `None`, and the commit fails with nothing to commit ⇒ `GitError`. Loud, not silent — listed for completeness. |

`verify.py:1116` is a second, inline instance of the same `ls-files --others` read, inside the patch-capture path: each returned rel is opened and appended to the patch, so the phantom line becomes a file that cannot be opened. That same function already reads `proc.stdout` alone for its `git diff` leg three lines above (`verify.py:1113`) — the fix shape is present in the file, just not applied consistently.

### Shape of the fix

Same as #441's: read `proc.stdout` through `_run_git` for the value, and keep the `stdout + stderr` merge **only** on the error path, where stderr is the informative half. `_git`'s merge is right for the "raise with a message" callers and wrong for the "parse the answer" callers; the split is per-caller, not a change to `_git` itself — `_git_raw` already exists as the precedent (`verify.py:150`, stdout verbatim for `-z` output).

Worth considering instead: a `_git_out` sibling returning `(rc, stdout, stderr)` so a reader cannot pick the merged form by accident, and leaving `_git` for rc-only callers. Whichever shape, all git subprocess calls must keep going through the `_run_git` chokepoint (AGENTS.md hard invariant).

### Test angle

A fixture that sets `core.fsyncMethod` to an unknown value in the sandbox repo (via `git config`, so it survives `GIT_CONFIG_NOSYSTEM`) and asserts each probe still answers correctly. This makes every one of the rows above fail before the fix, which no current test does — the suite has no host-noise dimension at all.

### Notes

- Pre-existing on `main`; not introduced by #441, which documented the family in its PR body and fixed only the two it was scoped to.
- Not a duplicate of #236 (locale pin) or #377 (decode taxonomy), though both are the same chokepoint.

pbean:
Closes #442

## Problem

`verify._git` returns `(rc, (stdout + stderr).strip())`. That merge is right for the ~45 callers
that only check `rc` and spend the text on a `raise GitError(f"…: {out}")`, where stderr is the
informative half. It is wrong for every caller that reads the text as the **answer**, because git
writes advisories to stderr while still exiting 0 — an unknown `core.fsyncMethod` value, a
`core.fsmonitor` hook that cannot exec, a stale-index advisory, `core.hooksPath` pointing at a
missing directory. Against the merged stream those are indistinguishable from data.

This is not an error path. It is the normal path on any host whose git config the orchestrator does
not control — a host-config property, and one an operator never sees reported as a failure.

Measured reproduction (git 2.55.0):

```console
$ git init -q r && cd r && git commit -qm init --allow-empty
$ git config core.fsyncMethod bmad-loop-not-a-method   # an unknown VALUE for a known KEY
$ git rev-parse HEAD; echo "rc=$?"
c2ccf69e07134d2cbdeb928cc1c0196ad0897484
rc=0
$ git rev-parse HEAD 2>&1 >/dev/null                   # …and the same invocation's stderr
warning: ignoring unknown core.fsyncMethod value 'bmad-loop-not-a-method'
```

`git ls-files --others --exclude-standard`, `git rev-list HEAD..HEAD`, `git status --porcelain`,
`git write-tree`, `git commit-tree`, `git stash create` and `git for-each-ref` all behave the same
way: rc 0, the real answer (or nothing) on stdout, the warning on stderr.

#441 (for #423) fixed the two most dangerous instances — `worktree_clean` and `path_tracked`, both
emptiness reads — inline, and deliberately did not widen to the rest of the family it documented.
This PR is that rest.

## Fix

**`src/bmad_loop/verify.py`** — one new helper plus thirteen conversions in eleven functions.

`_git_out(repo, *args, env=None) -> (rc, stdout.strip(), (stdout + stderr).strip())`. A sibling
helper rather than thirteen inline `_run_git` spawns, for two reasons: the split becomes
**structural** instead of per-caller discipline, so a future probe cannot pick the merged form by
accident; and the third element is what keeps error messages byte-identical — every converted
`raise` still carries stderr. It spawns through the same `_run_git` chokepoint (AGENTS.md hard
invariant), and `tests/test_portability_guard.py` stays green on the added sequence-form argv.

| function | site | what the text is |
| --- | --- | --- |
| `rev_parse_head` | `rev-parse HEAD` | the sha |
| `last_commit_for` | `log -n 1 --format=%H` | the sha |
| `current_branch` | `rev-parse --abbrev-ref HEAD` | the branch name |
| `untracked_files` | `ls-files --others` | one record per line |
| `commits_above` | `rev-list <baseline>..HEAD` | one sha per line |
| `_prune_refs` | `for-each-ref` | one ref per line |
| `worktree_list` | `worktree list --porcelain` | records |
| `capture_diff` | `ls-files --others` (untracked leg) | one rel per line |
| `snapshot_worktree` | `write-tree` | the tree sha |
| `snapshot_worktree` | `rev-parse <head>^{tree}` | the tree sha |
| `snapshot_worktree` | `commit-tree` | the snapshot sha |
| `safe_rollback` | `stash create` | the snapshot sha |
| `commit_paths` | `status --porcelain -- <specs>` | an emptiness test |

**Deliberately left alone.** `_git` keeps the merge for its ~45 rc-only callers, and `_git_raw`
keeps returning stdout verbatim for `-z` output, whose records can begin with a space that
`.strip()` would corrupt. `_git_env`'s remaining calls on the snapshot path are rc-only and stay
as they are.

**The one merged READ deliberately kept:** `safe_rollback`'s `"did not match" not in out`
tolerance, below the preserve restore. It inspects text on the **error** path, where git's
"pathspec … did not match" lands on stderr — stdout alone can never contain it, so the substring
would stop matching and a benign empty preserve dir would raise. That is asserted rather than
assumed, by inverse ablation: convert that one line to read the stdout half and
`test_safe_rollback_tolerates_empty_preserve_dir` fails with
`GitError: git checkout df775e44178a -- _bmad-output failed:` — note the empty diagnostic after
the colon, which is git having put all of it, "did not match" included, on stderr. A comment on
the line records this so the next sweep does not "finish the job".

## Scope note

Not in this diff:

- `_git` itself and its ~45 rc-only call sites. They are correct as they are.
- The non-git `stdout + stderr` merges elsewhere in the tree — `plugins/bus.py`,
  `verify.run_verify_commands`, the Unity plugin scripts. Different surface, out of scope for #442.
- No new issues were filed during phases 1–3 of this work; nothing was deferred out of it.

## Visible behavior diffs

On a host whose git warns at rc 0:

- `rev_parse_head`, `last_commit_for` and `current_branch` return the value instead of the value
  with a warning appended. The sha probes feed `same_commit()` comparisons and the baselines
  persisted in run state, so before this a resume graded a warning-carrying string against a clean
  one and read "moved".
- `untracked_files` stops reporting a phantom rollback candidate on a pristine tree, and
  `commits_above` stops handing `preserve_commits` a phantom sha.
- Three call paths stop raising: `PrunePreserveError` from preserve-ref retention (the phantom
  entered the ref list, landed in the `refs[keep:]` tail and was handed to `delete_branch`),
  `GitError` from `snapshot_worktree`, and `GitError` from `commit_paths` on an unchanged path set
  (where its contract returns `None`).
- `snapshot_worktree` is worth naming on its own: since #340 that recovery ref is a **gate**, not
  a safety net — a plain rollback refuses to reset past work it could not park — so a noisy host
  had no working rollback at all.
- `safe_rollback` stops attempting a restore from a non-ref after the reset. On a clean tree
  `stash create` emits empty stdout with the warning on stderr, so against the merge `snapshot`
  became a non-empty non-ref and the restore ran `git checkout 'warning: …' -- <dir>` — **after**
  `git reset --hard` had already run. Destructive first, then loud.

On a quiet git, nothing changes at all. Error messages are unchanged everywhere: every converted
`raise` still interpolates the merged text.

## Honesty callout

Two of the thirteen were **not** reproducibly corrupted by the measured warning shape, and #442's
text is wrong about both:

- `worktree_list` — #442 says the parse gains "an unparseable extra record". It does not. The parse
  keeps only lines starting with `"worktree "`, which screens the warning out. Correct by accident
  of the filter shape.
- `capture_diff`'s untracked leg — #442 says the phantom "becomes a file that cannot be opened" and
  corrupts the patch. Measured, `git diff --no-index -- /dev/null "<phantom>"` exits **1** ("could
  not access") with **0 bytes** of stdout, and rc 1 is exactly the code the loop already tolerates
  as "the files differ", so nothing is appended. Correct by accident too.

Both are converted **defensively**, so the reads stand on the split rather than on that accident,
and `worktree_list` keeps its `startswith` filter as a second, independent screen. Their tests
therefore cannot use the real warning; they use the house
`monkeypatch.setattr(verify.subprocess, "run", …)` seam with a synthetic stderr line chosen to
survive each filter, and each docstring says which axis it uses and why. Stating this because a
reviewer who re-measures will find it, and finding it unstated costs more than stating it.

## Testing

`tests/conftest.py` gains `make_git_noisy` — the suite's first host-noise dimension, a plain
function rather than a fixture per `docs/testing.md`. It sets an unknown value for a known key, so
git warns at rc 0.

Fourteen new tests. One line of ablation evidence each, from the phase commits — every ablation run
against a `cp` backup and restored byte-identical, never `git checkout`:

| test | ablation → result |
| --- | --- |
| `test_rev_parse_head_reads_stdout_alone_under_host_noise` | site back on `_git` → FAILED alone: `assert "c2ccf69e…not-a-method'" == 'c2ccf69e0713…'` |
| `test_last_commit_for_reads_stdout_alone_under_host_noise` | site back on `_git` → FAILED alone, same shape |
| `test_current_branch_reads_stdout_alone_under_host_noise` | site back on `_git` → FAILED alone: `assert "main\nwarnin…" == 'main'`. The pre-existing `test_current_branch` passed throughout — which is why the new row is needed |
| `test_rev_parse_head_failure_still_carries_git_stderr` | swap the raise's `detail` back to `out`, keeping `_git_out` → FAILED: the message trails off empty after the colon, git's whole diagnostic gone |
| `test_untracked_files_reads_stdout_alone_under_host_noise` | site back on `_git` → 1 failed, 4 passed: `assert {"warning: ig…"} == set()` |
| `test_commits_above_reads_stdout_alone_under_host_noise` | site back on `_git` → 1 failed, 4 passed: `assert ["warning: ig…"] == []` |
| `test_prune_preserve_refs_survives_host_noise` | site back on `_git` → 1 failed, 4 passed: `PrunePreserveError: … could not delete warning: ignoring unknown core.fsyncMethod value '…'` |
| `test_worktree_list_reads_stdout_alone` | site back on `_git` → 1 failed, 4 passed: "Left contains one more item: `PosixPath('/phantom')`" |
| `test_capture_diff_untracked_leg_reads_stdout_alone` | site back on `_git` → 1 failed, 4 passed: a `new file mode 100644` hunk synthesized for the already-**tracked** `src.txt` |
| `test_snapshot_worktree_parks_a_dirty_tree_under_host_noise` | `commit-tree` back on `_git_env` → FAILED alone: `update-ref … fatal: df3963370a71…warning: …: not a valid SHA1` |
| `test_snapshot_worktree_still_noops_on_a_clean_tree_under_host_noise` | `rev-parse ^{tree}` back on `_git` → FAILED alone: a preserve ref where `None` was due. (`write-tree` back on `_git_env` reddens **both** snapshot rows — a corrupt `tree` cannot equal a clean `head_tree`, so the no-op stops returning early; the overlap is recorded in the tests rather than hidden.) |
| `test_safe_rollback_restores_a_preserve_dir_under_host_noise` | `stash create` back on `_git` → FAILED with `git checkout warning: ign -- _bmad-output/… : fatal: invalid reference: warning: …` |
| `test_safe_rollback_degrades_on_a_clean_tree_under_host_noise` | same ablation → FAILED alongside it |
| `test_commit_paths_noops_on_unchanged_paths_under_host_noise` | site back on `_git` → FAILED alone: `GitError: git commit failed: … nothing to commit, working tree clean / warning: …` |

Plus the fixture's own anti-vacuity guard, `test_make_git_noisy_produces_rc_zero_stderr`, which
exists because `core.fsyncMethod` only arrived in **git 2.36**: below that the key is merely
unknown, git ignores it in silence, and every dependent row would pass with the bug fully restored.
The helper measures the premise instead of predicting it and **skips**, naming the version it saw
(`git version X does not warn at rc 0 for an unknown core.fsyncMethod value`) — same argument as
`needs_strict_codec`. Ablated both ways: delete the `git config` line and the dependent rows report
`4 skipped, 0 passed`, none of them a false green; delete the probe as well, so nothing masks the
dead premise, and it is `4 failed`. Deleting the probe **alone** reddens nothing while the host git
still warns, which is the point rather than a gap — the probe is what turns a future silent git
into skips instead of false greens.

Two further pieces of ablation hygiene, recorded because they are easy to get wrong here:

- A **full** revert of a phase-1 commit cannot grade a test that asserts on an error *message* — it
  restores a single `out` that is already the merged text, so the assertion still holds. All the
  value-read ablations above are surgical, one site at a time.
- The two synthetic-shape rows (`worktree_list`, `capture_diff`) were re-run with the **real**
  warning substituted for the injection, against the **unfixed** code: **2 passed**. That pass is
  the evidence the synthetic shape is necessary and the honesty callout above is measured rather
  than guessed.

Finally, `test_prune_preserve_dirty_refs_keep_zero_never_runs_git` is hardened in this PR because
this diff narrowed it: the `for-each-ref` listing moved to `_git_out` while the deletes still route
via `_git`, so the `_git`-only guard stopped watching every git path the `keep <= 0` early return
skips. It patches both helpers now. The discriminating ablation — early return deleted **and** no
dirty ref present, so the listing runs and nothing is deleted — fails on the hardened form
(caught at the `for-each-ref` argv) and **passes** on the old `_git`-only form. That pass is the
hole, measured rather than argued.

## Gates

`uv run pytest -q -n logical` → 5515 passed, 50 skipped, 5 xfailed · `uv run pyright` → 0 errors ·
`trunk fmt` clean · `trunk check --all --no-fix` → 256 files, no issues ·
`uv run --no-project python scripts/release.py check` → version-sync ok.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,19 @@ breaking changes may land in a minor release.
 
 ### Fixed
 
+- **A git that warns while still exiting 0 no longer corrupts the answers the orchestrator reads
+  back from it (#442).** An unknown `core.fsyncMethod` value or a `core.fsmonitor` hook that cannot
+  exec is a property of the host's git config rather than a failure, but the shared git helper handed
+  callers `stdout` and `stderr` merged — so every probe that read git's text as the ANSWER took the
+  advisory for data. Silently: a rollback candidate list grew a phantom path, a commit list grew a
+  phantom sha, and the sha and branch probes answered with the warning appended — the last of which
+  reached the baselines persisted in run state, so a resume graded a warning-carrying string against
+  a clean one and read "moved". Loudly: preserve-ref retention raised `PrunePreserveError`,
+  `commit_paths` raised where its contract reports nothing to commit, and the worktree snapshot
+  failed outright — which since #340 is a gate and not a safety net, so a noisy host had no working
+  rollback at all. Thirteen value-reading probes now take stdout alone; the ~45 call sites that only
+  check a return code are unchanged, and a failing git call still carries its stderr.
+
 - **An indented, tab-separated or empty heading in the deferred-work ledger no longer lets a `gate:`
   below it hold a story the entry never named (#516).** A canonical entry ended only at a `#` run at
   column zero with exactly one space after it, so three shapes CommonMark spells as ATX headings —
--- a/src/bmad_loop/verify.py
+++ b/src/bmad_loop/verify.py
@@ -480,10 +480,16 @@ def _path_under_any(path: str, prefixes: tuple[str, ...]) -> bool:
 def untracked_files(repo: Path) -> set[str]:
     """Untracked, non-ignored paths (repo-relative posix), mirroring what a
     plain `git clean -fd` (no -x) treats as removable. Ignored files are
-    excluded, so they are never rollback candidates."""
-    rc, out = _git(repo, "ls-files", "--others", "--exclude-standard")
+    excluded, so they are never rollback candidates.
+
+    Reads stdout ALONE (`_git_out`): `ls-files` exits 0 while still writing to
+    stderr, and against `_git`'s merged stream that chatter splits into a phantom
+    untracked path — a PRISTINE tree answers with one, and because this function's
+    contract is what `git clean -fd` would remove, the phantom is a rollback
+    candidate. Silent, on every host whose git config warns (#442)."""
+    rc, out, detail = _git_out(repo, "ls-files", "--others", "--exclude-standard")
     if rc != 0:
-        raise GitError(f"git ls-files --others failed in {repo}: {out}")
+        raise GitError(f"git ls-files --others failed in {repo}: {detail}")
     return {line.strip() for line in out.splitlines() if line.strip()}
 
 
@@ -664,10 +670,15 @@ def commits_above(repo: Path, baseline: str) -> list[str]:
     not assume a strict newest-first / HEAD-first ordering across merges or clock
     skew; callers that need the tip should read HEAD directly). Empty when HEAD is
     at or behind baseline. Raises GitError on a git failure (a bad baseline is a
-    real error, never quietly "no commits")."""
-    rc, out = _git(repo, "rev-list", f"{baseline}..HEAD")
+    real error, never quietly "no commits").
+
+    Reads stdout ALONE (`_git_out`): git exits 0 while still warning on stderr, and
+    against the merged stream that warning is a phantom sha handed to
+    :func:`preserve_commits` — "Empty when HEAD is at or behind baseline" stops
+    holding on any host whose git config warns (#442)."""
+    rc, out, detail = _git_out(repo, "rev-list", f"{baseline}..HEAD")
     if rc != 0:
-        raise GitError(f"git rev-list {baseline}..HEAD failed in {repo}: {out}")
+        raise GitError(f"git rev-list {baseline}..HEAD failed in {repo}: {detail}")
     return [line for line in out.splitlines() if line]
 
 
@@ -733,10 +744,16 @@ def _prune_refs(
     :class:`PrunePreserveError` — after attempting every tail ref — when any
     individual delete failed. One stuck ref must not wedge the retention for
     everything behind it, so deletes are per-ref best-effort and the error
-    carries both what was deleted and what was not."""
+    carries both what was deleted and what was not.
+
+    Reads stdout ALONE (`_git_out`): `for-each-ref` exits 0 while still warning on
+    stderr, and against `_git`'s merged stream that warning enters ``refs``, lands
+    in the ``refs[keep:]`` tail and is handed to ``delete`` — which fails, so
+    retention raises :class:`PrunePreserveError` on every host whose git config
+    warns (#442)."""
     if keep <= 0:
         return []
-    rc, out = _git(
+    rc, out, detail = _git_out(
         repo,
         "for-each-ref",
         # ties on committerdate (same-second rollbacks) break by ascending
@@ -750,7 +767,7 @@ def _prune_refs(
         prefix,
     )
     if rc != 0:
-        raise GitError(f"git for-each-ref {label} failed in {repo}: {out}")
+        raise GitError(f"git for-each-ref {label} failed in {repo}: {detail}")
     refs = [line.removeprefix(strip) for line in out.splitlines() if line]
     deleted: list[str] = []
     failed: list[str] = []
@@ -1178,10 +1195,17 @@ def worktree_prune(repo: Path) -> None:
 
 
 def worktree_list(repo: Path) -> list[Path]:
-    """Paths of every worktree attached to `repo` (the main checkout first)."""
-    rc, out = _git(repo, "worktree", "list", "--porcelain")
+    """Paths of every worktree attached to `repo` (the main checkout first).
+
+    Reads stdout ALONE (`_git_out`) so the record parse does not depend on no
+    stderr line ever starting with ``"worktree "``. The advisories measured for
+    #442 — an unknown `core.fsyncMethod` value and its family — do NOT start that
+    way, so the `startswith` filter screens them out and this parse was correct by
+    accident rather than by construction; the filter stays as a second, independent
+    screen."""
+    rc, out, detail = _git_out(repo, "worktree", "list", "--porcelain")
     if rc != 0:
-        raise GitError(f"git worktree list failed in {repo}: {out}")
+        raise GitError(f"git worktree list failed in {repo}: {detail}")
     paths = []
     for line in out.splitlines():
         if line.startswith("worktree "):
@@ -1356,7 +1380,14 @@ def capture_diff(repo: Path, baseline: str, *, max_file_bytes: int | None = None
     for forensics. Returns "" when there is nothing to capture.
 
     Unlike `_git`, the tracked diff is read from stdout alone and left verbatim
-    (no strip, no stderr merge) so the patch stays applyable.
+    (no strip, no stderr merge) so the patch stays applyable, as is the
+    `--no-index` spawn below it. The untracked leg now matches those two
+    (`_git_out`, #442): its `ls-files` exits 0 while still
+    warning on stderr, so against the merged stream the warning splits off as a
+    phantom rel. Measured, that phantom is inert here — `diff --no-index` cannot
+    access it and exits 1, exactly the code the loop below already tolerates as
+    "the files differ", with empty stdout — so this leg is converted for the same
+    reason its two neighbours read stdout alone, not on a demonstrated corruption.
 
     max_file_bytes caps the size of each *untracked* file included: a file larger
     than the cap is skipped and replaced with a one-line marker naming it and its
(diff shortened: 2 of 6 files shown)


pbean:
@codex review

pbean:
@coderabbitai review

PR #239, opened 2026-07-21
feat(adapters): coding-CLI adapter registry (Seam A) — out-of-tree adapters with zero core edits

pbean:
## What & why

The **transport axis** has long been extensible out-of-tree (`register_multiplexer` + the `bmad_loop.mux_backends` entry-point group). The **coding-CLI adapter axis** had no equivalent: a CLI needing a new adapter *class* forced a name-branch in the run bootstrap (`if profile.hookless: …`) and a hardcoded valid-kinds set. This PR lands the missing registry so a new adapter family — and its selecting profile — plugs in with **zero core `.py` edits**, exactly like a transport backend.

This is **Session 1 (Seam A)** of the PR #226 review plan. No Hermes-specific code, no hook-scope changes, no relay changes.

> **Rebased as a redesign.** Main absorbed none of Seam A, but the integration surface moved ~640 commits: the dispatch now lives in `runsetup.make_adapters` (not `cli._make_adapters`), the config digest gained the #461 `profiles=` pin, and `_parse_profile` grew invariants. The branch was re-cut onto main rather than merged — see *Rebase deltas* below for what changed versus the original cut.

## Design

Mirrors `adapters/multiplexer.py`, with two **deliberate** asymmetries (documented in the module):

- **No `lru_cache`/`cache_clear`** — adapters are built *per run* in `make_adapters` (its own `by_cfg` cache), so there is nothing to cache.
- **No `configure_*`/`matches(platform)`/platform defaults** — an adapter kind is selected by the `profile.adapter` field alone, not a policy knob or `sys.platform` predicate.

Two dataclasses:

- `AdapterKind(name, needs_mux, load)` — `load` is a lazy thunk, so registration/validation/listing never import a heavy adapter module (nor an optional dep like `httpx`).
- `AdapterBuilder(plain, dev, construct_error)` — the plain class, the `_DevSynthesisMixin`-composed dev class, and the family's construction-failure type(s) (`()` = none; `(OpencodeServerError,)` for HTTP).

## Changes

- **New `adapters/registry.py`**: `register_adapter` (first-wins so builtins can't be shadowed), builtin loader (`generic` needs_mux=True, `opencode-http` needs_mux=False), the `bmad_loop.adapters` entry-point scan degrading into `_EXTERNAL_ERRORS` (never raises), `get_adapter_kind` (fail loud, naming known kinds), `known_adapter_kinds`, `external_adapter_errors`, `detect_adapters`.
[...]

PR #625, opened 2026-08-16
fix(adapters): registry residuals — builtin shadowing, collapsed load errors, adapter signature mismatch (#565, #566, #569)

### LINKED ISSUE #565: register_multiplexer: an external can shadow a bundled backend when imported before the first mux resolution
Split out of the PR #239 review round (the adapter registry got the same fix there; this is the transport-axis twin, in code #239 does not touch).

## The shape

`adapters/multiplexer.py`:

- `_load_builtin_backends()` is lazy — it runs from `get_multiplexer()`, not at module import.
- `register_multiplexer(name, matches, factory)` appends straight to `_BACKENDS` without seeding the builtins first.
- Selection breaks ties on registration order, so the first registration of a name wins.

So any out-of-tree module that calls `register_multiplexer("tmux", ...)` at import time, and is imported before the first `get_multiplexer()`, lands ahead of the bundled tmux backend and is selected in its place.

## Why it is reachable

A plugin `[python]` module is loaded via `importlib.util.spec_from_file_location` (`plugins/registry.py`), which runs arbitrary module-level code in-process — including a `register_multiplexer` call — with no ordering relationship to the first mux resolution. That trigger predates #239.

## The fix

Same one applied to the adapter registry in #239 (`adapters/registry.py`):

- set `_BUILTINS_LOADED = True` **before** the registration loop in `_load_builtin_backends`, so the loop can re-enter safely;
- call `_load_builtin_backends()` at the top of `register_multiplexer`.

That makes builtins-first an invariant of the registration function rather than a property of which entry point happened to run first.

## Test note

The existing builtins-first test cannot catch this: it arms only the backend entry-point scan, which `get_multiplexer` reaches *after* loading builtins. The regression needs a test that registers through some other import trigger before any mux resolution — the adapter-side analogue is `test_builtins_win_over_an_external_imported_by_the_profile_scan`.

### LINKED ISSUE #566: Same-named broken adapter/profile distributions collapse to one recorded error
Two broken adapter/profile distributions that publish the **same entry-point name** collapse to a single recorded error, so `bmad-loop adapters` and `bmad-loop validate` report one failure where there are two.

### Where

Both external-load error maps key on the entry-point name alone:

- `src/bmad_loop/adapters/registry.py:244` — `_EXTERNAL_ERRORS[ep.name] = f"{type(exc).__name__}: {exc}"`
- `src/bmad_loop/adapters/profile.py:519` — `_PROFILE_LOAD_ERRORS[ep.name] = f"{type(exc).__name__}: {exc}"`

### Why it bites

`importlib.metadata.entry_points(group=...)` does **not** deduplicate across distributions — two installed packages may both publish `bmad_loop.adapters = acme`. This was proved directly while reviewing #239 (real `*.dist-info` on both 3.11 and 3.13): a same-name tie is resolved by `sys.path` order, and both entry points are yielded by the scan.

So when two same-named distributions are *both* broken, the scan writes `_EXTERNAL_ERRORS["acme"]` twice and the second write wins. The operator sees one `warning: external adapter 'acme' failed to load: ...` and fixes it, only to hit the second package's failure on the next run with no indication it was ever there. The recorded-degrade contract — a failure is never a crash, but it *is* always surfaced — is silently under-delivered.

Both accessors are affected identically: `external_adapter_errors()` (registry) and `external_profile_errors()` (profile).

### Why this was not folded into #239

The entry-point name is not just a dict key — it is the value of `entry_point` in the validate finding `detail`:

```python
report.warn(
    "adapter.external",
    f"external adapter '{ep_name}' failed to load: {reason}",
    {"entry_point": ep_name, "error": reason},
)
```

(`cli.py`, for both `adapter.external` and `adapter.external-profile`), and it appears verbatim in the human-readable warning. `detail` is part of the `--json` contract in `machine.py`, so changing the key's shape is a compatibility call that deserves its own change rather than riding along on the registry seam.

### Candidate shapes

1. Key on `(dist, name)` and render as `dist:name` — most informative, biggest `--json` impact.
2. Append reasons instead of overwriting — keeps the key stable, so no contract change; the `detail` string just carries both reasons.
3. Add the distribution to the human message only, leaving `detail` untouched — smallest change, no contract impact, but `--json` consumers still see one row.

(2) and (3) are contract-preserving; (1) is not. Worth deciding explicitly.

### Acceptance

Two same-named entry points from *different* distributions, both failing to load ⇒ both reasons reach `external_adapter_errors()` / `external_profile_errors()` and both surface through `bmad-loop adapters` and `bmad-loop validate`.

**Ablation:** restore the single-key write and confirm the new test fails — a test that merely asserts "two reasons present" can pass for the wrong reason if the fixture accidentally uses distinct names.

### Related

- #239 (adapter registry seam) — this code is new there (`registry.py` is added whole; the `_PROFILE_LOAD_ERRORS` writes are new). Found during its review and deliberately deferred for the contract reason above, not because it predates the PR.
- #565 — `register_multiplexer` has the same lazy-builtins/first-wins shape; that one *is* reachable pre-#239 via a plugin `[python]` module. Distinct bug, already filed.

### LINKED ISSUE #569: make_adapters surfaces an out-of-tree adapter's signature mismatch as a bare traceback, after the run dir exists
Found while triaging a codex finding on PR #568. The compatibility half of that finding does not apply (see below); this is the half that outlives it.

## The gap

`runsetup.make_adapters` builds every adapter class with:

```python
by_cfg[key] = cls(**build_kwargs)
except builder.construct_error as e:
    raise SystemExit(f"error: {e}") from e
```

`construct_error` is family-declared and covers what a family says its `__init__` can raise — `()` for generic, `(OpencodeServerError,)` for opencode. A **signature** mismatch is not that: an out-of-tree class whose `__init__` does not accept a keyword core passes raises `TypeError`, which no family declares, so it escapes.

Two lines up, the `kind.load()` arm already states why that matters:

> By here `compose_run` has already written the run state and pid, so an escaping ImportError strands a run directory behind a traceback.

The same is true here, and one step later. The `ImportError` arm was given a clean `error:` line naming the profile and the kind; the construct call has no equivalent for the mismatch case.

## Why it is not urgent

The out-of-tree adapter-**class** capability (#226, PR #239) has not shipped in a tagged release — the last tag is `v0.9.1` and the entry is under `## [Unreleased]`. The authoring guide states there is no out-of-tree adapter-class package to copy yet, so today the only classes reaching this line are the bundled ones, which are edited in the same commit as any new keyword.

PR #568 documents the obligation on the author side (accept `**kwargs`; the bootstrap keyword set grows), which is the cheap half and the one that belongs before first release. This issue is the core-side half: even a correct contract gets violated, and the failure should read like the rest of the bootstrap rather than like a crash.

## Shape

Catch `TypeError` around the construct call and re-raise as `SystemExit` naming the profile, the kind, and the offending keyword — matching the `ImportError` arm above it. `TypeError` only, on the same rule that arm states: a declared failure gets a clean line, anything else is a bug in that package and must surface as itself.

Care needed on one point: a `TypeError` raised from *inside* a working `__init__` is a bug in that package and must not be relabelled as a signature mismatch. Distinguishing the two means inspecting the exception rather than merely its type — worth pinning with a test that a `TypeError` thrown from the constructor body still surfaces as itself.

Related: #226, PR #239, PR #568.

pbean:
Closes #565, Closes #566, Closes #569

## Problem

Three residuals of the same seam: the adapter registry PR #239 introduced, its transport-axis twin
in `adapters/multiplexer.py`, and the run bootstrap that consumes both. Each is a place where the
registry/entry-point machinery answers a question — *which backend owns this name*, *what went wrong
loading this package*, *why did construction fail* — and answers it wrongly or unusably.

- **#565 — a bundled transport can be shadowed.** `_BACKENDS` is an ordered list and every consumer
  takes the first entry under a name, but the bundled backends were seeded only from `_select` /
  `get_multiplexer`, so a `register_multiplexer("tmux", ...)` running before the first resolution
  landed ahead of the bundled entry and was selected in its place
  (`src/bmad_loop/adapters/multiplexer.py:430`). The trigger is a plugin's `[python]` module, which
  `plugins/registry.py` exec's in-process with no ordering relationship to mux resolution.

- **#566 — same-named load failures collapse.** `importlib.metadata.entry_points(group=...)` does
  not deduplicate across distributions, and all three external-load scans recorded failures by plain
  assignment into a name-keyed dict, so a second failing same-named distribution overwrote the first
  and its reason was never surfaced (`src/bmad_loop/adapters/registry.py:247`,
  `src/bmad_loop/adapters/profile.py:528`, `src/bmad_loop/adapters/multiplexer.py:547`).

- **#569 — a signature mismatch escapes as a traceback.** `make_adapters` converts only what an
  adapter family *declares* its `__init__` can raise, and a signature mismatch is refused by the
  interpreter rather than by the family, so an out-of-tree class rejecting a bootstrap keyword
  raised a `TypeError` no family declares and it escaped both arms
  (`src/bmad_loop/runsetup.py:551`).

## Fix

- **`src/bmad_loop/adapters/entrypoints.py`** (new leaf) — `record_load_error`: appends a reason
  under the stable `ep.name` key, labelling each with its distribution when one is resolvable, and
  `"; "`-joins on repeat.
- **`src/bmad_loop/adapters/multiplexer.py`** — `register_multiplexer` calls
  `_load_builtin_backends()` first; `_BUILTINS_LOADED = True` moves to between the two imports and
  the registrations (below the imports so a transient import failure leaves the seeding retryable,
  above the registrations because they now re-enter through `register_multiplexer`); the mux scan
  routes through `record_load_error` and is sorted by `(name, distribution)`.
- **`src/bmad_loop/adapters/registry.py`**, **`src/bmad_loop/adapters/profile.py`** — the adapter and
  profile scans route their failure writes through `record_load_error`.
- **`src/bmad_loop/runsetup.py`** — a second `except TypeError` arm below the existing
  `except builder.construct_error`, raising `SystemExit` with an `error:` line naming the profile,
  the kind and the rejected keyword. It discriminates on traceback depth: argument binding fails
  before any `__init__` frame is pushed, so a mismatch carries this frame alone (`tb_next is None`)
  while a `TypeError` raised inside a working `__init__` carries that frame too and is re-raised.
  `_unwind_composition`'s "five sites" enumeration is corrected to six, here and at its three other
  repetitions.

## Scope note

Two things are folded in beyond the filed issue text:

1. **The third error-collapse site.** #566's body names `registry.py` and `profile.py` only; the mux
   scan in `adapters/multiplexer.py` has the identical shape and is fixed with them. Leaving it
   would have shipped the fix to two of three scans.

2. **Sorting the mux scan by `(name, distribution)`.** The append order *is* the #566 fix's output —
   unsorted, two accumulated reasons render in `sys.path` order, so the same two installed packages
   read differently on two hosts and the test proving the accumulation is non-deterministic by
   construction. Commit `90a7ca9` added that sort to the adapter and profile scans and did not touch
   `multiplexer.py`; this brings the third scan into line.

Note that the two registries deliberately still differ: the mux registry is an ordered list where
registration order breaks selection ties, while the adapter registry is a dict using `setdefault`.
Only the builtins-first invariant is made common.

## Visible behavior diffs

1. An external backend registered under a bundled name (`tmux`, `psmux`) before the first
   resolution is **no longer selected** in the bundled backend's place (#565). It is still
   registered, behind the bundled entry.
2. Every recorded external-load failure **names its distribution** when one is resolvable — in
   `bmad-loop adapters`, `bmad-loop mux`, `validate`'s text output and `validate --json`
   (`findings[].message` and `detail.error`) (#566). `detail`'s key set and value types are
   unchanged, so `VALIDATE_SCHEMA_VERSION` stays 1.
3. **A limit, not a fix:** two failing same-named distributions still produce **one** finding and
   one warning line, whose text now carries both reasons `"; "`-joined. The row count does not
   double (#566). Keying on `(distribution, name)` would be more informative and is not
   contract-preserving.
4. Mux external entry points are visited in `(name, distribution)` order; a same-named tie no longer
   resolves by `sys.path` (#566).
5. A new `error:` line at run launch when an adapter class rejects a bootstrap keyword, where a
   traceback used to print (#569). The process exit code is unchanged at 1.

## Testing

Every new guard was ablated singly, restoring from a `cp` backup between runs (md5-verified), never
via `git checkout`. One line of evidence each:

**#565 — `tests/test_backend_registry.py`**

- `test_builtins_win_over_an_external_registered_before_any_resolution` — removing only the
  `_load_builtin_backends()` call from `register_multiplexer`: FAILED on
  `isinstance(get_multiplexer(), TmuxMultiplexer)`, the sentinel external having been selected.
  **Under that same ablation the pre-existing `test_externals_load_after_builtins` PASSED** — that
  asymmetry is the finding, and the reason the old test is insufficient rather than wrong (it is
  kept unchanged).
- `test_builtin_seeding_is_reentrant_and_registers_each_builtin_once` — moving
  `_BUILTINS_LOADED = True` back below the registrations: FAILED with `RecursionError` at
  `multiplexer.py:453`.
- `test_a_failed_builtin_import_leaves_the_seeding_retryable` — moving that flag to the very top,
  above the imports: FAILED on `assert _BUILTINS_LOADED is False`, isolating exactly the
  retryability property.

**#566 — `tests/test_adapter_registry.py`, `tests/test_profile.py`, `tests/test_external_backends.py`**

- `test_same_named_broken_distributions_both_record_a_reason` (one per scan) — reverting each site
  to the single-key assignment reddened **only that scan's** test, the other two staying green. That
  per-site independence is what proves the fix reached all three.
- `test_a_lone_failure_records_exactly_one_reason` — making `record_load_error` append
  unconditionally: FAILED on the leading `"; "`.
- `test_same_named_entry_points_are_visited_in_distribution_order` — reducing the mux sort key to
  `key=lambda e: e.name`: **only the `zeta-first` parameter reddened**, `alpha-discovered-first`
  stayed green, matching the asymmetry the adapter-side precedent already records.

**#569 — `tests/test_adapter_registry.py`** (wiring and predicate ablated on separate axes; an
ablation removing the whole arm proves nothing about the discriminator inside it)

- Deleting the whole `except TypeError` arm: `..._signature_mismatch_becomes_systemexit` FAILED with
  the raw `TypeError` propagating from `runsetup.py:551` — the pre-fix behavior. The three others
  PASSED, which is the honest way to show they do not measure the wiring.
- Deleting the two discriminator lines so every `TypeError` becomes a `SystemExit`:
  `..._typeerror_from_the_constructor_body_is_not_relabelled` and
  `..._typeerror_raised_deeper_in_the_constructor_is_not_relabelled` both FAILED on the mislabel the
  check exists to prevent.
- Swapping the two `except` arms: `..._declared_typeerror_keeps_the_construct_error_line` FAILED
  alone, the declared `TypeError` escaping instead of being converted.

Full suite, `uv run pyright` and `trunk check --all` are clean on this branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multiplexer backend discovery, ensuring bundled backends remain available and external backends load consistently.
  * Enhanced plugin and profile loading diagnostics by preserving multiple failures and identifying their distributions.
  * Made discovery order deterministic for more predictable behavior.
  * Improved adapter startup errors by clearly identifying unsupported configuration keywords while preserving genuine internal errors.

* **Documentation**
  * Updated changelog and diagnostic guidance to reflect improved error reporting and adapter startup handling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -272,6 +272,46 @@ breaking changes may land in a minor release.
   _implementing_ rate limiting prints all day — follow-up: #610. The four unseeded profiles
   (`codex`, `gemini`, `copilot`, `antigravity`) are unchanged and still ship none.
 
+- **An out-of-tree module that claims a bundled transport's name is no longer driven in the bundled
+  backend's place (#565).** Backend selection breaks ties on registration order, and the bundled
+  `tmux` / `psmux` entries were seeded only on the first multiplexer resolution — so any
+  registration that landed before it owned that name, and the whole process silently ran on a
+  third-party transport. The trigger is reachable rather than theoretical: a plugin's `[python]`
+  module is executed in-process by the plugin registry, which has no ordering relationship to when
+  the mux is first resolved. Registering a backend now seeds the bundled ones first, making
+  builtins-first an invariant of the registration itself rather than a property of which import ran
+  first — the same fix the adapter registry already carries. A shadowing external is still
+  registered, just behind the bundled entry.
+
+- **Two broken distributions that publish the same entry-point name no longer collapse to a single
+  recorded error (#566).** `importlib.metadata` does not deduplicate entry points across
+  distributions, and all three external-load scans — adapters, profiles and mux backends — recorded
+  failures by assignment into a name-keyed map, so the second failing package overwrote the first.
+  You fixed the one failure you were shown, then met the other on the next run with nothing saying
+  it had ever been there. Both reasons now surface, and every recorded external-load failure names
+  its **distribution** when one is resolvable — in `bmad-loop adapters`, `bmad-loop mux`,
+  `validate`'s human output and `validate --json` — because the entry-point name is not the
+  distribution you uninstall.
+
+  The limit, plainly: two failing same-named distributions still produce **one** finding and one
+  warning line, whose text now carries both reasons `"; "`-joined. The row count does not double.
+  That is what keeps the change contract-preserving — `detail`'s key set and value types are
+  unchanged, so `VALIDATE_SCHEMA_VERSION` stays 1.
+
+  **A deliberate scope addition:** the multiplexer's entry-point scan is now ordered by (name,
+  distribution), as the adapter and profile scans already were, so which of two same-named
+  distributions is recorded first is a fact about the packages rather than about `sys.path`.
+
+- **A run launch whose adapter class rejects a bootstrap keyword now aborts on a clean `error:`
+  line instead of a bare traceback (#569).** The construct call converted only the failures an
+  adapter family declares, and a signature mismatch is refused by the interpreter rather than by
+  the family — so an out-of-tree class whose `__init__` does not accept a keyword the bootstrap
+  passes escaped uncaught. The abort now names the profile, the adapter kind and the rejected
+  keyword; the exit code is unchanged at 1. Deliberately unchanged: a `TypeError` raised from
+  inside a working `__init__` still surfaces as itself, because that is a bug in that package
+  rather than a declared failure, and relabelling it as a signature mismatch would bury it. This is
+  message quality only — an error escaping here already unwound the composition it had started.
+
 ## [0.10.0] — 2026-08-14
 
 Much of this section is the `release/0.9.x` hotfix line brought forward onto `main` (#433).
--- a/src/bmad_loop/adapters/entrypoints.py
+++ b/src/bmad_loop/adapters/entrypoints.py
@@ -1,0 +1,63 @@
+"""Shared recording for the three ``bmad_loop.*`` entry-point scans.
+
+The adapter registry (:mod:`~.registry`), the profile loader (:mod:`~.profile`)
+and the multiplexer registry (:mod:`~.multiplexer`) each scan their own
+entry-point group, and each degrades a broken third-party distribution to a
+*recorded* reason rather than a crash. This module owns the one thing all three
+recordings must agree on: how a failure becomes an entry in that map.
+
+A leaf on purpose — standard library only, and **no import of a sibling
+adapters module**. Those three do not import each other today, and the package's
+builtins load lazily to keep that so; an edge from here into any of them would
+put a cycle one refactor away.
+
+Why the map stays keyed on the entry-point NAME. The key is what reaches
+``detail["entry_point"]`` in ``bmad-loop validate --json``. That document is
+schema-versioned and evolves additively (see ``documents.validate_document``,
+which contracts ``check`` as the matchable identity and states outright that
+``message``/``detail`` are for humans) — so widening the key to
+``(distribution, name)`` would change a value consumers can already see, while
+widening only the reason TEXT does not.
+
+Why reasons accumulate instead of overwriting.
+``importlib.metadata.entry_points(group=...)`` does not deduplicate across
+distributions: two installed packages may both advertise ``acme`` in one group,
+and the scan yields both. A plain assignment let the second failure silently
+overwrite the first, so an operator fixed one package and met the other on the
+next run with no sign it had ever been there. Reasons are joined with ``"; "``
+— reasons already contain colons, so a colon separator would be unreadable.
+
+Why the distribution labels each reason. The entry-point name is not the name
+you ``pip uninstall``, and two packages failing the same way otherwise render as
+the same sentence twice, with nothing to tell the operator there are two.
+
+The honest limit: a same-named collision still records ONE row, whose text now
+carries both reasons. The row count does not double — that is the price of
+leaving the key (and therefore ``--json``) untouched, and it still puts every
+reason in front of the operator.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+
+def record_load_error(errors: dict[str, str], ep: Any, exc: BaseException) -> None:
+    """Record ``exc`` against ``ep``'s name in ``errors``, appending to whatever a
+    same-named entry point already recorded.
+
+    ``ep`` is annotated ``Any`` deliberately: the callers pass a real
+    ``importlib.metadata.EntryPoint`` but every test passes a hand-rolled double,
+    so naming ``EntryPoint`` here would claim a contract this function does not
+    require (it touches ``.name`` and, defensively, ``.dist.name``).
+
+    The doubled ``getattr`` tolerates a double with no ``dist`` attribute at all
+    as well as a real entry point whose ``dist`` is ``None``; the truthiness
+    check then treats an empty distribution name as absent, the same
+    normalization the scans' ``or ""`` sort keys apply."""
+    dist = getattr(getattr(ep, "dist", None), "name", None)
+    reason = f"{type(exc).__name__}: {exc}"
+    if dist:
+        reason = f"{dist}: {reason}"
+    prior = errors.get(ep.name)
+    errors[ep.name] = f"{prior}; {reason}" if prior else reason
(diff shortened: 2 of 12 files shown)


pbean:
@codex review

Issue #366, opened 2026-07-29
sprintstatus._set_mapping_value fabricates an inline comment from a quoted value


pbean:
Split out of #357 part 2 (PR #365), which fixed the same defect class in `frontmatter._replace_value`.

`sprintstatus._set_mapping_value` preserves a trailing inline comment on the line it rewrites, and finds it with a wide value pattern:

```python
rf"^(?P<indent>\s*){re.escape(key)}:(?P<gap>[ \t]+)"
r"(?P<val>\S(?:.*?\S)?)(?P<rest>[ \t]+#.*)?$"
```

`val` may contain a `#`, so the split between value and comment is guessed from the LAST ` #` on the line rather than from where the scalar actually ends. Against a quoted value carrying a `#`:

```
  status: "a # b"   ->   val = '"a'   rest = ' # b"'
```

Rewriting that story to `done` writes `status: done # b"` — the trailing `b"` is promoted from scalar text into a comment the board never had, and the quoted value is truncated.

## Why the sibling fix does not carry over

PR #365 fixed this in `frontmatter._replace_value` by gating the carry on a conservative token class (`[A-Za-z0-9._-]`) instead of a wide one.

Note that a verified-edit oracle does **not** catch this — `yaml.safe_load` strips comments before comparing, so the corrupted line re-parses as a perfectly clean `status: done`. That was demonstrated by ablation on #365: widening `val` there made the write succeed and pass all three of its verification gates with the junk comment attached. `_set_mapping_value` has no oracle at all, so nothing stands between the guess and the file.

## Severity

Lower than the spec-writer case, which is why it is filed rather than folded into #365:

- `sprint-status.yaml` is orchestrator-owned (`sprintstatus.advance()` is the sole write path per AGENTS.md) and the values it writes are status tokens, epic keys and timestamps — none of which is a quoted string containing a `#`.
- Reaching it needs a hand-edited board with a quoted `#` in a story or epic value.

So it is a latent correctness defect on a hand-edit path, not something the loop trips on its own.

## Note for whoever picks this up

The asymmetry between the two patterns is documented in a comment above `frontmatter._VALUE_COMMENT_RE`, which points here and says explicitly not to unify them by widening the spec-side one. If this is fixed by narrowing the board-side pattern, that comment should be updated to match.

[...]

PR #626, opened 2026-08-16
fix(sprintstatus): preserve board line endings

### LINKED ISSUE #576: sprintstatus.advance relays every line ending in the board it edits
`sprintstatus.advance` is the last writer on the `read_text` → `write_text` pair that `bb75652` fixed everywhere else. Both sides translate line endings, so a write contracted to move one value relays **every line ending in the board**:

```python
text = path.read_text(encoding="utf-8")   # universal newlines: CRLF -> LF on the way in
lines = text.splitlines(keepends=True)
...
path.write_text("".join(lines), encoding="utf-8")   # newline=None: LF -> os.linesep on the way out
```

Reproduced against `main` (`a0d6467`), Linux:

```python
crlf = "last_updated: 01-06-2026 10:00\r\ndevelopment_status:\r\n  epic-3: backlog\r\n  3-1-x: backlog\r\n"
p.write_bytes(crlf.encode())
sprintstatus.advance(p, "3-1-x", "in-progress")
p.read_bytes()
# b'last_updated: 01-06-2026 10:00\ndevelopment_status:\n  epic-3: in-progress\n  3-1-x: in-progress\n'
```

Every `\r\n` is gone — from a call asked to move one story and lift one epic. The same run on Windows does it in the other direction (`os.linesep` is `\r\n` there, so an all-LF board comes back all-CRLF), and a mixed board is normalized to the platform ending on either OS.

## Why it matters

The board is orchestrator-owned but repo-resident and usually **tracked**, so the relay lands in the story commit: a whole-file diff on top of the one-line change the run actually made, on the first `advance` of every run against a CRLF checkout (`core.autocrlf=true` on Windows, or a `.gitattributes` that pins `*.yaml eol=crlf`). It also crosses a seam — a Windows contributor's board and a Linux CI run rewrite each other's endings on alternate runs.

No corruption of *content*: the values land correctly, and `load` re-parses fine either way. This is the sprint-board analogue of what `bb75652` fixed for the four spec frontmatter writers, filed while working #366 on the same function.

## Shape of the fix

`bb75652`'s: read bytes and decode, write bytes. `_set_mapping_value` already carries the edited line's own terminator (`splitlines(keepends=True)` + a `\n`-only `rstrip`), so the writer half needs no change — but that `rstrip("\n")` currently leaves a `\r` in the matched line, which today's pattern refuses (`val` must end at `\S`), so a byte-read alone would turn the relay into a silent no-op on a CRLF board. Both halves have to move together, and both directions want a test that asserts the board's **full bytes** (a `yaml.safe_load` oracle cannot see line endings at all — the same blindness that makes #366 invisible to a re-parse).

Out of scope for the #366 PR, which does not touch either the read or the write.

pbean:
## Problem

A one-story sprint-board edit passed every line ending through universal-newline translation. Parsed YAML content and the requested story, conditional parent-epic, and optional timestamp changes remained correct, but tracked boards could show whole-file newline-only diffs across platforms.

## Fix

Read raw bytes and decode them as UTF-8, carry each line's exact terminator through `_set_mapping_value`, then encode and publish the result with the atomic bytes helper. The established atomic-write behavior and default symlink following remain intact.

## Scope note

This adds no scope beyond #576: no lock or concurrency change, no line-ending normalization policy, no `load()` change, and no public API, config, or exit-code change. The merged #366 quote/comment behavior remains unchanged.

## Visible behavior diffs

Valid UTF-8 boards with CRLF, LF, bare-CR, or mixed endings retain their authored per-line terminators. Only the existing story, conditional parent-epic, and optional `last_updated` values change.

## Testing

- Focused current-head rows: T1-T5 plus the symlink-following contract passed, 6/6, after every restored mutation.
- Full suite: `5695 passed, 50 skipped` with `uv run pytest -q -n logical --durations=15`.
- Pyright: `0 errors, 0 warnings, 0 informations`.
- Trunk formatting: no issues in 3 modified files.
- Whole-repo Trunk lint: no issues in 257 files.
- Release check: version-sync and changelog checks passed.
- `git diff --check`: passed.
- A1 (`read_text`): T1/T3 failed 2/2 on byte equality because authored CRLF and mixed terminators became LF.
- A2 (LF-only matcher strip): T1 failed 1/1 because `advance()` returned `backlog` instead of editing the CRLF line.
- A3 (flat LF emit): T1/T3 failed 2/2 on edited-line byte equality.
- A4 (`atomic_write_text` with a string): T4 failed 1/1 because the captured payload was `str`, not `bytes`.

Closes #576

## Summary by CodeRabbit

- **Bug Fixes**
  - Sprint board updates now preserve existing line endings, including CRLF, LF, bare CR, and mixed formats.
  - Updating story status, conditional epic values, and timestamps no longer reformats unrelated content.
  - Board files continue to be written safely while retaining their original UTF-8 formatting.

- **Documentation**
  - Added an Unreleased changelog entry describing line-ending preservation during sprint board updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,13 @@ breaking changes may land in a minor release.
 
 ### Fixed
 
+- **Advancing a sprint board now preserves every authored line ending (#576).** A valid UTF-8
+  board keeps each CRLF, LF, bare-CR or mixed per-line terminator while the requested story,
+  conditional parent-epic and optional `last_updated` values still change. Previously the
+  read/write relay could translate the whole board to the platform line ending, producing
+  newline-only diff noise outside those intended value edits; parsed YAML content was not
+  corrupted.
+
 - **A `sweep --migrate` rewrite that drops a `gate:` token a pre-existing entry declared is now
   refused, instead of un-gating that story silently (#519).** Validation held the rewrite to each
   entry's id and status only, so losing the token passed, got committed, and the next dispatch ran
--- a/src/bmad_loop/sprintstatus.py
+++ b/src/bmad_loop/sprintstatus.py
@@ -15,7 +15,7 @@ from pathlib import Path
 
 import yaml
 
-from .platform_util import atomic_write_text
+from .platform_util import atomic_write_bytes
 
 EPIC_RE = re.compile(r"^epic-(\d+)$")
 RETRO_RE = re.compile(r"^epic-(\d+)-retrospective$")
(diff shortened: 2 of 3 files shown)


pbean:
@codex review

PR #634, opened 2026-08-17
Bound silent dev and review sessions with stall recovery

### LINKED ISSUE #411: opencode-http: parallel review subagents are interrupted and stall bmad-loop until manual user input
## Environment

- **bmad-loop version:** v0.9.0 (latest release as of 2026-07-31)
- **Adapter:** `opencode-http` (profile `opencode`, model `zai-coding-plan/glm-5.2` for dev and review)
- **OS:** Windows 11, bmad-loop TUI running under PowerShell with the experimental `psmux` backend
- **OpenCode:** 1.18.x (Desktop + local server via `opencode serve`), agent Sisyphus-Ultraworker with GLM-5.2

## Summary

When `bmad-dev-auto` executes step-04 review via `opencode-http` and launches **two review subagents in parallel** (Blind Hunter + Edge Case Hunter), both subagents are interrupted and the bmad-loop run stalls until I manually interact with OpenCode Desktop (e.g. sending `continue` or answering a follow-up question). From bmad-loop's perspective the story sits in `dev-running` with 0 tokens and no journal progress; OpenCode Desktop shows the subagents as paused/interrupted.

The agent itself acknowledges the interruption explicitly:

> "No — both review subagents were interrupted. Let me re-launch them and complete the review pass."

Without manual user input, the review appears to never complete, the subagents' work is lost and must be re-launched.

## Reproduction steps (real run)

1. Start bmad-loop on a sprint with story `6-5-localize-history-settings-help-diagnostic-screens`.
2. Dev leg finishes and routes to a fresh step-04 review (status `done`, `followup_review_recommended: false`).
3. OpenCode agent executes:
   - Set spec frontmatter status to `in-review`.
   - Construct a diff between `baseline_revision` and `final_revision`.
   - Then logs: "Now I'll launch both review subagents in parallel (synchronously, both blocking in this turn):" and calls the two subagents:
     - **Blind Hunter adversarial review** (first subagent)
     - **Edge Case Hunter review** (second subagent)
4. From this point:
   - bmad-loop TUI shows the run/story as `dev-running`, tokens `0`, no new `session-end` or `dev-decision` events in `journal.jsonl`.
   - OpenCode Desktop UI shows both subagent buttons as active, but **no further tool calls or progress occurs** until I type a message like `continue` or answer a follow-up question.
   - After manual input, the agent writes: "No — both review subagents were interrupted. Let me re-launch them and complete the review pass." and reissues both subagent calls.

This pattern has occurred on **multiple consecutive runs** of the same story, with the same adapter and model, and the same outcome: step-04 review stalls at the parallel subagent launch until manual user interaction in OpenCode Desktop.

## Expected behavior

- When `bmad-dev-auto` launches review subagents via `opencode-http` in parallel, the parent session should either:
  - stay in the foreground and stream subagent output until both complete, or
  - have a robust wake-nudge mechanism that can revive a stalled parent without requiring manual intervention.
- bmad-loop should be able to drive the full dev+review pass autonomously, without the human having to babysit OpenCode and type `continue` every time subagents are launched in parallel.

## Actual behavior

- Parent dev/review session goes idle after the parallel subagent launch.
- bmad-loop remains in a `dev-running` state with 0 tokens for that story, no ATTENTION or token-budget signals, and no journal progress.
- OpenCode Desktop shows the subagent buttons but they appear stuck until I send a manual message.
- The agent acknowledges that "both review subagents were interrupted" and re-launches them, implying that the original subagent sessions were cut off mid-run.

## Notes and hypotheses

- The `opencode_http.py` adapter source includes explicit handling for child sessions and an `activity` counter, with a test `test_sse_dispatch_filters_child_sessions` that says: "the parent is silent while a child session streams, exactly like subagent output in a tmux pane". This explains why the parent is quiet while the subagents run, but does not explain why both are *interrupted* and require manual input.
- bmad-loop's mid-session stall nudge (`limits.dev_stall_nudges_cap`) is supposed to nudge idle sessions and then declare them stalled if they don't produce a result. In my runs, no automatic nudge is visible in `journal.jsonl` before I intervene manually.
- A similar issue was previously fixed for Claude Code by forcing `CLAUDE_CODE_DISABLE_BACKGROUND_TASKS=1` (see v0.9.0 changelog). I could not find an equivalent safeguard for the `opencode-http` adapter.
- This may be related to OpenCode's subagent design: subagents can run foreground or background, and there is currently no global toggle to force one behavior; a community workaround suggests using a very large `yieldMs` to keep tasks in the foreground.

## What would help

- Guidance on whether `opencode-http` is expected to handle parallel review subagents fully autonomously, or whether the current behavior is a known limitation.
- If this is a bug, a fix similar in spirit to the Claude Code safeguard: ensure that review subagent launches via `opencode-http` either stay foreground or are handled cleanly so the parent session does not stall.
- If there is a configuration-based workaround (e.g. in `opencode.json` or bmad-loop's profile for `opencode-http`) that can force foreground behavior or adjust the stall nudges, documenting it would help operators avoid this stall.

Thanks — happy to provide a sanitized `bmad-loop diagnose <run-id>` bundle or relevant slices from `journal.jsonl` / `logs/*.log` if that would help reproduce this on your side.

### LINKED ISSUE #470: Stall detection never arms: a silent session burns the full session_timeout_min (stall_armed: false after 3h12m of no output)
## Summary

A review session died ~17 minutes in on a transient API disconnect, then sat **completely silent for 3h12m** until the wall-clock timeout reaped it. The stall detector never armed: its heartbeat records `stall_armed: false` and `stall_nudges_sent: 0` after more than three hours of zero output, with `dev_stall_grace_s = 600` and `dev_stall_nudges = 2` configured.

The effect is that any transient disconnect costs a **full `session_timeout_min`** instead of the ~10 minutes the stall machinery is designed to cost. In our case that was the difference between ~10 minutes and 4 hours, twice in one run.

## Environment

- `bmad-loop` 0.9.0 (`uv tool install` from git `f512288f4e7ca4f028da324b280762fd413b9f90`)
- Stage adapters: dev = codex (`codex-cli` 0.146.0), review = claude (`2.1.222`)
- `scm.isolation = "worktree"`, macOS arm64
- Relevant policy:
  ```toml
  [limits]
  session_timeout_min  = 240
  dev_stall_grace_s    = 600
  dev_stall_nudges     = 2
  dev_stall_nudges_cap = 6
  ```

## What happened

A review session hit an `API Error: Connection closed mid-response` about 17 minutes in (the underlying cause on our side was host sleep dropping TCP — but that is incidental; any disconnect reproduces it). The agent stopped and never resumed. The session process stayed alive at an idle prompt.

Timeline for that session:

| | |
|---|---|
| session start | 22:56 |
| last byte written to its capture log | **23:42** |
| session killed | **02:54** |
| silent interval | **3h12m** |
| capture log size for the whole ~4h session | **29,638 bytes** |

## Evidence

`tasks/<story>-review-3/heartbeat.json`, after three hours of silence:

```json
{
  "ts": 1785997345.6384351,
  "remaining_s": 14125.55,
  "stall_armed": false,
  "stall_nudges_sent": 0
}
```

`tasks/<story>-review-3/session-lifecycle.jsonl`, the only two terminal events:

```json
{"ts": 1785999279766577000, "event": "timeout-fired", "expired_clock": "wall", "timeout_s": 14400, "mono_remaining_s": 14114.74}
{"ts": 1785999279800755000, "event": "kill-outcome", "reaped": true, "forced": [], "unreaped": [1372]}
```

Two things stand out:

1. **`stall_armed: false` with `stall_nudges_sent: 0`.** With `dev_stall_grace_s = 600`, the grace window should have expired ~10 minutes into the silence, armed, and sent up to 2 nudges before declaring the session stalled. None of that happened across 192 minutes.
2. **`expired_clock: "wall"`** confirms the wall-clock timeout was the *only* thing that ended it — the stall path never participated.

Also visible: `unreaped: [1372]` — a child pid survived the kill.

## Why this matters more than it first looks

The stall detector is the mechanism that makes a large `session_timeout_min` safe. We raised ours from 90 to 240 because a legitimate dev pass needed 1h40m — that is a reasonable thing for a user to do, and the documented stall settings imply a dead session will still be caught quickly. Because arming never happens, raising the wall clock instead **multiplies the cost of every dead session**, and the failure presents as "the review took 4 hours" rather than "the review died after 17 minutes".

## Possible cause worth checking

The knobs are named `dev_stall_*` and `dev_stall_nudges_cap`'s comment says "for a dev/review session", so reviews appear intended to be covered. Worth confirming the arming path is actually wired for review-stage sessions, and that arming keys off *capture-log growth* rather than process liveness — the process here was alive and idle the whole time, which is exactly the case the stall detector exists for.

## Suggested handling

- Arm the stall grace on absence of capture-log growth, independent of stage and of process liveness.
- Treat a terminal `API Error:` line in the transcript as an immediate stall signal rather than waiting for the grace window.
- Surface `stall_armed` / `stall_nudges_sent` in `bmad-loop status`, so "hung" is distinguishable from "working" without reading `heartbeat.json` by hand.
- Investigate the `unreaped` pid on the kill path.

### LINKED ISSUE #504: Two of three nudge sites are unguarded: send_text raises on a dead window and the exception escapes the session loop
## Summary

`send_text` raises on a nonzero exit — `BaseTmuxBackend.send_text` (`adapters/tmux_base.py:328-331`)
goes through `_tmux` → `_run(check=True)`, and `TmuxError` subclasses `MultiplexerError`
(`tmux_base.py:44`, `multiplexer.py:42`). Three call sites nudge an idle session, and only one of them
guards that:

| site | call | guarded |
|---|---|---|
| `generic.py:676` budget nudge | `self.send_text(handle, BUDGET_NUDGE_TEXT)` | **yes** — `except MultiplexerError`, comment: *"a dead/hung window can't take the nudge"* |
| `generic.py:782` stall nudge | `self.send_text(handle, STALL_NUDGE_TEXT)` | no |
| `generic.py:857` generic nudge | `self.send_text(handle, NUDGE_TEXT)` | no |

Nothing upstack catches it either: `grep -n "TmuxError\|MuxError" src/bmad_loop/engine.py` is empty, and
the three `except Exception` blocks in `generic.py` (`:1061`, `:1067`, `:1125`) are in the teardown
path, not the poll loop. So the exception escapes the session loop.

## Why it is reachable

A nudge fires precisely when the session has gone quiet, and the window can die inside the same tick.
The code says so itself, in the comment immediately after the unguarded stall nudge
(`generic.py:786-789`):

> Re-probe liveness before finalizing: this return exits the loop, so a hard death (no SessionEnd) in
> the gap since the top-of-tick probe would otherwise never be caught.

That re-probe sits on the *finalize* branch, after the nudge branch has already run. On the tick where
the stall deadline expires and the window died since the top-of-tick probe, the nudge is attempted
against a dead target first.

## Platform status

- **tmux (Linux/macOS): live today.** `send-keys` to an unresolvable target already exits 1:
  ```console
  $ tmux send-keys -t S:nosuchwindow 'echo x' Enter
  can't find window: nosuchwindow
  rc=1
  ```
  So an unguarded nudge raises out of the loop, and the run dies with a traceback instead of taking the
  crash path that honors the session's artifact.
- **psmux 3.3.7: masked, and masked by a worse bug.** `send-keys` to a stale target exits 0 there and
  the keystrokes are typed into whichever window is active — psmux/psmux#545, which we filed and which
  is now **fixed on psmux master** (`194fac0`, `send-keys` is in the new validator's command list). The
  moment 3.3.8 is admitted, psmux joins tmux and this path starts raising.

So the same admission that removes a silent-misroute hazard converts it into an unhandled exception on
two of three sites.

## Ask

Give the stall and generic nudge sites the guard the budget site already has, and decide what a failed
nudge means: the budget site treats it as "the window is gone, stop nudging" and continues to the
liveness re-probe, which looks right for all three.

Worth checking in the same pass whether any other `MultiplexerError`-raising seam call sits unguarded
inside the poll loop — `send_text` is the one that runs *because* the session looks dead, so it is the
most exposed, but the audit is cheap while someone is in there.

Found during the psmux pre-3.3.8 bughunt, while checking which adapter tolerance points the upstream
target-validation fix would harden

Refs: psmux/psmux#545, #222

pbean:
## Problem

Generic and OpenCode dev/review sessions could stay silent from launch without ever arming stall recovery, so an interrupted but still-live session could consume its full `session_timeout_min`. OpenCode parents interrupted around subagent work had the same unbounded wait before any idle event, while two generic completion-loop nudge sends could raise `MultiplexerError` against a dead target and abort before deterministic liveness and artifact handling ran.

## Fix

- Arm generic and OpenCode stall grace at session launch, re-arm it on pane/SSE activity and later turn-end evidence, and retain the configured bounded nudge budget.
- Let a silent OpenCode parent enter the same bounded recovery while parent/child SSE activity and a provably busy or retrying status keep delaying any nudge.
- Contain `MultiplexerError` at exactly the generic stall-nudge and result-less-Stop nudge sends so the following deterministic liveness/artifact logic owns the verdict.

## Scope note

This does not prevent OpenCode's upstream subagent interruption and cannot guarantee that an accepted `prompt_async` nudge wakes the parent. It does not add transcript `API Error` parsing, status counters, child-reaping changes, foreground/background/provider workarounds, psmux admission or quoting work, or broad multiplexer failure tolerance.

Completion semantics are unchanged: pane/log/SSE prose never completes a session. Completion still requires Stop/idle evidence or window/server death followed by deterministic artifact verification.

## Visible behavior diffs

| Before | After |
| --- | --- |
| A generic or OpenCode session silent from launch could leave stall recovery disarmed and wait for the full session timeout. | Positive stall grace starts at launch; activity re-arms it, while true silence spends bounded nudges and becomes `stalled` near the grace sequence. |
| A silently interrupted OpenCode parent had no bounded recovery before an idle event. | It enters bounded recovery from launch, but child/parent activity and provable busy/retry status postpone nudging. |
| Either of two generic completion-loop nudge sends could abort on a dead mux target. | Those two sends contain the transport error, after which deterministic liveness/artifact handling decides `crashed` or `stalled`. |

## Testing

Full gates:

```text
uv run pytest -q -n logical
uv run pyright
trunk check --all
uv run --no-project python scripts/release.py check
git diff --check
```

Results: `5768 passed, 50 skipped in 49.40s`; Pyright reported `0 errors, 0 warnings, 0 informations`; Trunk checked 257 files with no issues; release metadata and changelog checks passed; `git diff --check` passed.

Ablations (2026-08-16), each performed singly from an external `cp` backup, then restored byte-identically and re-run green:

- Generic launch arm (`uv run pytest tests/test_generic_tmux.py::test_dev_stall_arms_at_launch_without_stop -q`): `1 failed in 0.07s`; the ablated result was `('timeout', [])` instead of bounded `stalled` with two nudges. Restored: `1 passed in 0.05s`.
- Generic stall-nudge guard (`uv run pytest tests/test_generic_tmux.py::test_dev_stall_nudge_send_failure_reaches_liveness_verdict -q`): `1 failed in 0.10s`; the ablated path raised `MultiplexerError: window gone`. Restored: `1 passed in 0.13s`.
- Generic result-less-Stop guard (`uv run pytest tests/test_generic_tmux.py::test_resultless_stop_nudge_send_failure_reaches_liveness_verdict -q`): `1 failed in 0.10s`; the ablated path raised `MultiplexerError: window gone`. Restored: `1 passed in 0.03s`.
- OpenCode launch arm (`uv run pytest tests/test_opencode_http.py::test_dev_stall_arms_at_launch_without_stop -q`): `1 failed in 0.06s`; the ablated result was `('timeout', [])` instead of bounded `stalled` with two nudges. Restored: `1 passed in 0.03s`.

Closes #411
Closes #470
Closes #504

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery for silent development and review sessions, including sessions that stall immediately after launch.
  * Session activity and fresh completion signals now extend recovery time appropriately.
  * Failed wake-up attempts no longer interrupt completion monitoring.
  * Active work remains protected during busy or retry states.

* **Documentation**
  * Updated guidance for stall handling, recovery limits, long-running tests, and relevant settings.


pbean:
@codex review

pbean:
@codex review

PR #644, opened 2026-08-17
Expose verification command results to post-dev plugins

### LINKED ISSUE #641: Expose structured verification command results to post_dev_verify plugins
**Describe your idea**

Expose authoritative per-command verification results to the existing `post_dev_verify` plugin hook, while retaining those results durably in the run journal.

**Why is this needed?**

BMad Loop already owns verification command execution and commit gating, but plugins cannot currently consume the verifier's per-command results. Successful results are discarded after outcome classification, and the available output is a merged, bounded `output_tail`.

This makes plugin-based auditability, reproducibility, reporting, and external evidence integrations rely on indirect inference instead of the actual verifier result.

**How should it work?**

Following maintainer guidance on Discord, keep the existing plugin architecture rather than adding a new observer abstraction:

- journal structured per-command verification results, using the existing pointer idiom for output payloads rather than inlining unbounded output
- extend `CommandResult` with separate `stdout` and `stderr` while preserving `output_tail`
- pass `command_results` through the existing `post_dev_verify` hook context
- emit `post_dev_verify` on the fix/retry verification path as well as the ordinary dev path

The journal and hook context should retain existing run/story/stage correlation available in core. Verification pass/fail and commit gating remain core-owned.

**PR**

No implementation is underway. I will wait for maintainer feedback before proposing an implementation PR.

**Additional context**

Non-goals:

- core-owned evidence storage, uploading, signing, or database persistence
- required-durability commit blocking or pause-on-write-failure
- domain-specific audit policy
- a new observer/event framework or workflow engine

Compatibility expectations:

- existing `output_tail` remains available
- existing verification and commit semantics remain unchanged
- no new mandatory configuration
- plugins that ignore `command_results` are unaffected

Storage formats, upload, signing, and any pause-on-write-failure behavior remain opt-in plugin responsibilities.

kamiotakaaki:
## Problem

`post_dev_verify` did not expose the structured results core obtained from verification commands. Successful command results were discarded after classification, and the available failure output was only the merged bounded `output_tail`.

## Change

- retain one structured `verify-command-result` journal entry per dev or repair verification command, with run/story/attempt/stage/sequence correlation and pointers to separate stream logs;
- add separate `stdout` and `stderr` fields to `CommandResult`, while retaining the existing `output_tail` compatibility field;
- expose immutable `command_results` through the existing `post_dev_verify` hook context;
- emit that existing hook after repair verification as well as normal dev verification.

Core still executes and classifies verification. Hooks can observe results but cannot replace them, turn a failure into success, or alter commit gating. Storage, upload, signing, and any policy response remain plugin responsibilities.

## Compatibility

No configuration is required. Existing plugins that ignore `command_results` continue unchanged. `output_tail` remains available and retains its existing merged, bounded behavior.

## Validation

- `uv run pytest tests/test_verify.py tests/test_hook_bus.py -q --deselect tests/test_verify.py::test_z_output_undecodable_path_becomes_git_error` — 270 passed, 33 skipped, 1 deselected
- focused new verifier/hook/normal-dev/fix-path tests — 4 passed
- `uv run ruff check src/bmad_loop tests/test_verify.py tests/test_hook_bus.py tests/test_engine.py` — passed

On this APFS host, three existing non-UTF-8 filename tests fail before the code under test can run. Each fails identically on clean upstream base `01ad84341f17029ebee0615d67928ecfdaa5c89e`; this PR does not alter them.

Closes #641

## Summary by CodeRabbit

- **New Features**
  - Development and repair verification hooks now provide per-command results, separate output streams, verification stages, and sequence identifiers.
  - Verification output is stored separately with configurable capture limits, byte counts, truncation status, and capture-error reporting.
  - Hook results distinguish skipped, empty, and completed verification passes.

- **Bug Fixes**
  - Improved protection against invalid or redirected verification paths.
  - Improved preservation and validation of specification state during retries, repairs, reviews, and recovery.
  - Improved timeout output handling and sensitive journal-data sanitization.
  - Cleanup now removes verification output and accurately reports reclaimed storage.

- **Documentation**
  - Expanded verification storage, correlation, retention, configuration, and plugin guidance.
  - Added documentation for the session token-limit setting.


kamiotakaaki:
@coderabbitai review

kamiotakaaki:
@coderabbitai review

pbean:
@codex review

pbean:
@coderabbitai review

pbean:
@codex review

pbean:
@coderabbitai review

pbean:
@codex review

pbean:
## Phase 6 (final): scope boundary + bot gate

### Scope decision on the review leg

The journal records and the `post_dev_verify` surface cover the **dev phase only** — the dev
session's verification and each repair session's. The **review gate** runs the same
`[verify] commands` through the same core classifier and retains nothing.

Kept that boundary rather than widening this PR, and **wrote it down** instead of leaving it
implicit — an unstated boundary is the real defect in a feature sold as auditability:

- `docs/plugin-authoring-guide.md` now states it, including the two ways it can mislead a
  handler: `verify-command-result` is **not a census** of a run's verifier invocations (every
  story reaching a commit ran the commands at least once more than the records show), and a red
  journalled `fix` pass does not mean the commit was blocked.
- `CHANGELOG.md` carries the same scope note.
- `_dev_then_fix_run`'s docstring names it at the fixture that pins it: four scripted verifier
  returns, two journalled sequences — the inequality is the boundary, not a miscount.
- Filed as #656 with the five uncovered gates and a suggested shape.

`confirm --reverify` is outside this path **by construction**, not omission: the run that parked
the story is finished, so there is no run journal to append to and no hook bus to emit on.

Why it is a separate change: the three `verify_review*` functions end on `verify_commands_outcome`
*after* early-return gate chains, so the engine cannot pre-run the commands. Retaining results
means a changed core return shape used by all three engine classes, plus a new hook stage —
`post_dev_verify` is named for dev and now documents itself as firing on both legs of it.

### Bot gate: three rounds, two P1s, both real

**Round 1 — symlinked `verify/` parent (P1).** Confirmed by reproduction, not by reading:
`mkdir(parents=True, exist_ok=True)` accepts a symlink-to-directory (it re-raises only when
`is_dir()` is false, and that follows links), and `follow_symlinks=False` guards only the final
component — so the write landed outside the run dir while `write_verify_stream` returned the
pointer `verify/v.stdout.log`, claiming containment it did not have. Sessions really can plant
it: they are handed `BMAD_LOOP_RUN_DIR` and write `result.json` there.

Fixed with the primitive this repo already keeps for it (`tui/launch.py` writes its
control-window record the same way): `open_dir_confined` walks each component below the run dir
`O_NOFOLLOW`, and `atomic_write_text_at` never names a path again — so a re-plant between check
and write renames something the code no longer consults. Refusal raises `OSError`, which is the
caller's **existing** degrade path: the record still lands with a null pointer and
`capture_error`, so no new failure mode.

**Round 2 — junctioned `verify/` on the win32 fallback (P1).** The guard added in round 1
checked `is_symlink()`, which is False for a directory junction. `mklink /J` needs no elevation
while a directory symlink needs `SeCreateSymbolicLinkPrivilege` or Developer Mode — so it
covered the privileged half and left the **unprivileged** half open, with no race to win.

The predicate moves to `platform_util.is_link_like()`, reparse-tag based. Deliberately not
`os.path.isjunction()` (3.12+, and `requires-python` is `>=3.11`), and deliberately not "any
reparse tag" — OneDrive placeholders and dedup stubs are reparse points too. `events.py` and the
standalone hook relay keep their own copies on purpose: they run under the host's interpreter.

**Round 3 — clean** on `7d926f8d`.

Also fixed a **correct** CodeRabbit finding: `match="verify.stream_capture_kb"` is a regex whose
unescaped dots also accept a message naming a different key (measured). Its second half —
apply the same change to `core.toml` — did not apply; that file is TOML settings, not a regex.

### Notes on the work itself

- **Ablated, not assumed.** The two refusal guards **overlap**, so ablating the POSIX arm alone
  reddens only the *message* — removing **both** is what reproduces the escape. Recorded in the
  test docstring so a future reader does not mistake the weaker ablation for the proof.
- **The round-1 fix silently disarmed an existing test.** The ENOSPC degrade test patched
  `bmad_loop.journal.atomic_write_text` only, which on POSIX now intercepts nothing — it was
  green without ever running the degrade arm. It patches both writers now. Only the full suite
  caught this.

### Validation on `7d926f8d`

`uv run pytest -q -n logical` — **5817 passed, 50 skipped**
`uv run pyright` — **0 errors, 0 warnings, 0 informations**
`trunk check --all` — **no issues**, 258 files

### Outstanding, for the maintainer

- **Merge conflict:** exactly one file, `CHANGELOG.md` (everything else auto-merges). Pre-dates
  this phase's commits. Left unresolved deliberately — the branch is intentionally not rebased.
- **CI has not run:** cross-repository fork PR, so the workflows need maintainer approval.
- **CodeRabbit has not covered these commits.** Its last review is at `b82ba46b`; manual requests
  are declined as redundant because it auto-reviews incrementally on push. It reviews from its own
  last-reviewed commit, so a later pass can still cover the whole span.

pbean:
@codex review

pbean:
@codex review

pbean:
@codex review

pbean:
@codex review

pbean:
@codex review

pbean:
@codex review

pbean:
## Updated summary

The earlier summary comment predates the merge from `main`, six rounds of
review-driven fixes, the reclaim work and the in-memory stream ceiling, so here is the
current state of the branch. Head is `9533ab63`.

### What this adds

`post_dev_verify` plugins now receive immutable per-command verification results —
separate `stdout`/`stderr` alongside the existing bounded `output_tail`, plus
`verification_stage` (`"dev"` / `"fix"`) and `verification_sequence`. That sequence is
the only way to tell a dev verification from a repair one (both emit the same stage
from the same phase) and is the key that joins a hook context to its own journal
records. Core writes `verify-command-result` records with stream pointers under the
run's own `verify/` directory — deliberately not the adapter-owned, TUI-consumed
`logs/`. Plugins observe; they still cannot alter verification or commit decisions.

### Scope boundary — deliberate, documented, ticketed

Records and hook cover **the dev phase only**. The review gate runs the same
`[verify] commands` and retains nothing, so the journal records are *not* a census of
a run's verifier invocations. Closing that is not a small change: the three
`verify_review*` functions end on `verify_commands_outcome` **after** early-return
gate chains, so the engine cannot pre-run the commands — it needs a changed core
return shape adopted by all three engine classes plus a new hook stage.
`confirm --reverify` is outside by construction (a finished run has no journal and no
bus). The boundary is stated in `docs/plugin-authoring-guide.md`, in the CHANGELOG,
and on the `_dev_then_fix_run` fixture; **#656** tracks closing it. No closing
keywords are used anywhere for it — negated or otherwise.

### Review rounds

Six rounds. Every finding is fixed; each bullet names the commit and what was
measured before the fix:

- **P1 — path redirect (POSIX).** `atomic_write_text(follow_symlinks=False)` guards
  only the *final* path component; parents are still OS-resolved, and
  `mkdir(parents=True, exist_ok=True)` accepts a symlink-to-directory because it
  re-raises only when `is_dir()` is false — which follows links. Measured: a planted
  `verify/` symlink put the file outside the run dir while the record still reported
  the pointer `verify/…`, i.e. it *claimed* containment it did not have, and a session
  can plant it via `BMAD_LOOP_RUN_DIR`. Fixed in `96b1beb5` with the repo's existing
  pair — `open_dir_confined` (walks components `O_NOFOLLOW`) + `atomic_write_text_at`
  (never names a path again), as `tui/launch.py` already does.
- **P1 — the same escape on Windows.** `is_symlink()` returns **False** for a
  directory junction, and `mklink /J` needs no elevation at all while a directory
  symlink needs `SeCreateSymbolicLinkPrivilege`/Developer Mode — so the unprivileged
  half of the escape was the one left open. Fixed in `7d926f8d` via reparse tags
  (`platform_util.is_link_like`), not `os.path.isjunction` (3.12+, floor is 3.11) and
  not "any reparse tag" (OneDrive placeholders are reparse points).
- **P2 — the store was invisible to `diagnose`.** `verify/` was missing from
  `diagnostics._FILE_CATEGORIES`, so dumps omitted its count and size — up to 512 KiB
  per command per attempt. Fixed in `0e39fc60`, registered through the `VERIFY_DIR`
  constant so the reporter cannot drift from the writer.

- **P2 — the reclaim path had the mirror image of the write-path escape.**
  `shutil.rmtree` *refuses* a directory symlink by design (following it would delete
  the target's contents), and `ignore_errors=True` swallows that refusal — so
  `trim_run_dir` appended a planted `verify/` link to `removed` and left it exactly
  where it was. `_dir_size` had the same shape of bug against its own docstring:
  `os.walk` does not descend into links but *does* follow the top path it is handed,
  so sizing a redirected entry billed the reclaim for out-of-run bytes. Both measured
  before fixing — the link survived and the estimate came back at exactly the target's
  size. Fixed in `0dc29283`: remove the redirect itself (`unlink`, falling back to
  `rmdir` for the win32 directory symlink/junction that `DeleteFileW` rejects), and
  refuse a link-like path in `_dir_size`. Both use `is_link_like`, not `is_symlink` —
  the junction is the unprivileged arm and `is_symlink()` reports False for it.

- **P2 ×2 — the same class again, in the two paths that *measure* a run.**
  `summarize_files` admitted a category root on `root.is_dir()` (follows a link) and
  walked it with `rglob("*")`: a redirected `verify/` made `diagnose` report an outside
  tree's 2 files and 3100 bytes as this run's retained verifier output. And `_dir_size`,
  now refusing a link-like *top*, still used a plain `os.walk` — which prunes a
  symlinked subdirectory via `os.path.islink`, False for a junction, so a *nested*
  junction was still descended into on win32. Fixed in `f722d8a5` with one helper,
  `walk_files_unlinked`, since both callers got it wrong identically.

- **P2 — `stream_capture_kb` bounded only the DISK.** Full streams stayed in the
  results list while the remaining commands ran, so peak memory scaled with the
  **number** of configured verify commands rather than the largest one. `336d5198`
  adds a hard `verify.MAX_STREAM_MEMORY_BYTES` (32 MiB per stream) applied at the
  subprocess boundary on both the normal and the timeout leg — a backstop, not a
  tuning dial, so deliberately a constant rather than a policy knob. The subtlety is
  that truncating in memory silently breaks the `*_truncated` contract: the record
  sizes off the string in hand, so a ceilinged stream would report what was HELD as
  what was EMITTED and compute `captured < full` against a false baseline — calling a
  cut stream whole, the one thing that flag exists to prevent. `CommandResult.*_full_bytes`
  carries the truth (default `None`, so existing three-field constructions stay
  correct); the ablation reads 100 against 9,000,000. `verify.byte_tail` is now the
  single tail-slice implementation for both bounds, because a second copy of that
  slice is where `errors="replace"` overflows the cap it is enforcing.

- **P1 — a plugin could erase a CRITICAL escalation, and this one predates the
  branch.** `HookContext` copies `result_json` so a plugin observes rather than
  rewrites history — and the comment at the copy says exactly that — but `dict()` is
  shallow, so the nested `escalations` list stayed the engine's own object. Both
  verify legs emit `post_dev_verify` before reading
  `critical_escalations(result.result_json)`, so an in-process plugin that cleared
  that list erased the escalation before the audit ran and a verify-green repair
  proceeded where the run owed a pause. Verified against the merge-base: the shallow
  copy and the dev-leg emit-before-audit ordering are both on `main`, so this branch
  did not open the hole — it added a second emit that shares it. Fixed deep in
  `68703aaa`; it goes straight at the feature's "plugins observe, cannot alter"
  claim, which is why it is closed here rather than deferred.

- **P2 — a regression this branch introduced, in `f722d8a5` above.** Swapping
  `rglob("*")` for `walk_files_unlinked` dropped the `is_file()` guard the old loop
  carried: `os.walk` puts every non-directory entry in `files`, FIFOs included, and
  the `logs` arm *opens* what it counts — so a FIFO planted in `logs/` blocked
  `diagnose` forever. Fixed in `9533ab63`. The inventory now filters on `lstat` +
  `S_ISREG` (not `is_file()`, which follows and answers about the target), and the
  line count anchors on the descriptor the way `runs.read_trusted_config_digest` and
  `tui.launch._read_ctl_window` already do for this hazard: `O_NONBLOCK` so a FIFO
  cannot block the open, `S_ISREG` on the **fd** because a path check before the open
  is a race on a directory the session can write, and `O_NOFOLLOW` so the final
  component cannot redirect the read out of the run. `walk_files_unlinked` no longer
  *promises* regular files in its docstring — that false claim is what made the swap
  look safe.

Worth stating plainly: this class kept re-opening because registering `verify/` made a
**session-plantable** directory reachable by code that had only ever seen
engine-created ones. Each round found the next path that assumed its own tree —
reclaim, then sizing, then diagnostics, then the nested junction. Those rounds are not
unrelated defects; they are one wrong assumption surfacing wherever a walk, a size, or
a delete trusted a name. AGENTS.md treats that as evidence about the approach rather
than a defect queue, so round six was escalated for a scope decision instead of being
ground through: the call was to close both findings here, since one is a regression
this branch shipped and the other attacks the feature's central claim.

CodeRabbit's `match=` finding is fixed in `f20440d6` (an unescaped regex in a
`pytest.raises` pattern).

### Two failures worth calling out

- **A Windows-only defect the local gates structurally could not see.** Test fixtures
  stubbed `run_verify_commands` with the command name `"check"`. On POSIX that is
  inert — `env_fault_reason` keys on shell rc 126/127 and never reads the name. On
  Windows, `_win32_env_fault_reason` *also* does a PATH lookup on the leading token,
  so a failing result naming a nonexistent binary classifies as an environment fault,
  the run pauses, and the tests read `done=0`. Three tests broke on Windows CI against
  a fully green local suite. Fixed in `1c80f647` by using `"pytest -q"`, which is what
  every other stub in `test_engine.py` already uses. The three local gates are
  Linux-only and cannot reach a win32 branch — worth remembering.
- **The verifier stream store was never reclaimed.** `stream_capture_kb` defaults to
  256 KiB per stream, so a run accumulated up to 512 KiB per verify command per
  attempt and kept it for as long as the run dir survived — the P2 above taught
  `diagnose` to *show* the problem, but nothing closed it. `75ab6582` adds `VERIFY_DIR`
  to `runs._HEAVY_RUN_ENTRIES` so `clean` trims it with the worktrees, on the same
  bargain worktrees already make: a trimmed run is one you can still see and resume,
  not one you can still re-read every artifact of. It also sizes the reclaim estimate
  over `heavy_run_entries` instead of `worktrees/` alone — both changes are needed and
  they fail differently: with only the tuple changed, `clean` removes the store and
  reports freeing **0 bytes**.

### Verification

Every new test was ablated, including two whose docstrings claimed *less* than the
ablation actually showed — those were corrected rather than left flattering. One
caveat recorded on the spot: overlapping guards make a single-axis ablation prove less
than it looks. Ablating the POSIX arm alone only reddened the *message*, because the
win32 fallback still refused; removing **both** is what reproduces the escape.

Local on `9533ab63`: **6061 passed, 50 skipped**; `pyright` **0 errors**;
`trunk check` clean.

The round-six work adds five tests across five ablation axes, and each axis reddens
exactly one test — `O_NONBLOCK` (the alarm fires), `S_ISREG` on the fd (counts three
piped lines *and* drains the pipe), `O_NOFOLLOW` (returns the symlink target's two
lines), the inventory `S_ISREG` (the group reports three files instead of one), and
the deep copy (the audit comes back empty, naming the escalation that vanished).
Disjoint failures are the evidence that five guards are not standing in for each
other. One docstring was corrected after its ablation disagreed with it: the fed-FIFO
case was predicted to return 0 via `BlockingIOError` and actually returned **3**, so
the test now says what was measured rather than what was expected.

One ablation was worth more than it looked: removing the link-like-top refusal reddens
tests in *two* independent call sites, which is what shows the shared helper is
load-bearing rather than decorative. Another was worth less than it looked and is
documented as such — `os.walk` prunes symlinked subdirectories on its own, so the
nested-redirect guard is invisible on POSIX. Its test simulates the
`is_link_like`/`is_symlink` disagreement that *is* the win32 behaviour, rather than
planting a real symlink and quietly grading `os.walk` instead of the fix.

### Notes for the merge decision

- The CHANGELOG conflict with `main` was resolved by **merging** `origin/main`
  (`3282a83b`), deliberately not rebasing — a rebase rewrites every SHA and would
  orphan the bot verdicts anchored to them.
- The Linux `test (py3.11)` job on this PR wedged four times on `Install tmux` — an
  infrastructure fault unrelated to this branch: `apt-get update` produced no output
  for 14 minutes until the job cap, so `Run tests` never ran at all. Root cause and fix
  are in **#657**, kept out of this PR because this branch does not touch `.github/`.
  That investigation turned up something worth knowing here: the runner image already
  ships tmux, so the step had never installed anything — it was taking the network's
  risk for a no-op. #657 now asks `tmux -V` first; measured on its own CI, the step
  went from 62s to **0-1s** on all four legs with zero network calls. Merging #657
  first would stabilize this job.

Issue #640, opened 2026-08-17
[BUG] `rearm_escalation` advances the task baseline but not the spec's `baseline_revision` — completed re-drive attempts are mechanically rejected in a loop


mkalika:
### Description

After an escalation is resolved in **from-scratch mode** (no `restore_patch`), `rearm_escalation`
advances the orchestrator's `task.baseline_commit` to the project's current HEAD (per its own
docstring: *"The baseline itself is advanced to the project's current HEAD … so commits and files
the resolve session produced count as the rebuild's starting point"*). However, the story spec's
`baseline_revision:` frontmatter is **not** updated — rearm edits `status:` and strips the stale
`## Auto Run Result`, but never touches `baseline_revision`.

The re-driven dev session then completes and (correctly, from its point of view) reports the
baseline its spec states. The orchestrator's result validation rejects it:

```json
{"kind": "dev-decision", "story_key": "1-7-corpus-distillation-immutable-snapshots",
 "attempt": 2, "session_status": "completed", "action": "retry",
 "reason": "spec baseline f4574e21b96b does not match orchestrator-recorded baseline 925077efee11",
 "env_fault": false, "session_vanished": false}
```

The completed work is rolled back (parked on `attempt-preserve/` refs) and a retry is dispatched.
The retried session reads the **same spec** — BMAD artifact folders are deliberately preserved
through resets while `task.resolved_redrive` is latched — so it inherits the same stale
`baseline_revision` and, in our case, also the prior session's written justification for keeping
it. Result: the identical rejection repeats until `max_dev_attempts` is exhausted and the story
defers. We observed **two consecutive completed implementations** (+22k lines each, one of them
having passed the session's own four-layer review) discarded on this exact line, at a cost of
roughly 4.5M weighted tokens, before we intervened manually.

Aggravating factor: the session had *noticed* the discrepancy and chosen the spec's stale value
deliberately, documenting a plausible rationale in the Spec Change Log ("advancing to HEAD would
make the review diff empty"). Nothing in the dev contract tells the session that the
orchestrator-recorded baseline is authoritative, and the mechanical validator neither nudges nor
explains — it discards a completed attempt.

### Steps to reproduce

 1. Story completes dev+review; orchestrator commits it (commit **C**); a follow-up review then
   raises a CRITICAL escalation → run pauses. Spec frontmatter holds `baseline_revision: B`
   (the pre-story baseline).
[...]

Issue #624, opened 2026-08-16
test_sweep: feedback-file read at :3146 uses the platform default encoding


pbean:
`tests/test_sweep.py:3146`, in `test_migration_validation_failure_restores_ledger_then_escalates`:

```python
feedback = open(prompts[1].split("--feedback ", 1)[1]).read()
```

`open()` here uses the platform default encoding and leaks the handle. The default is not UTF-8 on Windows, and the repo runs its suite on `windows py3.11` and `windows py3.14`, so a feedback file carrying a non-ASCII byte would raise `UnicodeDecodeError` there and nowhere else. The ledger assertions in the same test family already pin `encoding="utf-8"` for exactly this reason.

Fix is the one-liner:

```python
feedback = Path(prompts[1].split("--feedback ", 1)[1]).read_text(encoding="utf-8")
```

**Provenance.** CodeRabbit raised this against the *sibling* call at `test_sweep.py:3306` during the #519 review (PR #616), where it was fixed. This line has the identical problem but is pre-existing on `main` and outside that PR's diff, so it was deliberately left rather than widening the change. Filing it so the remaining instance is not silently dropped.

Low severity: latent, and only reachable if a feedback file ever carries non-ASCII.

PR #647, opened 2026-08-18
test(verify): cover bundle-path baseline refusals; correct an abbreviation claim

### LINKED ISSUE #674: docs(verify): `same_commit`'s docstring misstates git's abbreviation default
`same_commit`'s docstring justifies its 7-character floor with a claim about git that does not hold:

```python
def same_commit(a: str, b: str) -> bool:
    """Hash equality tolerant of abbreviated forms (>= 7 chars, git's default
    --short length); sessions sometimes report `git rev-parse --short HEAD`."""
```

There is no fixed "default `--short` length". With `core.abbrev` unset git uses `auto`, which scales the abbreviation with the repository's reachable-object count and clamps **upward** to 7 only for small repositories. Measured:

- the `project` test fixture's sandbox repo (~7 reachable objects) → `git rev-parse --short HEAD` yields **7**
- this repository → yields **8** (`78c3a748`)

So the sentence is accidentally true in the test sandbox and false in the repo the orchestrator actually runs in. Git's documented hard minimum of **4** is a separate mechanism governing explicit `core.abbrev` / `--short=<n>` values, and git still lengthens beyond it whenever uniqueness demands.

The 7 is a constant mirroring nothing external — `_OBJECT_ID`'s comment gets this right (*"Length floor mirrors `same_commit`'s 7"*), so the two sit in the same file disagreeing about where the number comes from.

**Impact is documentation-only** — no behavior depends on the claim — but it is the kind of premise that gets reused. It was on its way into a test docstring on #647 before being caught, and a related restatement was drafted into a review comment.

Also worth recording while this is open: `same_commit` currently has **zero executable call sites**. The only remaining references are prose (`verify.py:255`, `verify.py:346`, two test docstrings). Whether it should be retired is a separate question from the wording, and not one this issue proposes to answer.

Polloinfilzato:
## What

Four tests (five cases) covering the deferred-work bundle path's baseline gate — `verify_dev_bundle` with `allow_ancestor_baseline=True` — plus one documentation-only correction in `verify.py`.

Closes #674.

## Why

The gate's refusal tests on `main` all exercise the **dev** path. The bundle path takes the ancestor relaxation, and the invariant that makes that relaxation safe — canonicalize the claim **before** the ancestor leg, never after — has no test. A refactor could reintroduce the pre-#645 ordering without reddening anything.

This PR originally proposed a production change (`names_its_own_commit`) for the same problem. That requirement reached `main` through #645 instead, so the production change is dropped and the branch is reduced to the coverage `main` was missing.

## What each test pins

| Test | Claim under test |
|---|---|
| `..._symbolic_baseline_is_refused` | `baseline_revision: HEAD` on the bundle path |
| `..._all_hex_ref_baseline_is_refused[branch, tag]` | an all-hex ref pointing at a genuine ancestor |
| `..._single_char_ref_baseline_is_refused` | a ref that is a prefix of its own target |
| `..._below_floor_abbreviation_is_refused` | characterizes the deliberate 7-character floor |

The last two close boundaries `main` has no test for; both are enforced structurally by `_OBJECT_ID` today, with nothing asserting it.

Each test asserts `_canonical_commit_oid(...) is None` directly rather than relying on the refusal string. The canonical-is-None leg and the ancestor leg emit byte-identical reason templates, so a substring assertion cannot distinguish which one fired.

## Ablation

Per the repo's ablation doctrine, each docstring records the mutation that reddens it, and both were measured against the full suite:

- Bypass `_canonical_commit_oid` in the `allow_ancestor_baseline` leg only, leaving the newer-baseline leg canonical-gated → **5 failed, 6113 passed** — exactly the five new cases. The dev-path refusal tests stay green because none of them set the flag. A *full* pre-#645 restore is a strictly larger mutation that also strips canonicalization from the newer-baseline leg; under that one `test_verify_dev_refuses_an_all_hex_ref_baseline` reddens on both parameters.
- `_OBJECT_ID` floor relaxed from 7 to 4 → **1 failed, 6117 passed** — the floor test alone.

## Docstring correction (`src/bmad_loop/verify.py`)

`same_commit`'s docstring justified its 7-character floor as "git's default `--short` length". There is no fixed default: with `core.abbrev` unset git uses `auto`, which scales with the repository's reachable-object count and clamps upward to 7 only for small repos. The test fixture's sandbox yields 7; this repository yields 8. Documentation only — no behavior change, and no caller is affected (`same_commit` currently has no executable call sites).

## Testing

`uv run pytest -q -n logical`: 6118 passed, 53 skipped. `uv run pyright`: 0 errors. `trunk check --all`: 258 files, no issues.

No changelog entry — tests plus a documentation-only correction.

### DIFF
--- a/src/bmad_loop/verify.py
+++ b/src/bmad_loop/verify.py
@@ -317,8 +317,11 @@ def worktree_clean(repo: Path) -> bool:
 
 
 def same_commit(a: str, b: str) -> bool:
-    """Hash equality tolerant of abbreviated forms (>= 7 chars, git's default
-    --short length); sessions sometimes report `git rev-parse --short HEAD`."""
+    """Hash equality tolerant of abbreviated forms (>= 7 chars); sessions
+    sometimes report `git rev-parse --short HEAD`. The 7 is git's *minimum* auto
+    abbreviation, not a fixed default: `core.abbrev` defaults to `auto`, which
+    scales the length with repository size and clamps upward to 7 only for small
+    repos, so there is no single "default --short length" to mirror."""
     if len(a) < 7 or len(b) < 7:
         return a == b
     return a.startswith(b) or b.startswith(a)
--- a/tests/test_verify.py
+++ b/tests/test_verify.py
@@ -1891,6 +1891,150 @@ def test_verify_dev_bundle_ancestor_baseline_retains_untracked_proof(project):
     assert verify.verify_dev_bundle(task, project, rj).ok
 
 
+def test_verify_dev_bundle_symbolic_baseline_is_refused(project):
+    """The ancestor relaxation must not be buyable with a claim that pins
+    nothing. `baseline_revision: HEAD` is resolved when the gate runs, not when
+    the session stamped it, so it reads as an ancestor of the unit baseline
+    whenever the checkout's HEAD still equals the recorded baseline — exactly
+    the stale premise this gate exists to refuse. Canonicalization has to happen
+    before the leg, not after.
+
+    Ablation: bypass ``_canonical_commit_oid`` in the ``allow_ancestor_baseline``
+    leg only, letting it consume the raw ``claimed_baseline``, and the attempt
+    passes, failing the refusal assertion. The dev-path twin
+    (``test_verify_dev_refuses_a_symbolic_baseline_revision``) stays green under
+    it: it never sets the flag, and ``_OBJECT_ID`` rejects ``HEAD`` ahead of the
+    newer-baseline leg either way.
+    """
+    (project.project / "story-work.txt").write_text("stories 1.1-1.3\n")
+    git(project.project, "add", "-A")
+    git(project.project, "commit", "-q", "-m", "story work")
+    task = make_bundle_task(project, dw_ids=("DW-1",))  # baseline = new HEAD
+    assert verify.is_ancestor(project.project, "HEAD", task.baseline_commit)
+
+    sp = project.implementation_artifacts / "spec-1-1-a.md"
+    write_spec(sp, "in-review", "HEAD")
+    (project.project / "src.txt").write_text("review fixes\n")
+    rj = {"workflow": "auto-dev", "spec_file": str(sp), "dw_ids": ["DW-1"]}
+
+    assert verify._canonical_commit_oid(project.project, "HEAD") is None
+    out = verify.verify_dev_bundle(task, project, rj)
+    assert not out.ok and "does not match" in out.reason
+
+
+@pytest.mark.parametrize("ref_kind", ["branch", "tag"])
+def test_verify_dev_bundle_all_hex_ref_baseline_is_refused(project, ref_kind):
+    """Hex spelling does not make a claim immutable on the bundle path either.
+    An all-hex branch or tag pointing at a genuine ancestor would satisfy the
+    relaxation through the ref namespace, so the gate must disambiguate the
+    object id independently of refs before the leg is reached.
+
+    Ablation: bypass ``_canonical_commit_oid`` in the ``allow_ancestor_baseline``
+    leg only and both parameters pass verification. The dev-path twin
+    (``test_verify_dev_refuses_an_all_hex_ref_baseline``) stays green under that
+    mutation, because it never sets the flag. It does redden under a *full*
+    pre-#645 restore, which strips canonicalization from the newer-baseline leg
+    as well — a strictly larger mutation than the one named here.
+    """
+    ancestor = verify.rev_parse_head(project.project)
+    (project.project / "story-work.txt").write_text("stories 1.1-1.3\n")
+    git(project.project, "add", "-A")
+    git(project.project, "commit", "-q", "-m", "story work")
+    task = make_bundle_task(project, dw_ids=("DW-1",))
+
+    claimed_ref = "abcdef0"
+    if ref_kind == "branch":
+        git(project.project, "branch", claimed_ref, ancestor)
+    else:
+        git(project.project, "tag", claimed_ref, ancestor)
+    assert git(project.project, "rev-parse", claimed_ref) == ancestor
+    assert git(project.project, "rev-parse", f"--disambiguate={claimed_ref}") == ""
+    assert verify.is_ancestor(project.project, claimed_ref, task.baseline_commit)
+
+    sp = project.implementation_artifacts / "spec-1-1-a.md"
+    write_spec(sp, "in-review", claimed_ref)
+    (project.project / "src.txt").write_text("review fixes\n")
+    rj = {"workflow": "auto-dev", "spec_file": str(sp), "dw_ids": ["DW-1"]}
+
+    assert verify._canonical_commit_oid(project.project, claimed_ref) is None
+    out = verify.verify_dev_bundle(task, project, rj)
+    assert not out.ok and "does not match" in out.reason
+
+
+def test_verify_dev_bundle_single_char_ref_baseline_is_refused(project):
+    """A ref short enough to be a prefix of its own target defeats any test that
+    reasons about spelling: a branch named for its target's first character does
+    begin with its own name. The claim is refused because it cannot be
+    canonicalized at all — one character is below ``_OBJECT_ID``'s floor and
+    below the four ``rev-parse --disambiguate`` requires — so the ref namespace
+    is never consulted, and the self-prefix property never gets a chance to
+    matter.
+
+    Ablation: bypass ``_canonical_commit_oid`` in the ``allow_ancestor_baseline``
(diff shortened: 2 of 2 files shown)


pbean:
@Polloinfilzato hrmm on review with codex it determined this PR is superseeded by the first pr 645 that i merged. but the pinning requirement and adversarial cases are useful. It wants me to close this PR but I think it missed what your trying to do and im hoping to save time working on the massive issue list by letting contributors help where they can, so im not going to go through the code myself at this point yet.  go ahead and please rebase now that 645 is in and structure the pr for what changes you intended that way i can have it take another look at it or myself if it still wants to drop it once I have some free time.

pbean:
@Polloinfilzato On the 18th I asked you to rebase this and restructure it around what you intended. Two days later I went ahead and did it myself instead of waiting — that was my call, not a sign your rebase would have been unwelcome, and you're owed the reasoning rather than a branch that silently changed under you.

What changed my mind was finally reading the code properly. A rebase would have cost you real work on a diff that mostly doesn't survive contact with current `main`, and I didn't want to hand that back to you as a conflict to resolve. So I reduced the branch to the part that still earns its place and pushed it via `maintainerCanModify`. **Your authorship is preserved on the commit** — this still merges as your PR. Full reasoning below so you can push back on any of it.

## The production change is already on `main` — it arrived through #645

#645 merged on 2026-08-18 and absorbed this requirement during its review rounds. Your commit `83d81da4` *"require an object id, not a revision expression"* is on that branch, and the `_OBJECT_ID` regex it added still stands verbatim on `main` at `verify.py:347`. The canonicalization built around it (`_canonical_commit_oid`, `verify.py:350`) came from the later review commits on that same PR.

The upshot is that every refusal `names_its_own_commit()` was written to add is already enforced:

| #647 refuses | What refuses it on `main` today | Test on `main` |
|---|---|---|
| symbolic `HEAD` | `_OBJECT_ID` (`verify.py:347`) — a non-hex claim never reaches git | `test_verify_dev_refuses_a_symbolic_baseline_revision` (`:1118`) |
| all-hex ref `abcdef0` | `rev-parse --disambiguate=<claim>` (`verify.py:364`) never consults the ref namespace | `test_verify_dev_refuses_an_all_hex_ref_baseline` (`:1142`, branch + tag) |
| single-char ref | the same regex — one character is below the length floor | *no dedicated test; structural only* |

Two ways `main`'s shape came out stronger than this PR's:

- **`--disambiguate` answers the ref question and the abbreviation question in a single invocation.** Your docstring is candid that the two `rev-parse` calls leave a window where a writer creating a matching ref *between* them isn't seen. That window doesn't exist on `main` — the ref namespace is never consulted at all, so there's nothing to race.
- **Canonicalization runs first.** The claim is canonicalized at `:2176`, *before* the equality check (`:2184`) and before both the `allow_ancestor_baseline` leg (`:2193`) and the `newer_ok` leg (`:2201`); both legs consume the canonical oid, never the raw claim. In #647 the prefix-tolerant `same_commit(...)` still ran ahead of `names_its_own_commit` — the hole codex flagged. A git fault also escalates typed at `:2178` rather than being swallowed to `False`.

## The one place we go the other way: sub-7 abbreviations

#647 deliberately accepts them; `main` deliberately refuses them; I'm keeping `main`'s direction. Your PR body argues that *"a `--short` abbreviation below git's default is legitimate and is accepted today"* — I checked that carefully, because it's the crux of the disagreement, and the premise doesn't hold.

With `core.abbrev` unset git uses `auto`, which scales the abbreviation with repository size and clamps **upward** to 7 for small repos — it never returns fewer than 7. This repository already yields **8** (`git rev-parse --short HEAD` → `78c3a748`). Getting below 7 requires explicitly setting `core.abbrev` or `--short=` to 4–6 (4 is git's hard minimum), and even then git lengthens the result whenever uniqueness demands it.

So the floor doesn't discard legitimate `--short` output — nothing git produces by default falls below it. And the stamp isn't `--short` output to begin with: it's `git rev-parse HEAD` by contract (`verify.py:344`), so the floor costs a well-behaved session nothing, while accepting 4–6 characters would re-admit the prefix collisions the gate can't distinguish from drift.

## What's kept, and why it earns its place

The branch is now **one test-only commit** (`tests/test_verify.py`, +130): four tests / five cases, all on the **bundle path** (`verify_dev_bundle`, `allow_ancestor_baseline=True`). Every refusal test on `main` exercises the dev path, so the ordering invariant — canonicalize *before* the ancestor leg — is currently untested.

That's measured, not asserted. Each test carries an ablation record naming the mutation that reddens it, and both mutations were run against the full suite:

- **Restore the pre-#645 shape** (ancestor leg consumes the raw claim ahead of canonicalization) → **5 failed, 6113 passed**: exactly the five new cases. Every existing dev-path refusal test stays green, because none of them enter the relaxation leg.
- **Relax `_OBJECT_ID`'s floor from 7 to 4** → **1 failed, 6117 passed**: the floor test alone. Nothing else in the suite pins that boundary.

Two of the cases also close boundaries `main` has no test for at all — the single-character ref and the sub-7 abbreviation. Both are enforced structurally by the regex today, with nothing asserting it.

Two changes I made to your assertions, both worth flagging:

1. `assert "baseline" in out.reason` → `assert "does not match" in out.reason`. The proof-of-work refusal (*"no changes in worktree since baseline commit"*) also contains the substring `baseline`, so under a weakening that lets the claim reach that leg, the original assertion passes for the wrong reason — a false green. The tightened form matches `main`'s own dev-path tests.
2. Each test now asserts its fixture is genuinely dangerous before asserting the refusal (the ref resolves; `--disambiguate` comes back empty; `is_ancestor(claim, baseline)` is true; the 6-character claim *is* resolvable by git). Without those, a test can pass because the fixture was inert rather than because the gate fired.

No changelog entry — test-only changes don't take one.

The adversarial fixtures are the part of #647 that `main` didn't have, and they're why this is landing as your PR instead of being closed. Thanks for chasing those designs down — the one-character ref in particular is a nice trap. If you'd rather I dropped any of it, say so.

pbean:
@codex review

pbean:
@coderabbitai full review

pbean:
Correction to my comment above, on one point — the ablation record.

I labelled the mutation that reddens the five new cases *"restore the pre-#645 shape"*. That label is wrong. The measurement it carries — 5 failed / 6113 passed, exactly the five new cases, every existing dev-path refusal test green — belongs to a narrower mutation: bypass `_canonical_commit_oid` in the `allow_ancestor_baseline` leg only, leaving the newer-baseline leg canonical-gated.

A genuine pre-#645 restore is strictly larger, because it strips canonicalization from the newer-baseline leg as well. Under *that* mutation `test_verify_dev_refuses_an_all_hex_ref_baseline` reddens on both parameters: a raw ref claim reaches `commit_reachable_above_baseline`, which carries no opt-in flag, and a hex-named ref pointing above the baseline buys the gate. So "every existing dev-path refusal test stays green" is true of the mutation I ran, not of the one I named.

The conclusion doesn't move — the five cases are additive, and the reason they're additive is unchanged. But the branch's docstrings now name the mutation precisely instead of by that label, and I'd rather correct the comment than leave it disagreeing with the code it describes.

Two related changes on the branch while I was at it, both since your last look: each test now asserts `_canonical_commit_oid(...) is None` directly, because the canonical-is-None leg and the ancestor leg emit byte-identical reason strings and the old substring assertion couldn't tell which one fired — the same false-green shape I flagged in your `assert "baseline" in out.reason`. And there's now a second commit, mine not yours, correcting `same_commit`'s docstring: it justified the 7-character floor as "git's default `--short` length", which is the same mistaken premise your PR body and my comment above both leaned on, sitting in production source since well before either. That's #674.

pbean:
@codex review

Polloinfilzato:
Apologies for the long silence — I finally took a few days off and am only now caught up on this thread.

Thank you for doing the rebase yourself rather than handing back a conflict, and for keeping my authorship on the commit. Both were the right call, and writing out the full reasoning was more than I would have asked for.

On the sub-7 abbreviation: you were right and my PR body was wrong. I had assumed a `--short` abbreviation below 7 was legitimate output; with `core.abbrev` unset git clamps *upward* to 7, and this repo already yields 8. Since the stamp is `git rev-parse HEAD` by contract, the floor costs a well-behaved session nothing. No pushback from me — keeping `main`'s direction is correct, and #674 fixes the premise both of us had leaned on.

Your correction to the ablation label is noted as well: the mutation I measured was the narrower one, and your restatement is what the numbers actually came from.

On #640 — I had offered to take the ancestor half; I see #744 landed it and the issue is closed, so there is nothing owed there.

Picking things back up now. I would like to take #624 unless someone is already on it, and I have one plugin-loader finding I will file separately.

PR #657, opened 2026-08-19
ci: bound and retry the tmux install instead of wedging the job

pbean:
## The bug

The Linux `test` job's tmux step guarded `apt-get update` with `|| true`. That
handles a non-zero exit. The failure actually hitting the job is the other one:
**`apt-get update` does not exit at all**, and `|| true` has no exit status to
discard.

Measured on run `32230727567`, job `test (py3.11)` — from the job log:

```
08:04:27 Get:5 https://archive.ubuntu.com/ubuntu noble-security InRelease [126 kB]
08:18:54 ##[error]The operation was canceled.
```

**No output for 14 minutes** until the 15-minute job cap killed it. `Install the
project` and `Run tests` never started, so the code under review never ran — the
red was uninformative. It wedged that way four times in a row. apt's own
`Acquire::*::Timeout` (default 120s) did **not** fire during the stall, so a tuned
apt option would not have helped either.

## The finding that changed the fix

The first iteration of this PR added bounds and retries around the apt calls. Its
own CI run then answered a question the PR had not thought to ask:

```
tmux is already the newest version (3.4-1ubuntu0.1).
0 upgraded, 0 newly installed, 0 to remove and 1 not upgraded.
```

**The ubuntu-24.04 runner image already ships tmux.** This step has never
installed anything. It was taking the full risk of the network for a no-op — and
that risk is what kept killing the job. The retry loop made that survivable, but
it also made every Linux leg pay a 60s timeout (measured on all four legs) to
discover there was nothing to do.

## The fix

Ask first. `tmux -V` up front short-circuits the entire step, so the normal path
touches the network zero times. apt becomes a **fallback** for an image that stops
shipping tmux — announced with a `::warning::`, bounded by `timeout` (the bound
that actually holds here), and retried once. The closing `tmux -V` is still the
assertion, so a genuine absence fails the job loudly rather than silently skipping
the tmux integration tests.

Also bounds `dpkg --configure -a`, and sizes the step cap **above** the fallback's
worst case rather than under it: with `--kill-after`, the retry loop's worst case
was 430s against a 420s cap, so the final `tmux -V` could have been cancelled
instead of reached. It is now 260s against a 360s cap.

## Verification

`trunk check` (actionlint/yamllint/prettier) clean. The `run:` script was extracted
verbatim from the YAML and executed under `bash -e` against stubbed
`sudo`/`apt-get`/`dpkg`/`tmux`:

| scenario | result |
| --- | --- |
| tmux present (the real runner) | `rc=0`, 0s, **apt never invoked** — empty call trace |
| tmux absent, apt healthy | `rc=0`, falls back, one `update` + one `install` |
| tmux absent, install fails once | `rc=0`, retries with bounded `dpkg --configure -a`, succeeds |
| tmux absent, install always fails | 2 attempts, then `tmux -V` fails → **job fails loudly** |
| tmux absent, `update` hangs (`sleep 600`) | `rc=0` at 45s — `timeout` killed the hang, install proceeded |

Ablation, the same hang against the **original** step: it never returns. The trace
stops at `update`, `install` is never reached, and an external 25s bound had to
kill it — the CI failure above, reproduced and then closed.

Workflow-only change; no CHANGELOG entry, matching the existing `ci:` commits.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,51 +33,51 @@ jobs:
           enable-cache: true
           python-version: ${{ matrix.python-version }}
 
-      - name: Install tmux (for generic_tmux integration tests)
-        # Bounds the step even if every attempt below stalls: the job cap is 15
-        # minutes and a wedged install used to consume all of it.
-        timeout-minutes: 7
+      - name: Ensure tmux (for generic_tmux integration tests)
+        # Bounded above the fallback's worst case (~260s) so the loop always
+        # reaches the final `tmux -V` instead of being cancelled mid-attempt.
+        timeout-minutes: 6
         run: |
-          # Two distinct failure modes, both measured on this runner fleet.
+          # The ubuntu-24.04 runner image already ships tmux. Measured on this
+          # workflow's own run: "tmux is already the newest version
+          # (3.4-1ubuntu0.1) ... 0 upgraded, 0 newly installed". The apt work this
+          # step used to do unconditionally therefore installed nothing, while
+          # taking the full risk of the network to do it:
           #
-          # 1. `apt-get update` EXITS NON-ZERO when a third-party repo baked into
-          #    the runner image (e.g. dl.google.com/chrome) serves a stale index
-          #    (Hash Sum mismatch). Harmless here — the Ubuntu archive lists still
-          #    refresh — so its status is discarded.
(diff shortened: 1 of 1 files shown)


pbean:
@codex review

pbean:
@codex review

PR #672, opened 2026-08-20
fix(psmux): pin display-message self-probes to the calling pane

### LINKED ISSUE #669: psmux: current_window_id/current_pane_id answer for the server's active window, not the calling pane
`BaseTmuxBackend._display_message` (`src/bmad_loop/adapters/tmux_base.py:530`) runs `display-message -p <fmt>` with **no `-t`**. On psmux a target-less probe resolves the *server's active window*, not the pane the calling process runs in — so `current_window_id()`, `current_pane_id()` and therefore `current_return_target()` answer for a **foreign window** whenever the caller's window is not the active one.

This is already known at the seam, in one place: `PsmuxMultiplexer._parked_trailer` pins its own probe with `-t $env:TMUX_PANE` and the comment says why — *"a target-less probe would resolve the server's active window, which is another window's key the moment focus moves"* (`src/bmad_loop/adapters/psmux_backend.py:136`). The pwsh trailer got that treatment; the Python probes never did.

## Measured

`psmux 3.3.8 (66cf613)`, Windows 11. A probe process running in window `@8` / pane `%10` (`$TMUX_PANE=%10`) while a **different** window (`@3`) was active:

| probe | answered | truth |
| --- | --- | --- |
| `current_window_id()` | `bl659-ctl:@3` | `@8` |
| `current_pane_id()` | `%4` | `%10` |
| `current_return_target()` | `=bl659-ctl:%4` | `=bl659-ctl:%10` |

Reproduced a second time with a real attached client sitting in yet another window: the probe answered for the window holding the client, never for its own.

## Why it matters

- **`_ctl_window_candidates` can kill the window it is running in.** It excludes its own window with `win_id == current` (`src/bmad_loop/tui/launch.py:660`), and the seam docstring states the contract plainly: *"Must match `list_windows`' `window_id` form — the ctl prune skips its own window by comparing them, so a mismatch makes a prune run from inside a ctl window kill that window."* On psmux the comparison is against whichever window happens to be active, so a prune triggered from a non-active ctl window does not exclude itself; the exclusion lands on an unrelated window instead.
- **`return_attached_client` reads and clears the return option on the wrong window** (`src/bmad_loop/tui/launch.py:589`): it can replay another window's recorded return target, and it clears `@bmad_return_pane` on a window that never had a hand-back to do.
- **`current_return_target()` records a foreign pane at attach time** (`src/bmad_loop/cli.py:3142`, `src/bmad_loop/tui/app.py:453`), so the return move is aimed at a pane the human never came from.

The unattended paths are the exposed ones: a sweep's window is active only while someone is looking at it, and several runs share the control session.

`current_session()` is unaffected in practice — one server per session means the active window's session is this window's session too.

## Fix direction

Override `_display_message` in the psmux leaf to pin the probe to the caller's own pane, exactly as `_parked_trailer` already does:

```
display-message -p -t $env:TMUX_PANE <fmt>
```

psmux sets `TMUX_PANE` in every pane, and a bare `%N` target resolves globally via `DisplayMessageById`. Keep the existing `TMUX` guard for "not inside the multiplexer", and degrade to the current behaviour (or to `None`) when `TMUX_PANE` is unset rather than silently probing the active window.

Not verified on tmux — the equivalent question there (does a target-less `display-message` follow the calling pane or the current client?) should be measured before assuming the base needs changing rather than the leaf.

Found while running the attended-console pass for #659; unrelated to that issue's verdict-source change, which is why it is filed separately.

dracic:
A target-less `display-message -p` on psmux resolves the **server's active window**, not the calling pane, so `current_window_id()`, `current_pane_id()`, `current_session()` and `current_return_target()` answered for a foreign window whenever the caller's window was not the focused one. The ctl prune's own-window exclusion could land on an unrelated window (a prune run from a non-active ctl window could kill the window it was running in), and the attach return could record and replay a pane the human never came from.

## Fix

- `PsmuxMultiplexer` now overrides `_display_message` to pin the probe with `-t $TMUX_PANE`, exactly as the pwsh parked trailer already did — one chokepoint, all four inherited probes fixed. The tmux base is untouched (its behavior for the target-less form is unmeasured).
- Fail closed instead of answering for a foreign window: `None` without spawning when `TMUX_PANE` is absent, and when it holds a non-`%N`-shaped value — measured on psmux 3.3.8: `@N`-shaped values and a live session name resolve rc=0 to the *active* window, while plain garbage already exits nonzero.

## Tests

- Unit: pinned argv asserted exactly (single spawn); three refusal paths (`TMUX` unset, `TMUX_PANE` unset, non-pane-shaped) each proven ablation-red — `None` with zero subprocess calls.
- Live gate (real psmux 3.3.8, zero tokens): a new `test_premise_*` probe observes the founding divergence itself — it goes red the day psmux starts honoring `TMUX_PANE` for target-less probes, signalling the pin is droppable — and a new `test_adopted_*` probe covers the global bare-`%N` resolution (`DisplayMessageById`) the pin rides on. 11/11 passing on a Windows host.
- Smoke-verified end to end with a real attached client: with the client focused on another window, the fixed probes answer the caller's own window/pane; the target-less form still answers the active window.

Closes #669

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved psmux probing to target the calling pane reliably.
  - Prevented unpinned queries when pane context is unavailable or invalid.
  - Avoided incorrect results caused by changes to the active window.

- **Tests**
  - Added coverage for pane-targeted queries, missing context, invalid pane identifiers, and execution failures.
  - Added live checks for behavior across window focus changes.

- **Documentation**
  - Documented the psmux probe targeting fix in the unreleased changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,12 +27,13 @@ breaking changes may land in a minor release.
   the return option survives, but the sweep goes unattended instead of blocking a later `--repeat`
   cycle on `input()`. `False` now carries the joint claim its caller always read it as. Out-of-tree
   backends answering a plain bool still work; `detach_client` is unchanged.
-- psmux probes (`current_window_id`, `current_pane_id`, `current_session`,
-  `current_return_target`) now
-  pin `display-message` to the calling pane via `TMUX_PANE`; a target-less probe
-  answered for the server's _active_ window, so a caller in a non-active window could
-  prune its own window or record another window's return target (#669). With `TMUX`
-  set but `TMUX_PANE` unset the probe returns `None` instead of running unpinned.
+- **A psmux probe now answers for the calling pane, not the focused window (#669).** A target-less
+  `display-message -p` resolves the server's _active_ window, so `current_window_id`,
+  `current_pane_id`, `current_session` and `current_return_target` answered for a foreign window
+  whenever the caller's own window was not the focused one — the ctl prune could kill the window it
+  was running in, and the attach return could record a pane the human never came from. The probes
+  now pin to the calling pane via `TMUX_PANE`, and answer `None` without spawning when that value
+  is absent or not pane-shaped.
 
 ## [0.11.0] — 2026-08-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ breaking changes may land in a minor release.
 
 ## [Unreleased]
 
+### Fixed
+
+- psmux probes (`current_window_id`, `current_pane_id`, `current_session`,
+  `current_return_target`) now
+  pin `display-message` to the calling pane via `TMUX_PANE`; a target-less probe
+  answered for the server's _active_ window, so a caller in a non-active window could
+  prune its own window or record another window's return target (#669). With `TMUX`
+  set but `TMUX_PANE` unset the probe returns `None` instead of running unpinned.
+
 ## [0.11.0] — 2026-08-19
 
 ### Added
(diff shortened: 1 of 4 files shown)


dracic:
@codex review

Issue #474, opened 2026-08-06
policy: a wrongly-typed policy.toml field escapes as a bare ValueError, so validate --json emits no document


pbean:
**Measured on `main` (a4b9ad8) and on `feat/384-validate-exclude-pollution`, 2026-08-06.** Not introduced by #386 — the `cmd_validate` policy block and `policy.py`'s conversions are byte-identical to `main`'s.

A `.bmad-loop/policy.toml` field that is TOML-legal but the wrong *type* escapes `policy.load` as a bare `ValueError` instead of `PolicyError`:

```console
$ mkdir -p proj/.bmad-loop
$ printf '[limits]\ngit_timeout_s = "x"\n' > proj/.bmad-loop/policy.toml
$ bmad-loop validate --project proj --json
error: invalid literal for int() with base 10: 'x'
$ echo $?
1
```

**stdout is empty** — no JSON document at all, so the schema-versioned one-object `--json` contract (AGENTS.md) is broken, and nothing names the file or the key at fault. The contrast is the point: the same project with a *valid* policy that fails every other check still emits the whole document at rc 1 (`"ok": false`) — "a failing check still emits the whole document, at exit 1" is the documented behavior in docs/FEATURES.md.

Reproduced with four fields; the shape is generic, not per-key:

| policy.toml | stderr |
|---|---|
| `[limits] git_timeout_s = "x"` | `error: invalid literal for int() with base 10: 'x'` |
| `[limits] cache_read_weight = "x"` | `error: could not convert string to float: 'x'` |
| `[scm] max_parallel = "x"` | `error: invalid literal for int() with base 10: 'x'` |
| `[cleanup] run_retention = "x"` | `error: invalid literal for int() with base 10: 'x'` |

## Mechanism

`policy.loads` coerces ~25 numeric knobs with raw `int()`/`float()` (`policy.py` — `limits_d`, `sweep_d`, `scm_d`, `cleanup_d` blocks). Those raise bare `ValueError`/`TypeError` on a wrong-typed value, and `load`'s guard converts only `PolicyError` (plus `UnicodeDecodeError` at the read). Every consumer keys its fault handling on `PolicyError`, so the escape lands wherever the caller has no backstop — in `validate`'s case `except policy_mod.PolicyError` one frame up, then `main`'s bare-except printing one `error:` line.

The blast radius is wider than `validate`, and `policy.load`'s own docstring already states the rule this misses:

[...]

PR #688, opened 2026-08-22
fix(policy,adapters,plugins): typed errors for every user-authored TOML read (#440, #473)

### LINKED ISSUE #440: policy.loads raises raw ValueError/TypeError from its int()/float() coercions, escaping every (PolicyError, OSError) degrade handler
`policy.loads` raises raw `ValueError`/`TypeError` from its `int()`/`float()` coercions, escaping the same `except (PolicyError, OSError)` degrade handlers that exist to survive a bad policy file.

Found while reviewing #439 (6C of the #433 forward-port), which closed the neighbouring
`UnicodeDecodeError` leg of the same class. **Pre-existing on `main`, not introduced by that PR** —
filing rather than widening a forward-port sub-phase.

## The gap

`policy.loads` validates some `[limits]` fields explicitly (`policy.py:757` raises a typed
`PolicyError` for `session_budget_grace_s`) but constructs the rest by bare coercion at
`policy.py:762` onward:

```python
limits = LimitsPolicy(
    max_review_cycles=int(limits_d.get("max_review_cycles", LimitsPolicy.max_review_cycles)),
    ...
    cache_read_weight=float(...),
)
```

`int()` and `float()` raise `ValueError`/`TypeError`, neither of which is a `PolicyError` or an
`OSError`. Measured on `main`:

```
[limits] max_review_cycles = "abc"   -> ValueError: invalid literal for int() with base 10: 'abc'
[limits] cache_read_weight  = "x"    -> ValueError: could not convert string to float: 'x'
[limits] max_review_cycles = [1, 2]  -> TypeError: int() argument must be a string, a bytes-like
                                        object or a real number, not 'list'
```

## Why it matters

Three handlers exist specifically to degrade to defaults rather than take the process down, and all
three catch exactly `(PolicyError, OSError)`:

- `cli.py:152` — `_configure_mux`, which runs at `cli.py:3099` immediately before
  `args.func(args)`, i.e. on **every** subcommand. Its own comment (`cli.py:143-146`) promises
  diagnostics keep working on a misconfigured host.
- `tui/app.py:229` — `_stories_defaults`.
- `tui/screens/dashboard.py:188` — `DashboardScreen.__init__`.

Concrete failure: a project whose `.bmad-loop/policy.toml` carries
`[limits]\nmax_review_cycles = "abc"`. `BmadLoopApp.__init__` (`tui/app.py:149`) eagerly constructs
`DashboardScreen`, whose `__init__` is the first policy load; the `ValueError` walks straight past
`(PolicyError, OSError)` and the TUI dies at construction with a traceback, instead of falling back
to defaults and letting the operator open the settings editor to fix the file.

On the CLI the blast radius is smaller — `main`'s bare `except Exception` (`cli.py:3123`) still
returns `ExitCode.FAILURE` — but the operator gets an untyped `error:` line naming an `int()`
conversion rather than the offending key, and `_configure_mux`'s degrade contract is defeated.

Note `policy.load` prefixes `PolicyError` with the file path (`policy.py:690`); a raw `ValueError`
escapes that wrapper too, so the message does not even name the file.

## Fix shape

Coerce through a typed helper that names the key, in the spirit of the existing
`_opt_*`/`_tui_dim` validators (`policy.py:620`), so every `[limits]` field reports as
`limits.<key> must be an integer: got <repr>` — matching what `session_budget_grace_s` already
does two statements above. The float field (`cache_read_weight`) needs the same treatment.

Worth sweeping the other sections for the same shape while in there rather than fixing `[limits]`
alone — the bug is the bare coercion, not the section.

## Test

The regression test belongs in `tests/test_policy.py` next to `test_adapter_timing_knobs_reject_negatives`
(`tests/test_policy.py:288`), which already parametrizes bad-value → `PolicyError` for a neighbouring
knob group. Pin the *type*, not just the message: the whole point is that the exception is catchable
by the three degrade handlers.

### LINKED ISSUE #473: validate: a non-UTF-8 profile overlay crashes the command instead of reporting adapter.profile
**Measured** (2026-08-06). Originally measured on `feat/384-validate-exclude-pollution` @ `5065190`; PR #386 was closed unmerged, so re-anchor any repro on `main` — the code paths named below are untouched by that branch and the fault reproduces identically:

```
mkdir -p proj/.bmad-loop/profiles
printf '[adapter]\nname = "claude"\nmodel = "opus"\n' > proj/.bmad-loop/policy.toml
printf 'name = "b\xffad"\n' > proj/.bmad-loop/profiles/bad.toml
bmad-loop validate --project proj --json
# stderr: error: 'utf-8' codec can't decode byte 0xff in position 9: invalid start byte
# stdout: (nothing) — rc 1
```

No document at all: `main`'s bare-except backstop prints one `error:` line, so the `--json` one-object contract (AGENTS.md) is broken and nothing names the file at fault.

**Mechanism.** `load_profiles` reads each overlay with `read_text(encoding="utf-8")`; a non-UTF-8 file raises `UnicodeDecodeError` — a `ValueError`, not a `ProfileError` — and `cmd_validate`'s role loop catches only `ProfileError`.

**Precedent.** `policy.load` converts exactly this fault into its domain error so validate reports `policy` with "not valid UTF-8" instead of crashing (`test_validate_reports_an_undecodable_policy_instead_of_crashing`, #205 family). The profile loader should convert the same way — and since `get_profile` backs every adapter resolution, the crash surface is not just validate: run/sweep preflights call the same loader.

**Fix shape.** Convert `UnicodeDecodeError` to `ProfileError` at the profile read (naming the file), mirroring `policy.load`; a validate-level test through `machine_json` pins the document contract, per the policy twin.

⚠️ **Correction (2026-08-06).** The Mechanism paragraph previously ended with "(The #384 candidate assembly is unaffected: `discovered_profiles` guards `UnicodeError` per item and skips the file.)" — deleted, because `discovered_profiles` **does not exist on `main`**. It was introduced by `69eb6c8` on the closed PR #386 branch and died with it. There is no such call site to reason about.

### LINKED ISSUE #689: tui settings: a non-UTF-8 project plugin.toml crashes the settings screen instead of reporting PluginError
**Measured** (2026-08-22) on `main` @ `4bf939ee`, with a project plugin folder-dropped at
`proj/.bmad-loop/plugins/bad/plugin.toml` written as the bytes
`b'[plugin]\nname = "b\xffad"\napi_version = 1\n'` (valid TOML, but not valid UTF-8):

```python
from pathlib import Path
from bmad_loop.tui.settings import PolicyDoc

proj = Path("proj")
doc = PolicyDoc.load(proj / "missing.toml")  # template-backed, valid
doc.validate()                # -> None: the POLICY is fine, only the plugin read is not
doc.validate(project=proj)    # -> raises
```

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 18: invalid start byte
  tui/settings.py:98        PluginRegistry.build(project, pol).validate(pol)
  plugins/registry.py:131   manifests = load_plugins(project, journal=journal)
  plugins/loader.py:96      for manifest in discover(project):
  plugins/loader.py:85      yield from _discover_project(project)
  plugins/loader.py:72      toml.read_text(encoding="utf-8"), ...
```

**Mechanism.** `_discover_project` reads each project `plugin.toml` with
`read_text(encoding="utf-8")` (`plugins/loader.py:72`); a non-UTF-8 file raises
`UnicodeDecodeError` — a `ValueError`, not a `PluginError`. The settings screen degrades on
`except (policy_mod.PolicyError, PluginError)` (`tui/settings.py:99`), so the raw error walks
past it and that surface dies at construction instead of rendering a message naming the file
to fix. `PluginRegistry.build` loads **every discovered** manifest, not just those in
`plugins.enabled`, so dropping the folder in is enough to arm it — no config change needed.

The `OSError` arm is the same escape and was measured the same way: `chmod 000` on an otherwise
valid manifest (as an unprivileged user, so the mode is genuinely enforced) raises
`PermissionError` out of the same call, likewise outside the tuple. The existing
`is_dir()`/`is_file()` checks rule out absence and nothing else, so present-but-unreadable
(permissions, a dead mount) reaches the read.

The packaged built-ins read the same way (`plugins/loader.py:46`). They are trusted, but a
corrupt install is a packaging bug and should say so rather than surface as a decode traceback.

**Scope note.** `bmad-loop validate --project ... --json` does **not** reach this path — it was
re-measured on `main` and still prints a complete document with empty stderr. The crash surface
is the settings screen and any other `PolicyDoc.validate(project=...)` caller. That is what
distinguishes this from #473, whose repro and contract are both the `validate` command.

**Precedent.** `policy.load` converts exactly this fault into `PolicyError`; #473 does the same
for profile overlays via `ProfileError`. `plugins/loader.py` is the third and last parser that
reads user-authored TOML, and the only one still without the conversion.

**Fix shape.** Convert `UnicodeDecodeError` and `OSError` into `PluginError` naming the path, at
**both** read sites (builtin `:46` and project `:72`), mirroring #473's profile guard.
`plugins/manifest.py` needs no change — its value-coercion funnel already converts; the gap is
strictly the unguarded read.

The fix is carried by PR #688, which bundles the same conversion for the policy (#440) and
profile (#473) parsers.

pbean:
## The problem

Three separate parsers read user-authored TOML — `policy.loads`, `adapters/profile.load_profiles`,
and `plugins/loader` — and each has a typed domain error (`PolicyError`, `ProfileError`,
`PluginError`) that every consumer keys its fault handling on. But a file whose *content* was
wrong escaped as a **raw** `ValueError` / `TypeError` / `UnicodeDecodeError` instead, walking
straight past the handlers written to survive exactly that:

- `cli.py`'s `_configure_mux` — `except (PolicyError, OSError)`, and it runs before dispatch on
  **every** subcommand, so one bad character traced back at you from an unrelated command.
- `tui/app.py` and `tui/screens/dashboard.py` — the TUI died at construction rather than falling
  back to defaults and letting the operator open the settings editor to fix the file.
- `tui/settings.py` — `except (PolicyError, PluginError)`, the degrade that renders the message
  into the settings screen's error line.
- `cmd_validate`'s role loop — the raw escape hit `main`'s bare-except backstop, which prints one
  `error:` line and **no document**, breaking `machine.py`'s one-object `--json` contract.

Two of the policy faults never announced themselves at all: `notify.desktop = "false"` is a truthy
string, so `bool()` turned the feature **on**; `verify.commands = "pytest"` iterated a bare string
into six one-character commands that read back as applied configuration.

## What changed, phase by phase

| # | Commit | Change |
|---|--------|--------|
| 1 | `d2e728c6`, `025d57ca` | **policy crash-class.** The four `[limits]`-only helpers generalized to section-aware `_typed_int` / `_typed_float` / `_typed_bool` / `_typed_str(d, where, key, default)` (+ `_typed_str_tuple`); every remaining raw `int()`/`float()` in `loads()` routed through them; `_opt_grace`/`_opt_nudges` given inline type guards; array shape guards on `verify.commands` and both `extra_args`. |
| 2 | `cda395d2` | **policy silent-class.** Every remaining bare `bool()`/`str()` routed through the typed helpers — 16 boolean fields, 22 string fields, the `plugins.enabled` entries, and the deprecated-`[engine]` fold. Allowlist fields keep their allowlist and blame the *type* first. |
| 3 | `bcddfed9` | **profile read guard.** `_read_profile_text(entry, source)` at both `load_profiles` read sites (packaged + overlay), converting `UnicodeDecodeError`/`OSError` into `ProfileError` naming the file. |
| 4 | `66d6e2e8` | **plugin loader read guard + ship (#689).** The same `_read_manifest_text(toml, source)` at both `plugins/loader` read sites (builtin + project); CHANGELOG. |

`[limits]` messages are byte-identical throughout — all call sites pass `where="limits"`. Valid
policies parse to identical `Policy` objects: this changes only what happens to *invalid* values.
`usage_grace_s = 30` is still a legal float, and an unset `extra_args` still means "inherit the
profile's flags" where `[]` means "none". The lenient snapshot readers (`_snapshot_extra_args`,
`adapter_policy_from_snapshot`) are deliberately untouched — they parse the json-round-tripped
`RunState.policy_snapshot` and must never crash; `025d57ca` documents why.

`plugins/manifest.py` needed no work: its value-coercion funnel already exists and its messages
are pinned by tests. The live plugin gap was the unguarded read, not the parse.

## Repros, re-measured on this branch

| Repro | Before | After |
|-------|--------|-------|
| `[scm] max_parallel = "x"` | raw `ValueError` | `PolicyError: scm.max_parallel must be an integer: got 'x'` |
| `[cleanup] run_retention = "x"` | raw `ValueError` | `PolicyError: cleanup.run_retention must be an integer: got 'x'` |
| `[limits] dev_contract_nudge = "false"` | typed already | unchanged, byte-identical (unregressed) |
| `[notify] desktop = "false"` | silently ON | `PolicyError: notify.desktop must be a boolean: got 'false'` |
| non-UTF-8 profile overlay, `validate --project proj --json` | bare `error:` line, **empty stdout**, rc 1 | **one** JSON document at rc 1 with an `adapter.profile` finding naming the file; stderr empty |
| non-UTF-8 project `plugin.toml` | raw `UnicodeDecodeError` through the TUI settings degrade | `PluginError: plugin <path>: not valid UTF-8: …`, settings surface degrades |

## Verification

- `uv run pytest -q -n logical` — **6397 passed, 53 skipped** (6225 before phase 1; +172 cases).
- `uv run pyright` — 0 errors. `trunk fmt` + `trunk check --all` (258 files) — clean.
- Every new gate was ablated **singly** and confirmed to redden, restored from a `cp` backup with
  an md5 check rather than `git checkout`. Where a fix had two axes, both were measured and the
  redden-sets are **disjoint** — which is what proves neither half stands in for the other:
  - *call-site axis* (phase 4): reverting the project read reddens 2 tests, the builtin read 1.
  - *arm axis* (phase 4): dropping the `UnicodeDecodeError` arm reddens 1, the `OSError` arm 2.
  - phase 2's wiring axis: reverting one call site (`scm.keep_failed`) to a bare `bool()` with
    both helpers intact reddens exactly 3 rows while 62 sibling rows stay green.
- The validate tests assert on the parsed `machine_json` document, never on rc alone: `main`'s
  bare-except backstop also returns 1, so an rc-only assertion is green with the fix reverted.

Closes #440. Closes #473. Closes #689.

#689 was filed for the plugin-manifest leg after review: #473's repro and contract are both
the `validate` command, which never reaches the plugin loader, so that leg had no issue of its
own even though it is the same fault at the third and last parser reading user-authored TOML.
Filed rather than split out — the guard is 37 lines in `loader.py` mirroring the profile guard,
and shipping two of three parsers fixed would leave a known crash on `main`.

Context, no status change intended: #587 (the earlier `[limits]` leg this extends), #474
(a duplicate of #440, contributing the validate-document carry test), #278 (where the
truthy-string hazard was recorded).

## Summary by CodeRabbit

- **Bug Fixes**
  - Policy validation now rejects incorrectly typed TOML values with clear, field-specific errors instead of coercing them.
  - Unreadable or invalid-encoding profile files and plugin manifests now produce structured, file-specific diagnostics.
  - Dashboard startup gracefully falls back to default settings when unrelated policy values are invalid.
- **Validation**
  - Expanded coverage for policy fields, profiles, and plugin manifests.
  - JSON validation output now includes actionable findings and an appropriate failure status.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,43 @@ breaking changes may land in a minor release.
 
 ### Fixed
 
+- **A policy field of the wrong TOML type now raises `PolicyError` naming `section.key`
+  (#440).** `loads()` coerced with bare `int()`/`float()`/`bool()`/`str()` outside the
+  `PolicyError` funnel, so a wrong-typed value escaped as a raw `ValueError` or
+  `TypeError` — past every `except (PolicyError, OSError)` degrade handler in the
+  codebase. `_configure_mux` runs before dispatch on _every_ command, so one bad
+  character in `.bmad-loop/policy.toml` traced back at you instead of being reported;
+  the TUI died at construction rather than falling back to defaults; and `validate --json`
+  exited before printing the document its one-object contract promises, which is now
+  emitted with the fault as an ordinary `policy` finding. #587 had fixed `[limits]` this
+  way; this finishes the sweep across `[verify]`, `[sweep]`, `[scm]`, `[cleanup]`,
+  `[notify]`, `[review]`, `[gates]`, `[stories]`, `[dev]`, `[adapter]` and its stage
+  tables, `[tui]`, `[operator]`, `[mux]`, and `[plugins]`, and `[limits]`' own messages
+  are unchanged. Two of these never announced themselves at all: `notify.desktop = "false"`
+  is a truthy string, so it turned the feature **on** (the hazard #278 recorded), and
+  `verify.commands = "pytest"` iterated a bare string into six one-character commands
+  that read back as applied configuration — both now refuse. Fields with an allowlist
+  (`gates.mode`, `review.trigger`, `mux.backend`, …) keep it and blame the type first
+  rather than reporting a coerced value that was never written. Valid policies parse
+  identically; `usage_grace_s = 30` is still a legal float, and an unset `extra_args`
+  still means "inherit the profile's flags" where `[]` means "none".
+- **An undecodable or unreadable profile overlay or plugin manifest is a typed error
+  naming the file, not a traceback (#473).** `load_profiles` and the plugin loader read
+  each `*.toml` with `read_text(encoding="utf-8")` in the _argument expression_ of the
+  parse call — ahead of the conversion funnel that wraps parsing, so widening that funnel
+  could never have caught it. A non-UTF-8 file therefore escaped as a raw
+  `UnicodeDecodeError` past consumers that all key on the domain error: `validate --json`
+  printed a bare `error:` line and an empty stdout instead of an `adapter.profile`
+  finding, `get_profile` backs every adapter resolution so the same escape reached
+  run/sweep preflight, and a non-UTF-8 project `plugin.toml` walked through the settings
+  screen's `except (PolicyError, PluginError)` and took that surface down at
+  construction. Both loaders now raise `ProfileError`/`PluginError` naming the offending
+  path, and both arms are covered — a file that is present but unreadable (permissions,
+  a dead mount) was leaking a bare `OSError` on the same route, since the existing
+  `is_dir()`/`is_file()` checks rule out absence and nothing else. The packaged built-ins
+  read through the same guard: they are trusted, but a corrupt install is a packaging bug
+  and should say so.
+
 - **Merge-back failures are classified from the measured tree state, and the catch-all stopped
   inventing a conflict (#619).** A merge that died part-way through its checkout — all three
   strategies, `--ff-only` included — now raises `MergeHalfAppliedError` instead of "refused
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,40 +18,28 @@ breaking changes may land in a minor release.
 
 - **A policy field of the wrong TOML type now raises `PolicyError` naming `section.key`
   (#440).** `loads()` coerced with bare `int()`/`float()`/`bool()`/`str()` outside the
-  `PolicyError` funnel, so a wrong-typed value escaped as a raw `ValueError` or
-  `TypeError` — past every `except (PolicyError, OSError)` degrade handler in the
-  codebase. `_configure_mux` runs before dispatch on _every_ command, so one bad
-  character in `.bmad-loop/policy.toml` traced back at you instead of being reported;
-  the TUI died at construction rather than falling back to defaults; and `validate --json`
-  exited before printing the document its one-object contract promises, which is now
-  emitted with the fault as an ordinary `policy` finding. #587 had fixed `[limits]` this
-  way; this finishes the sweep across `[verify]`, `[sweep]`, `[scm]`, `[cleanup]`,
-  `[notify]`, `[review]`, `[gates]`, `[stories]`, `[dev]`, `[adapter]` and its stage
-  tables, `[tui]`, `[operator]`, `[mux]`, and `[plugins]`, and `[limits]`' own messages
-  are unchanged. Two of these never announced themselves at all: `notify.desktop = "false"`
-  is a truthy string, so it turned the feature **on** (the hazard #278 recorded), and
-  `verify.commands = "pytest"` iterated a bare string into six one-character commands
-  that read back as applied configuration — both now refuse. Fields with an allowlist
-  (`gates.mode`, `review.trigger`, `mux.backend`, …) keep it and blame the type first
-  rather than reporting a coerced value that was never written. Valid policies parse
-  identically; `usage_grace_s = 30` is still a legal float, and an unset `extra_args`
-  still means "inherit the profile's flags" where `[]` means "none".
-- **An undecodable or unreadable profile overlay or plugin manifest is a typed error
-  naming the file, not a traceback (#473).** `load_profiles` and the plugin loader read
-  each `*.toml` with `read_text(encoding="utf-8")` in the _argument expression_ of the
-  parse call — ahead of the conversion funnel that wraps parsing, so widening that funnel
-  could never have caught it. A non-UTF-8 file therefore escaped as a raw
-  `UnicodeDecodeError` past consumers that all key on the domain error: `validate --json`
-  printed a bare `error:` line and an empty stdout instead of an `adapter.profile`
-  finding, `get_profile` backs every adapter resolution so the same escape reached
-  run/sweep preflight, and a non-UTF-8 project `plugin.toml` walked through the settings
-  screen's `except (PolicyError, PluginError)` and took that surface down at
-  construction. Both loaders now raise `ProfileError`/`PluginError` naming the offending
-  path, and both arms are covered — a file that is present but unreadable (permissions,
-  a dead mount) was leaking a bare `OSError` on the same route, since the existing
-  `is_dir()`/`is_file()` checks rule out absence and nothing else. The packaged built-ins
-  read through the same guard: they are trusted, but a corrupt install is a packaging bug
-  and should say so.
+  `PolicyError` funnel, so a wrong-typed value escaped every handler written to degrade on
+  one: `_configure_mux` runs before dispatch on _every_ command, so one bad character in
+  `.bmad-loop/policy.toml` traced back at you; the TUI died at construction instead of
+  falling back to defaults; and `validate --json` exited before printing the document its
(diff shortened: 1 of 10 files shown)


pbean:
@codex review

pbean:
@codex review

pbean:
@coderabbitai review

pbean:
@codex review

Issue #690, opened 2026-08-22
Consolidate runs.liveness and runs.engine_liveness into one tri-state reader


pbean:
#650 moved tui.data.liveness into runs.py beside engine_liveness; both wrap probe_liveness over one read_pid_identity and differ only in the no-pid arm (legacy-session fallback vs 'dead') and whether ProcessHostError degrades to 'unknown' or propagates — fold them into a single reader with explicit fallback/degrade parameters and rename the shadowing locals in prunable_sessions/request_graceful_stop.

pbean:
Closing as not worth tracking: pure refactor with no defect.

PR #691, opened 2026-08-22
fix: decouple the core CLI surface from the [tui] extra (#650 #678 #679)

### LINKED ISSUE #650: [BUG] error: No module named 'pyte'
### Description

pyte and rich are not declared as dependencies i

### Steps to reproduce

1. uv tool install bmad-loop
2. bmad-loop ls

### Expected behavior

Command runs and lists loop entries.

### Actual behavior

error: No module named 'pyte'
pyte and rich are not declared as dependencies in bmad-loop's package metadata, so uv tool install doesn't pull them in.

### Screenshots

_No response_

### Which area is this for?

Setup / init

### bmad-loop Version

0.10.0

### Which coding CLI are you using?

Other

### Operating System

Linux

### Relevant log output

```shell

```

### Diagnostic dump (bmad-loop diagnose)

_No response_

### Confirm

- [ ] If I attached a diagnose dump, I reviewed it and confirmed it contains no PII, secrets, or project-specific information
- [ ] I'm using the latest version

### LINKED ISSUE #678: cmd_tui's extras guard is dead: tui/app.py imports `rich` before `textual`, so the friendly message never prints
`cmd_tui`'s `ModuleNotFoundError` guard maps only `("textual", "tomlkit")`, but the name that actually surfaces on a no-extras install is `rich` — so the guard misses, re-raises, and the operator gets a traceback instead of the install hint it exists to print.

## The guard

```python
# src/bmad_loop/cli.py:3568-3577
try:
    from .tui.app import run_tui
except ModuleNotFoundError as e:
    if (e.name or "").partition(".")[0] in ("textual", "tomlkit"):
        print(
            "error: the TUI requires optional dependencies — uv tool install 'bmad-loop[tui]'",
            file=sys.stderr,
        )
        return 1
    raise
```

## Why it never fires

`src/bmad_loop/tui/app.py` imports `rich` **before** `textual`:

```python
# src/bmad_loop/tui/app.py:20-21
from rich.text import Text
from textual import work
```

So `from .tui.app import run_tui` raises with `e.name == "rich"`, which is not in the tuple. The `raise` arm runs.

Compounding it: **`rich` is declared in no extra and no dependency group at all** (`pyproject.toml`) — it arrives only as a transitive dependency of `textual`. A guard that enumerates the extra's own package names therefore cannot ever catch it.

## Reproduction

Under a `sys.meta_path` blocker for `pyte` / `rich` / `textual` / `tomlkit`, reproducing `cmd_tui`'s try/except verbatim:

```
import .tui.app raised ModuleNotFoundError name='rich'
  guard tuple ('textual','tomlkit') matches? False
  => GUARD MISSED -> bare raise, user sees a traceback
```

## Scope

Same family as #650 (`cmd_list` → `tui.data` → `import pyte`), and the two should ship together. They are **not** the same defect: #650 is a core command that should not need the extra at all, whereas `tui` genuinely requires it — here the bug is purely that the failure is unhandled rather than explained.

## Fix notes

Gate on the failure, not on a package allowlist — the import target is the extra's own module tree, so any `ModuleNotFoundError` escaping `from .tui.app import ...` means the extra is not installed. If an allowlist is kept for precision, it must include `rich`.

**The guard test must block `pyte`, `rich`, `textual` and `tomlkit`.** `tui/data.py:24`'s `import pyte` precedes its `rich` imports, so a blocker covering only `pyte` leaves the `rich` half untested — which is how this survived. Ablation rule applies: restore the narrow tuple and confirm the test reddens.

### LINKED ISSUE #679: settings_schema imports tui.settings (tomlkit) at module scope — inert today, but uncovered by any extra-less import test
`settings_schema.py` reaches into the `tui` extra at module scope. It is inert today only because of who imports it — nothing about the module itself keeps it off a core path.

## The chain

```python
# src/bmad_loop/settings_schema.py:39
from .tui.settings import STAGES
```

```python
# src/bmad_loop/tui/settings.py:15
import tomlkit          # declared only in the [tui] extra
```

Both are module-top imports, so any import of `settings_schema` on an install without the extra raises `ModuleNotFoundError: No module named 'tomlkit'`.

## Why it does not fire today

`settings_schema` is currently reached only from `src/bmad_loop/tui/screens/settings_screen.py:46` — a module that already requires the extra. The dependency is therefore satisfied by accident of the call graph, not by design.

## Why it is worth a line of defense

This is the third instance of one pattern, and the other two are live:

- #650 — `cmd_list` → `tui.data` → `import pyte`. A core command, broken since v0.6.1.
- #678 — `cmd_tui`'s guard misses `rich` because `tui/app.py` imports it before `textual`.

`tui/__init__.py` states the invariant the family keeps violating:

> `data` is the pure-stdlib observation layer and must stay importable without the extra

Nothing enforces it. There is no test anywhere in `tests/` that imports the core CLI surface with the extra's modules blocked, which is why #650 shipped through 23 releases undetected.

## Suggested handling

No behavior change is needed — the ask is coverage. When the guard test for #650/#678 lands (blocking `pyte`, `rich`, `textual` **and** `tomlkit`), extend it to assert that `bmad_loop.settings_schema` imports cleanly under the blocker, so a future core-path consumer of that module fails in CI rather than on a user's install.

If the import is instead made lazy inside the function that needs `STAGES`, the same test still pins it.

pbean:
Three open issues form one family — core CLI paths leaking into the optional `[tui]` extra. One commit per issue on one branch.

### `ls` crashed on a core-only install (#650)

`cmd_list` reached the run inventory through `tui.data`, which imports `pyte` and three `rich` names at module scope, so `bmad-loop ls` on a `uv tool install bmad-loop` died with `ModuleNotFoundError: No module named 'pyte'` — broken since v0.6.1. The reader moved into `runs.py` (core); `tui/data.py` re-exports the names so `data.discover_runs` and the monkeypatched `data.liveness` stay the TUI-facing surface.

### The `[tui]` install hint was dead code (#678)

`cmd_tui`'s guard allowlisted `("textual", "tomlkit")`, but `tui/app.py` imports `rich.text` *before* `textual` — so the real first failure escaped the arm and printed a traceback instead of the `bmad-loop[tui]` hint. The guard is now failure-gated: any `ModuleNotFoundError` whose root package is not `bmad_loop` prints the hint and returns 1, while a missing `bmad_loop.*` submodule still re-raises. `rich>=14.2` is now declared in the extra rather than arriving only as a transitive dependency of textual.

### The settings schema reached the extra at module scope (#679)

`settings_schema.py` imported `STAGES` from `tui/settings.py`, which imports `tomlkit` at module scope. Inert today — the settings screen is its only importer — but nothing kept it off a core path, which is the shape the first issue here took. `STAGES` now lives in `policy.py` beside the loop that consumes it; `tui/settings.py` keeps the name as a pinned re-export.

### The guard that would have caught all three

`tests/test_entry_point.py` gains a subprocess test that blocks `pyte`/`rich`/`textual`/`tomlkit` at `find_spec` — raising rather than returning `None`, since the dev venv has all four installed and a deferring finder would make every assertion vacuous. An `import pyte` floor proves the blocker bites. Under it: the core CLI and the settings schema both import, `list` returns 0, and `tui` degrades to the install hint. Each of the three arms has a recorded ablation in the docstring.

CI's build job gains an isolated wheel run of `bmad-loop list` against an empty project — a genuinely extra-less environment, stronger than the in-process blocker. `--version`, all the packaging job ran before, is answered by argparse before any subcommand code imports; that blind spot is why this family survived 23 releases.

Docs list `rich` in the extra's contents, and the TUI guide gains the long-missing `pyte`. A follow-up on consolidating the two liveness readers is filed as #690.

Tests: 6410 passed / 47 skipped, pyright clean, `trunk check --all` clean.

Closes #650, closes #678, closes #679

## Summary by CodeRabbit

* **New Features**
  * Added run discovery with statuses such as running, paused, finished, stopped, crashed, and interrupted.
  * Added graceful-stop and paused-stage indicators to run listings.
  * Added the `rich` dependency to the TUI option.

* **Bug Fixes**
  * Core-only installations can list runs without TUI dependencies.
  * Missing TUI dependencies now show a clear installation hint instead of a traceback.

* **Documentation**
  * Updated installation, testing, and dependency guidance for the TUI and core CLI.


pbean:
@codex review

pbean:
@coderabbitai review

pbean:
@codex review

pbean:
@coderabbitai review

Issue #692, opened 2026-08-22
Adopt `git config --show-scope` in `_shield_verify_activation` now that 2.26 is inside the floor


pbean:
`install._shield_verify_activation` verifies the git-add shield actually applies by asking git which excludes file it resolved, and infers from that whether an ambient override won. `git config --show-scope` would have git **name** the winning scope directly, which is the fact the function is reconstructing.

It was ruled out on a version argument: `--show-scope` is git 2.26, the shield's own gate was git 2.20, and "a probe flag with its own version floor turns every rc-branch below it into a silent default".

That reason has expired. With the project support floor at `verify.GIT_FLOOR` (git 2.34), `--show-scope` is present on every supported git.

**Why this was not done alongside the floor change.** It is a behavior change inside the code path that authorizes an irreversible repo-format write (`extensions.worktreeConfig`), so it earns its own review rather than riding along with a support-floor bump. Things that review has to settle:

- what each new scope token means for the degrade decision, including `command` (a `git -c` / `GIT_CONFIG_PARAMETERS` / `GIT_CONFIG_COUNT` override on the session's own invocation);
- what the rc branches become once the scope is answered rather than inferred;
- whether the `GIT_CONFIG_COUNT` / `GIT_CONFIG_PARAMETERS` enumeration argument in that docstring — the reason the function verifies the post-condition instead of detecting the override's origin — still holds in the adopted shape, or is restated. It should be re-read against the new code, not carried over verbatim.

**Marker in tree:** `src/bmad_loop/install.py:2126` names this issue in the docstring that records the deferral.

pbean:
Adopted in #696 — the activation probe now carries `--show-scope`, so the degrade reason names the winning scope instead of inferring one; decision space unchanged.

Issue #321, opened 2026-07-26
bmad-loop mux renders a broken table when a backend's version() is multi-line (psmux)


pbean:
## Summary

`TerminalMultiplexer.version()` may return a **multi-line** string, and two of its four consumers
render it as if it were one line. Observed on a native-Windows host during the #310/#315 live
verification (psmux 3.3.7, `05cc5d4`).

`psmux -V` prints two lines — the `tmux X.Y.Z` compatibility line plus its own — and
`BaseTmuxBackend.version()` (`adapters/tmux_base.py:455`) returns `proc.stdout.strip()`, which
removes surrounding whitespace but not the embedded newline:

```
tmux 3.3.7
psmux 3.3.7 (05cc5d4 2026-07-20)
```

## What breaks

`bmad-loop mux` on that host:

```
NAME   PLATFORM  AVAILABLE  VERSION                                      SELECTED
tmux   no        yes        tmux 3.3.7
psmux 3.3.7 (05cc5d4 2026-07-20)
psmux  yes       yes        tmux 3.3.7
psmux 3.3.7 (05cc5d4 2026-07-20)  * platform default for win32
```

`cmd_mux` (`cli.py:400-414`) computes column widths with `len()` and pads with `ljust`. The
VERSION cell is 43 chars, so every row's VERSION column is padded to 43 — hence the gap in the
header — and the embedded newline then splits each data row across two printed lines, stranding
SELECTED on the second. The table is unreadable, and `* platform default for win32` looks like it
belongs to a version string.

## The consumers disagree about this already

- `diagnostics.py:218` — **defends**: `sanitize.scrub_text(raw, max_lines=1)`. Someone hit this before.
- `cli.py:407` (`bmad-loop mux`) — no defense; the breakage above.
- `runsetup.py:183` (`bmad-loop validate` preflight) — no defense. The finding message becomes
  `multiplexer PsmuxMultiplexer available (tmux 3.3.7\npsmux 3.3.7 (…))`, and the raw value also
  lands in the structured `{"version": version}` payload.
- `multiplexer.py:477` (forced-backend warning) — safe **by accident**: it uses `{version!r}`, so
  the newline renders escaped.

One defended caller and two undefended ones says the invariant belongs at the seam, not at each
call site. The ABC docstring (`multiplexer.py:290`) still says "Used by the diagnostic dump",
which is now stale — there are four consumers.

## Suggested shape (not prescriptive)

[...]

PR #694, opened 2026-08-22
feat(verify,cli,install,docs): pin the git support floor at 2.34 and enforce it

### LINKED ISSUE #693: Collapse the two test-only `(2, 31)` version guards in tests/test_install.py
Two tests in `tests/test_install.py` skip when the host git is below 2.31, because they drive the shield through `GIT_CONFIG_COUNT`, which landed in git 2.31:

- `tests/test_install.py:3664` — the parametrized channel case, skipping the `GIT_CONFIG_COUNT` arm with "the GIT_CONFIG_PARAMETERS case covers older git";
- `tests/test_install.py:3723` — skipping outright with "GIT_CONFIG_COUNT is git 2.31".

With the project support floor at `verify.GIT_FLOOR` (git 2.34), 2.31 sits below it: no supported host can reach either skip. They are dead guards, and a dead skip is the bad kind of dead code — a skipped test reads as a pass in the summary, so if one ever did fire it would weaken the suite silently.

Deleting the two `pytest.skip` guards is trivial. The actual work is deciding what the surrounding test design should be once "older git" means 2.34:

- the parametrized case's `GIT_CONFIG_PARAMETERS` arm is justified as the one that "covers older git" — with no older git in the supported range, that arm needs a reason of its own or needs to go;
- the `'k=v'` encoding is chosen there because it is "the only one honored at BOTH ends", where the ends were 2.20.4 and 2.55; at the floor, `'k'='v'` is honored too;
- the docstring's comparison table (`git 2.20.4` vs `git 2.55.0`) records real measurements and should stay readable as a measurement record, but it is written around a range that no longer starts where the prose says it does.

Unrelated: the `sys.platform == "win32"` skip on the `GIT_CONFIG_PARAMETERS` arm is about sq-quoting vs. backslash paths, not about a version, and is not part of this.

pbean:
## Why

git is the tool every unit of work moves through, and it was the one prerequisite this project never declared — absent from the README's Requirements and the setup guide's platform list. The only git version number in the codebase was `install._WORKTREE_CONFIG_GIT = (2, 20)`, a **capability** floor belonging to one feature (the git-add shield's `extensions.worktreeConfig` write), which `verify.py:926` and `:1002` had since started citing as if it were the project's **supported range**.

That conflation has a concrete cost: review rounds keep re-litigating legacy-git behavior because the shield module has to keep writing 2.20-era code.

## What

One declared, enforced floor — `verify.GIT_FLOOR = (2, 34)`, inclusive — so nothing in the project has to reason about older git again.

- **`verify.py`** owns it, since it owns the `_run_git` chokepoint. `git_version_at_least` moves there from `install.py` and becomes public; `git_below_floor` is the probe; `git_floor_text` is the single formatter so the four messages naming the floor cannot drift. Probe and predicate are deliberately **separate seams** so they can be ablated independently (the #464 trap — one ablation does not cover both axes).
- **`run`, `sweep`, `resume` hard-refuse** below it, ordered ahead of the configuration refusals: an under-floor git is a fact about the machine that no edit to `policy.toml` can answer, so pointing the operator at their isolation setting first would send them at the wrong problem. `_sweep_factory` raises instead of returning an rc, matching how the isolation conflict is handled there, which leaves the parent's trigger unspent.
- **`validate` reports `git.version` as a problem** (exit 1), not a warning, so its verdict cannot disagree with the run abort — the house convention at `cli.py:341-342`, mirroring `policy.isolation-repo-root`. Both surfaces render the same string. It stays silent on `GitError`, which `git.probe` immediately above already owns.
- **`diagnose` records the host's git version** — additive `EnvInfo.git_version`, so no `SCHEMA_VERSION` bump; scrubbed *then* folded, per the #321 ordering rule.
- **The shield's gate becomes the project floor.** Its old message made a capability claim that would be false at 2.34 — git 2.25 has both `extensions.worktreeConfig` and `git config --worktree` — so it now refuses as **policy**. The gate survives as defence in depth: with the run refusal in place no supported entry point reaches it with an old git, but its unreadable-answer arm still fires on a perfectly current host.

**Fail closed on every arm.** A git that cannot be spawned, times out, or answers unparseably is refused exactly like one that is simply too old. The failure a floor must not have is the generous one.

## Why 2.34, and why it is a *support* floor

2.34 is [Ubuntu 22.04 LTS's stock git](https://packages.ubuntu.com/jammy/git), keeping that LTS in support through its April 2027 EOL, while dropping [Ubuntu 20.04 (2.25.1)](https://packages.ubuntu.com/focal/git) and [Debian 11 (2.30)](https://packages.debian.org/bullseye/git), both at or past end of standard support.

Nothing in `src/bmad_loop`'s git argv actually requires 2.34 — the highest capability in use is `git config --worktree`, from 2.20. This is deliberately a **support** floor set above the capability floor, and the distinction is kept explicit throughout because it has burned this project before. It is not operator-configurable, so there is no `policy.py` field and no `data/settings/core.toml` entry.

## Tests

Each at the lowest layer that catches its regression: a `git_version_at_least` table (moved, boundary retargeted to 2.34/2.33, vendor-soup rows kept), `git_below_floor` unit coverage for below/unparseable/at-or-above, seam tests for all four refusal sites, `validate`'s finding **and exit code** plus the `--json` one-object shape on the failing path, and `diagnose`'s env field.

The behavior change itself is pinned by a **git 2.25** case — the pre-existing 2.19.4 test was refused before and after the bump and therefore proves nothing about it.

All four plan-required ablations were run red-then-green:

1. delete the `_reject_under_floor_git` call site in `cmd_run` → refusal test red *(wiring axis)*
2. make `git_version_at_least` return `True` on no-match → unparseable test red *(predicate axis)*
3. delete the `report.fail` → validate test red **on the exit code**, not just the finding (a `report.warn` typo would still emit `git.version` and keep a findings-only assertion green)
4. point the shield gate back at `(2, 20)` → the 2.25-refuses test red

Versions are faked by monkeypatching `verify.git_below_floor` or `verify.git_bytes` — never by touching real git.

## Also in here

The 2.20-era **prose** is retired across `install.py`, `verify.py` and the tests: docstrings that stated a *support range* off a *capability* constant, and rationales phrased as "above/at the 2.20 floor" that become false at 2.34. Leaving a stale rationale is how the next reviewer re-derives the wrong conclusion.

Two dead `(2, 31)` test guards are deleted — they could only ever have fired below the new floor, and a skip reads as a pass. The skip count is unchanged at 47 across their removal, confirming they were never firing.

`tests/test_install.py:4350` previously attributed a git message wording to "the floor"; that was measured at 2.20, and "the floor" now names a version whose wording is unmeasured, so it is restated on the stable half (the rc) — which is what the code actually keys on.

## Not in here

Adopting `git config --show-scope` in `_shield_verify_activation` now that 2.26 sits inside the floor is tracked separately as #692, and `src/bmad_loop/install.py:2126` points at it. It is a behavior change inside a code path that authorizes an irreversible repo-format write, so it earns its own review rather than riding along with a floor bump.

No version bump: `## [Unreleased]` is the contract and `scripts/sync_version.py` stamps at release. The CHANGELOG entries record that this is a **minor** — a support-floor break *and* an exit-code change on a released command.

## Verification

- `uv run pytest -q -n logical` — 6427 passed, 47 skipped
- `uv run pyright` — 0 errors, 0 warnings, 0 informations
- `trunk fmt` — no issues
- `trunk check --all` — 258 files, no issues

Closes #693

## Summary by CodeRabbit

- **New Features**
  - Added Git version validation with a minimum supported version of 2.34.
  - `validate` now reports Git version problems with clear diagnostics.
  - Environment diagnostics include the detected Git version.

- **Bug Fixes**
  - `run`, `sweep`, and `resume` now safely refuse unsupported, unavailable, or invalid Git versions.
  - Updated worktree protection behavior to consistently follow the Git 2.34 support policy.

- **Documentation**
  - Updated requirements, setup guidance, feature documentation, and changelog to describe the Git version requirement and command behavior.


pbean:
@codex review

pbean:
@codex review

pbean:
@codex review

pbean:
@codex review

PR #696, opened 2026-08-22
Adopt `git config --show-scope` in the shield's activation check

### LINKED ISSUE #692: Adopt `git config --show-scope` in `_shield_verify_activation` now that 2.26 is inside the floor

pbean:
`install._shield_verify_activation` proves the git-add shield actually applies: it re-reads the `core.excludesFile` it just wrote at worktree scope and byte-compares it to the written path. When the value mismatched it could only *infer* that "another configuration scope outranks it" — it could not say which. The probe now carries `--show-scope`, so git **names** the winning scope on the same single call.

### The decision space is unchanged

This runs inside the path that *revokes* an irreversible repo-format write (`extensions.worktreeConfig`), so the proceed/degrade outcome for every input real git can produce is preserved:

- a byte-identical value → **proceed**, whatever scope supplied it;
- every mismatch → **degrade**;
- every nonzero rc → **degrade**, with wording byte-identical to before.

`--show-scope` refines the *message*, not the verdict. The one added branch is fail-closed: an rc-0 answer carrying no seam NUL is now reported as an answer naming no scope, where before it fell through to the mismatch arm and fabricated a scope out of the unparsed bytes.

What the operator now gets told apart — because the two send you to opposite repairs:

| answering scope | what it means |
|---|---|
| `command` | an ambient override (`git -c`, `GIT_CONFIG_PARAMETERS`, `GIT_CONFIG_COUNT`) outranks the write |
| `worktree` | our own scope answered a different value — this code, not the environment |
| `local`/`global`/`system` | the worktree-scoped write is not in force **at all** |
| anything else | a token this code does not know, quoted verbatim |

A **matching** value under `command` scope still activates the shield. The post-condition is "git reads the file we wrote", and provenance is not a fault; a new test pins that against a future scope-first refactor.

### Floor context

`--show-scope` is git 2.26. It was ruled out when the shield's own gate was 2.20; #694 pinned the project support floor at `verify.GIT_FLOOR = (2, 34)`, so the flag is present on every supported git — that is what unblocked this.

### Measured, not predicted

- `-z --show-scope` emits `scope NUL value NUL` at **both** ends of the supported range: git 2.34.1 (the floor itself, in `ubuntu:22.04`) and git 2.55.0.
- The flag does not move the rc taxonomy — an absent key is rc 1 with and without it, measured at 2.34.1.
- `printf '\000'` emits a real NUL on dash (measured in `ubuntu:22.04` and `ubuntu:latest`, the CI runner's `/bin/sh`). The shell stub relies on this; a comment that asserted the opposite is corrected in place.

Every new refused-behavior test carries an ablation that was actually run and reddened it; the pre-existing inverse ablation record was re-measured, since the rc-gate deletion now lands in a different arm. The scope fence holds — the seed read, the home probe and the shared-config reads do **not** grow the flag, since rc 1 means ABSENT there.

Tests 6447 passed / 47 skipped, pyright clean, trunk check clean.

Closes #692

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved git-add shield activation checks by confirming both the configured excludes file and the Git configuration scope supplying it.
  * Added clearer notifications when command-level overrides, worktree mismatches, or other configuration scopes prevent activation.
  * The shield now safely remains inactive when Git returns incomplete or unrecognized configuration information.
  * Existing activation, degradation, and repository-format safeguards remain unchanged.

* **Documentation**
  * Updated feature documentation and the unreleased changelog to describe the enhanced verification and diagnostics.


pbean:
@codex review

pbean:
@coderabbitai full review

PR #700, opened 2026-08-23
fix(tui): show the pause reason for spec-less gate pauses

### LINKED ISSUE #515: tui: the story-gate pause reuses the spec viewer and never shows the pause reason
Split out of #502 review (codex P2). Pre-existing shape, newly reachable.

`action_review_pause` routes `PAUSE_STORY_GATE` to `_review_gate` (`src/bmad_loop/tui/app.py:546`), which builds a `SpecReviewModal` titled `"<label> — review the finalized spec"` (`:591-606`). A story gate fires **before** the story is recorded in `state.tasks` — deliberately, so a resume re-picks it and re-asks the ledger (`src/bmad_loop/engine.py:839`) — so `_paused_spec` always returns `(None, "")` for it (`app.py:783-793`). The operator gets `(no spec file resolved)` / `(empty spec)` (`src/bmad_loop/tui/screens/modals.py:474,478`) under a title promising a spec, with `Approve & resume` as the only action; if the entry is still open, resuming re-pauses at the same gate.

The actionable payload is `state.paused_reason` — it names the blocking entries and the remedy (`engine.py:939-943`). `_review_gate` never renders it.

Scoping, because codex overstated this:

- **Not introduced by #502.** That branch does not touch `tui/app.py`; `PAUSE_STORY_GATE` was already in the routing tuple and already badged (`src/bmad_loop/tui/widgets.py:77`). It was a reserved constant nothing raised, so the path was dead until #502 made it live.
- **Not unique to story-gate.** `PAUSE_EPIC_BOUNDARY` raises without a `story_key` (`engine.py:5390-5394`) and hits the identical spec-less path with a *worse* subtitle (`"?"`, `app.py:796`). Story-gate at least renders `<story-key> — <title>`.
- **The operator is not stranded.** The reason is shown in the run-header banner (`src/bmad_loop/tui/widgets.py:224-228`) and in the resume confirmation (`modals.py:308-311`) — just not in the `p` viewer. #502 documents that in `docs/tui-guide.md:443`.

Fix shape: render `paused_reason` in `_review_gate` when there is no spec (or give the spec-less gates their own small viewer), which fixes epic-boundary at the same time.

No TUI test exercises a story-gate pause — `tests/test_tui_app.py` has no `story-gate` coverage; its `SpecReviewModal` tests all use a pause that has a real `task.spec_file`. Worth adding with the fix.

pbean:
## What

The TUI's `p` viewer routed `PAUSE_STORY_GATE` and `PAUSE_EPIC_BOUNDARY` into `SpecReviewModal`, which promised a spec neither pause has. `_review_gate` now branches on whether a spec resolved at all:

- **Spec-carrying gates** (spec-approval, and every plan-checkpoint viewer that shares the modal) are untouched — same title, same actions, byte-identical.
- **Spec-less gates** open a new compact `PauseReasonModal`: the pause reason, which names the blocking deferred-work entries and the remedy, with **Resume** and **close**. An epic boundary — which has no story key at all — is subtitled with the run id instead of the old `?`.

## Why

A story gate fires *before* the story is registered in `state.tasks`, deliberately, so a resume re-picks the story and re-asks the ledger (`engine.py`). So `_paused_spec` returns `(None, "")` and the operator got `(no spec file resolved)` / `(empty spec)` under a title promising a spec, with `Approve & resume` as the only action. The actionable payload — `state.paused_reason` — was rendered in the run-header banner and the resume confirmation, but never in the viewer the badge points you at. The path went live when #502 merged.

The branch condition is `spec_path is None`, not `not spec_text`: an *unreadable* spec file returns `(path, "")` and must stay on the spec viewer, which shows the path plus `(empty spec)`. That distinction has its own test.

Two shapes reach the reason viewer and both are covered: the engine gate (no task registered) and sweep's ledger-migration gate (task registered, no `spec_file`).

Reason text is arbitrary engine text, so it renders as rich `Text`, never markup. The modal uses `ConfirmModal`'s bounded `height: auto; max-height: 60%` tier, and joins the existing 39x9 minimum-size matrix.

## Tests

Six new/extended cases in `tests/test_tui_app.py`, next to the existing `test_gate_pause_resume` regression guard (untouched):

- `test_story_gate_pause_shows_reason_and_resumes` — both spec-less shapes, asserts the routing, the reason text, and that **Resume** reaches `resume_detached`
- `test_epic_boundary_pause_shows_reason_and_run_id_subtitle` — positive assert on the run-id subtitle
- `test_spec_approval_unreadable_spec_still_uses_spec_viewer` — pins the branch condition
- `test_story_gate_empty_reason_renders_fallback` — `paused_reason == ""` is reachable
- a `pause-reason` arm in the 39x9 minimum-size matrix
- `test_pause_tag_and_label_render` extended — the gate titles now come from `pause_label`, so those strings are load-bearing

Each was ablated against the whole file. Reverting the spec-less branch reddens the four routing tests; flipping the condition to `not spec_text` reddens only the unreadable-spec test; restoring the `?` subtitle reddens only the epic-boundary subtitle assert; opting the modal out of `BaseDialog`'s `max-width` clamp reddens only the new matrix case.

Full suite, `pyright` and `trunk check` clean.

Closes #515

## Summary by CodeRabbit

- **Bug Fixes**
  - Story-gate and epic-boundary pauses now display a clear pause-reason viewer instead of an empty specification panel.
  - Pause details identify the blocked story or run and explain the required remedy.
  - Resuming correctly reselects the story and re-asks the ledger.

- **Documentation**
  - Updated paused-run guidance to distinguish specification, story-gate, and epic-boundary viewers.
  - Added an unreleased changelog entry describing the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,8 @@ breaking changes may land in a minor release.
 
 ### Fixed
 
+- TUI: story-gate and epic-boundary pauses open a pause-reason viewer naming the blocking
+  entries and the remedy, instead of an empty spec pane (#515)
 - **Provisioning refuses an unparseable seeded hook config instead of silently replacing it
   (#592).** An isolated worktree's seeded `.claude/settings.json` that failed to parse was read
   as an empty document, so the relay merge always ran against a blank baseline and published a
--- a/docs/tui-guide.md
+++ b/docs/tui-guide.md
@@ -486,11 +486,13 @@ artifacts the engine already wrote.
   interactive agent as `R`; **Re-arm & resume** (offered once the resolve agent has
   recorded a resolution) re-arms and resumes — deleting a sentinel with a preserved
   copy for a clean re-dispatch. Both refuse a still-live engine.
-- **Spec-approval / epic / story gate** — reuses the spec viewer (view the finalized
-  spec, then **Approve & resume**), so the pre-existing sprint-mode gates inherit the
-  same richer surface. A story gate fires before the story is recorded, so it has no
-  spec to show; read its reason — which names the blocking entries and the remedy — in
-  the run-header banner or the resume confirmation.
+- **Spec-approval / epic / story gate** — a spec-approval gate reuses the spec viewer
+  (view the finalized spec, then **Approve & resume**), so the pre-existing sprint-mode
+  gate inherits the same richer surface. Story-gate and epic-boundary pauses have no
+  spec to show — a story gate fires before the story is recorded, an epic boundary has
+  no story at all — so they open a compact pause-reason viewer instead: the reason names
+  the blocking entries and the remedy, and **Resume** re-picks the story and re-asks the
+  ledger, so a gate that is still open legitimately re-pauses.
 
 `p` and `R` overlap for an escalation (both reach Resolve); `p` also exposes
 Re-arm & resume inline once a resolution exists. Pause badges in the run list and
--- a/src/bmad_loop/tui/app.py
+++ b/src/bmad_loop/tui/app.py
@@ -46,6 +46,7 @@ from .screens.modals import (
     ConfirmResumeModal,
     DecisionModal,
     EscalationModal,
+    PauseReasonModal,
     SpecReviewModal,
     StartRunModal,
     StartSweepModal,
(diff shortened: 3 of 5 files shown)


pbean:
@codex review

PR #741, opened 2026-08-27
fix(verify): run the review gates in the right tree and spare an empty park (#695, #676)

### LINKED ISSUE #676: 0.10.0 field report: park-through-commit at a story's front gate classifies as 'no changes' -> retry, auto-rollback of the park commit, defer
## Summary

Field report from a production run on **0.10.0** (adapter dev=`claude`, review=`codex`, `isolation="worktree"`, sprint mode). A dev session that ended in a deliberate `awaiting-operator` park — frontmatter flip + doc-only commit, no code yet by design — was classified by the dev-decision as `no changes in worktree since baseline commit`. That triggered `retry`, an `rollback-auto` that discarded the (correct) park commit, a second dev session that redid the same work (~10 min, ~4.6M tokens) and parked identically, and finally `defer` + run end. The park itself was the story working exactly as authored; the loop around it read it as failure.

Refs: #335 (awaiting-operator program), #348 (park-through-commit entry path), #337 (prior field-report precedent). Filing this before our next epic, which has several mobile stories with the same mock-gate design (park **before** any code is written).

## Journal evidence (run `20260819-234424-58a3`, story `30-8-ui-checklist-modulos-mobile`)

```
dev-decision   attempt=1 session_status=completed action=retry
               reason="no changes in worktree since baseline commit"
               env_fault=False session_vanished=False
rollback-auto  baseline=41f82a44 note="reverting tracked changes + run-created untracked files"
attempt-commits-preserved ref=attempt-preserve/20260819-234424-58a3-1f0922c3 count=1
attempt-worktree-preserved ref=refs/attempt-preserve-dirty/20260819-234424-58a3-41f82a44-1
(session-start dev-2 ... redoes Task 0 + Task 1, parks the same way)
dev-decision   attempt=2 session_status=completed action=defer
               reason="no changes in worktree since baseline commit"
story-deferred reason="no changes in worktree since baseline commit"
run-complete
```

The telling detail: `attempt-commits-preserved count=1`. The engine **knew** the attempt had produced a commit — it preserved it while rolling back — yet the decision reason says "no changes". The classifier appears to compare the worktree against the baseline and to ignore commits made on the attempt branch, so any attempt whose entire output is a coherent commit (a park is the purest case) reads as no progress.

## Why the mock-gate park is the worst case for this

The story's design (a visual mock gate) requires the dev to stop for operator approval **before writing any product code**: execute the setup tasks, write the mock/spec artifacts, flip the story frontmatter to `awaiting-operator`, commit (doc-only), end the session. So at decision time the worktree is clean and the only evidence of progress is the commit — precisely the input the classifier discounts. A park that happens mid-story (code already in the tree) coincidentally survives; a park at the story's front gate deterministically hits retry → rollback → defer. In an earlier epic, a mid-story `awaiting-operator` park on this same project (0.9.x era, story `28-8`) was recognized fine — the difference matched exactly "was there anything uncommitted in the tree".

No park-related journal kind was emitted at any point in this run, and no record appeared under `.bmad-loop/operator/` — consistent with the dev declaring the park by editing frontmatter + committing rather than through whatever channel `record_park` expects. If the #348 entry path requires a specific hook signal, that contract does not seem to be part of the dev-facing skill's instructions, so a session can produce a park that is coherent to a human and invisible to the engine.

## What we'd hope for

1. A completed session whose attempt branch carries commits should not classify as "no changes" — the commit list the engine already builds for `attempt-commits-preserved` seems to be the missing input to the decision.
2. If the attempt's final commit leaves the story frontmatter in `awaiting-operator`, treat it as a dev-declared park (or surface a validation error telling the dev which channel to use), rather than retry/defer.
3. After the operator resolves the gate, resuming currently requires manually invoking a fresh run; acceptable, but worth documenting as the expected flow if intentional. (The final promotion to `done` also happened outside any run, via manual spec+board edit.)

We read the v0.11.0 notes (#123 — spec-only lifecycle-flip retries no longer demand a false manual rollback, park-record restore hardening) and have upgraded; from the changelog alone we could not tell whether this exact case — completed session, park-through-commit, clean tree — now classifies as progress. Happy to test a scenario or supply the full journal/log excerpts if useful.

## Secondary observations (same run family, minor)

- One `spec-deferrals-harvested` entry wrote a deferred-work item with a truncated title (we repaired the ledger by hand; it later surfaced as a duplicate of an existing entry).
- Two deferred-work items that did reach the ledger never appeared in any journal event, so the run's telemetry undercounts harvested deferrals.

Environment: macOS (darwin), tmux 3.7b, uv-managed install from this repo, `bmad-loop 0.10.0` at the time of the run (since upgraded to 0.11.0 pinned).

### LINKED ISSUE #695: verify_review* passes paths.project as verify cwd instead of repo_root
### Summary

When `_bmad/bmm/config.yaml` sets `repo_root` to a different path than `project` (the documented split-repo setup — planning artifacts in one repo, code in another), `[verify].commands` runs in the wrong directory on the post-review gate, so verify fails on every story even though the same commands succeed if run manually in `repo_root`.

### Root cause (verify.py, v0.9.0)

- `verify_dev`'s companion verify-commands call (`engine.py:1741`) uses `self.workspace.root`, which for `isolation = "none"` correctly resolves to `repo_root`.

- But `verify_review`, `verify_review_stories`, and `verify_review_bundle` (`verify.py:2096`, `2128`, `2168`) all call:

  ```python
  return verify_commands_outcome(policy, paths.project)
  ```

  using `paths.project` instead of `paths.repo_root` / `workspace.root`. This runs whether or not the standalone review session fires — it's the gate hit on virtually every story (including "review disabled" dev-only finalization).

### Repro

`_bmad/bmm/config.yaml`:

```yaml
planning_artifacts: "{project-root}"
implementation_artifacts: "{project-root}"
repo_root: "/abs/path/to/code-repo"   # a different git repo than project-root
```

`.bmad-loop/policy.toml`:

```toml
[verify]
commands = ["npx nx affected -t lint test --base=HEAD~1"]
```

Result — every story's post-review verify gate fails:

```
NX   Could not find Nx modules at "<ancestor of project, NOT repo_root>".
Have you run npm/yarn install?
```

even though `cd repo_root && npx nx affected ...` succeeds by hand.

### Impact

With `scm.rollback_on_failure = false` (recommended for in-place isolation to avoid discarding committed work), a verify failure on a real commit pauses the run for manual recovery (`ACTION REQUIRED` / rescue-branch-then-reset instructions) instead of continuing. From the outside this looks like "the loop stops after every story instead of iterating" — it's not a review/commit bug, it's the verify gate running in the wrong repo every time.

### Fix

`verify_review` / `verify_review_stories` / `verify_review_bundle` should pass `paths.repo_root` (or the equivalent `workspace.root`) to `verify_commands_outcome`, matching what the dev-side verify-commands call already does, so both gates run `[verify].commands` in the same directory.

### Workaround

Hardcode the real repo into the command string itself (wins over whatever cwd is passed, since commands run with `shell=True`):

```toml
commands = ["cd /abs/path/to/code-repo && npx nx affected -t lint test --base=HEAD~1"]
```

### Environment

```
bmad-loop version: 0.9.0
python: 3.11.15
os: Darwin 25.6.0
```

pbean:
Closes #695
Closes #676

Two confirmed work-loss fixes at the front gates, both scoped narrowly and both covered by ablated tests.

## #695 — the review gates ran the verify commands in the wrong tree

All three `verify_review*` gates ended in `verify_commands_outcome(policy, paths.project)`, while the dev gate runs the same commands in `Engine.workspace.root` — which `Workspace.default` sets to `paths.repo_root`. Under `isolation = "none"` with a `repo_root` override the two gates therefore build **different trees on the same commit**: dev the code repo, review the artifact project. A `[verify] commands` entry that passes for dev can fail, or vacuously pass, for review.

The three sites now share one helper, `_verify_review_commands`. That is deliberate rather than three edited lines: the change splits the two roots the review gates use — the *artifacts* each gate reads (`paths.sprint_status`, `paths.deferred_work`, the claimed spec) stay project-rooted because that is where BMAD writes them, and only the operator's shell commands move because those are about the code. One helper states that split once and keeps it from drifting between the three gates.

Worth noting that the review gates were the only outlier, not one of two competing conventions: the dev gate runs them in `Engine.workspace.root`, and `cli._reverify` — the `confirm --reverify` path — is called with `paths.repo_root` at both of its call sites. `repo_root` was already the answer everywhere else.

Reporter's scope, confirmed and narrowed: reachable only under `isolation = "none"` plus a `repo_root` override.

## #676 — a park that produced no code was judged as having produced nothing

The parked leg of `verify_dev` passed `extra_exclude=engine_written` unconditionally, so proof-of-work ran on a park. But `verify_dev_exclude_relpaths` excludes `paths.sprint_status` and the session's own spec by name — and those are exactly the only two files a park that got it right on its first pass writes. The gate read a correct park as "no changes", retried it until the attempt budget ran out, then rolled the park commit back and deferred.

The fix is the reporter's: `extra_exclude=None if parked else engine_written`, which is structurally the same idiom the plan-halt leg in `verify_dev_stories` already uses, for the same reason.

The `needs-design` question on that issue — "does park become a free pass?" — is answered by the surrounding code rather than by new design. `extra_exclude=None` skips exactly one block inside `_verify_shared_gates`, the `has_changes_since` proof-of-work check; everything above it still runs on the parked leg, including the workflow-tag gate, the expected-status gate and the baseline-match gate. And `_operator_actions_gate` runs *before* the shared gates on that leg, already refusing a park that declares no usable `operator_actions`, while the (spec status, board status) pair still has to match. What this does give up is that a park which *did* produce code no longer clears proof-of-work either. That is a small policy widening rather than a pure bug fix, and it is harmless in the only direction that matters: skipping a gate never fails work that would have passed it.

## Tests

Two new rows in `tests/test_verify.py`, both ablated — the gating code deleted and the row confirmed to go red.

**The empty-proof park.** No existing row could see this bug: the shared `_park` helper writes `src.txt`, so every current park row has a non-empty proof diff. The helper takes `code=False` now. The row carries its own control in the same test — the *same* empty tree under the *same* flags, differing only in the observed spec status, must still fail proof-of-work. Without that control the row would pass whenever the tree simply was not empty, which is the way this shape fails silently. Ablation: reverting `None if parked` reddens it with `no changes in worktree since baseline commit`, and the eleven pre-existing park rows all stay green.

**A `repo_root != project` review-gate row**, parametrized across all three gates. The `project` fixture sets no `repo_root`, so `repo_root == project` and no existing row can distinguish the two. This one splits them with `dataclasses.replace` and pins the command's cwd from both sides: a marker only `repo_root` holds must pass, and a marker only `project` holds must fail — one direction alone is satisfiable by a cwd that is neither. Ablation: reverting to `paths.project` reddens all three parameters.

## Gate

`uv run pytest -q -n logical` 7010 passed / 49 skipped · `uv run pyright` 0 errors · `trunk fmt` · `trunk check` · `trunk check --all` 258 files, no issues.

Related to #443, which names this `[verify] commands` split among the inconsistencies to settle; that bullet is now stale and comes out once this lands.

## Summary by CodeRabbit

* **Bug Fixes**
  * Review verification commands now run against the same code workspace as development verification, preventing inconsistent results when a repository root is configured.
  * Sessions that are intentionally parked without producing code now pass verification correctly.
  * Parks without valid operator actions continue to be rejected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,20 @@ breaking changes may land in a minor release.
 
 ### Fixed
 
+- **Review-side `[verify] commands` run in the same tree the dev gate uses** (#695). All three
+  `verify_review*` gates ran them in `paths.project`, while the dev gate runs them in the
+  workspace root — `paths.repo_root` under `isolation = "none"`. With a `repo_root` override the
+  two gates built different trees on the same commit, so a command that passed for dev could
+  fail (or vacuously pass) for review. The artifacts each gate reads stay project-rooted; only
+  the operator's shell commands move, and the three gates now share one helper so the split
+  cannot drift.
+- **A park that legitimately produced no code passes the front gate** (#676). The parked leg
+  ran proof-of-work with the spec and the board excluded by name — the only two files a
+  first-pass park writes — so a correct park read as "no changes", retried until the attempt
+  budget ran out, and was then rolled back. That leg now skips the gate outright, the same
+  idiom the plan-halt leg already uses. Park is not thereby a free pass: the gate refusing a
+  park that declares no usable `operator_actions` runs ahead of it, and the spec/board pair
+  still has to match.
 - **The git-add shield refuses to enable `extensions.worktreeConfig` over an operator's
   explicit disable** (#396), instead of enabling it and deleting the line on rollback. The
   probe read the flag `--type=bool`, so a `false`/`off`/`no`/`0` in the shared config read
--- a/src/bmad_loop/verify.py
+++ b/src/bmad_loop/verify.py
@@ -3468,7 +3468,19 @@ def verify_dev(
         expected_status=(
             AWAITING_OPERATOR if parked else ("in-review" if review_enabled else "done")
         ),
-        extra_exclude=engine_written,
+        # A park may legitimately produce no code: the story's remaining work is a
+        # human's, and a session that correctly recognized that on its first pass
+        # wrote only the spec and the board — both of which the proof-of-work
+        # exclude already removes, so the gate reads the park as "no changes" and
+        # retries it to budget exhaustion, then rolls the park back (#676). Skip
+        # the gate on this leg (`extra_exclude=None`), the same idiom the
+        # plan-halt leg uses in `verify_dev_stories` for the same reason. Park is
+        # not thereby a free pass: `_operator_actions_gate` above already refused
+        # a park declaring no usable `operator_actions`, and the (spec, board)
+        # pair still has to match. The cost is that a park which *did* produce
+        # code no longer clears proof-of-work either — harmless, because skipping
+        # a gate never fails work that passed it.
+        extra_exclude=None if parked else engine_written,
         fm=fm,
     )
     if gate is not None:
@@ -3979,6 +3991,27 @@ def verify_commands_outcome(policy: Policy, cwd: Path) -> VerifyOutcome:
     return verify_command_results_outcome(run_verify_commands(policy, cwd), cwd)
 
 
+def _verify_review_commands(policy: Policy, paths: ProjectPaths) -> VerifyOutcome:
+    """Run the review-side verify commands in the tree the dev side already uses.
+
+    `paths.repo_root`, not `paths.project`: the dev gate runs them in
+    `Engine.workspace.root` (`_verify_commands_with_results`), which is
+    `paths.repo_root` for `isolation = "none"` and the worktree dir under
+    worktree isolation. With a `repo_root` override the two gates otherwise
+    disagree — dev builds the code tree, review builds the artifact tree — and a
+    `[verify] commands` entry that passes for dev fails (or vacuously passes) for
+    review on the same commit (#695).
+
+    This deliberately splits the two roots the review gates use: the *artifacts*
+    each gate reads (`paths.sprint_status`, `paths.deferred_work`, the claimed
+    spec) stay project-rooted, because that is where BMAD writes them; only the
+    operator's shell commands move, because those are about the code. The three
+    `verify_review*` gates share this one function so the split cannot drift
+    between them.
+    """
+    return verify_commands_outcome(policy, paths.repo_root)
+
+
 def verify_review(
     task: StoryTask,
     paths: ProjectPaths,
@@ -4051,7 +4084,7 @@ def verify_review(
             f"sprint-status for {task.story_key} is {sprint!r}, expected {expected!r}"
         )
 
-    return verify_commands_outcome(policy, paths.project)
+    return _verify_review_commands(policy, paths)
 
 
 def _is_signoff_regression(sprint: str | None, sprint_reached_done: bool, policy: Policy) -> bool:
@@ -4083,7 +4116,7 @@ def verify_review_stories(task: StoryTask, paths: ProjectPaths, policy: Policy) 
     status = status_of(fm)
     if status != "done":
         return VerifyOutcome.retry(f"spec status is {status!r}, expected 'done'")
-    return verify_commands_outcome(policy, paths.project)
+    return _verify_review_commands(policy, paths)
 
 
 def verify_review_bundle(task: StoryTask, paths: ProjectPaths, policy: Policy) -> VerifyOutcome:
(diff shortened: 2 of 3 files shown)


pbean:
@codex review

pbean:
Withdrawing this — the Wave 1 story is going to be executed through a different process. Reopening the underlying work under that flow instead; #695 and #676 stay open and untouched.

Issue #593, opened 2026-08-14
Atomic writes resolve parent directories by name, so follow_symlinks=False does not stop a symlinked parent


pbean:
`platform_util._atomic_write` — the shared body behind `atomic_write_text` and `atomic_write_bytes` —
resolves the target's **parent directories by name**, at both the stage and the publish step:

```python
target = path.resolve() if follow_symlinks else path
fd, tmp_name = tempfile.mkstemp(dir=str(target.parent), prefix=target.name + ".", suffix=".tmp")
...
atomic_replace(tmp, target)
```

`follow_symlinks=False` skips the resolve, so the **final component** is replaced rather than written
through — that part is sound, and its docstring argues it well. But `mkstemp(dir=target.parent)` and
the `os.replace` destination are still ordinary path lookups. If any directory *above* the file is a
symlink, both the temp and the published file land wherever that link points, and the no-follow buys
nothing.

Concretely: `follow_symlinks=False` at `.bmad-loop/policy.toml` stops a link planted **at
`policy.toml`**. It does not stop a link planted at **`.bmad-loop/`**. `mkdir(parents=True,
exist_ok=True)`, which several of these callers run first, succeeds against a symlink-to-a-directory,
so a planted parent survives the setup step too.

## Why this matters at these specific call sites

The reason `follow_symlinks=False` was chosen for them (#379, #590) is that they are machine-minted
files under a **session-writable** root — `runsetup` documents `.bmad-loop/policy.toml` as a path a
driven session can write. The escalation being closed is a sandboxed session getting the unsandboxed
orchestrator to write to a path of its choosing. A symlinked parent reopens exactly that escalation;
it just costs the attacker a directory swap instead of a file swap.

Affected: every `follow_symlinks=False` call site — `.bmad-loop/decisions.json`, the two per-run
sweep writes, `.bmad-loop/policy.toml` (three writers), story specs (three writers), and the park
records (three writers). The `follow_symlinks=True` sites have the same parent exposure, but there
the file itself is operator-curated and the threat model is different.

## The machinery already exists, and has one consumer

`platform_util` already ships the correct primitives, added for `tui/launch.py`'s control-window
record:

- `open_dir_confined(root, target)` walks each component `O_NOFOLLOW | O_DIRECTORY` relative to the
  one above it and hands back a **descriptor**, not a verdict — deliberately, because a boolean
  "is this path confined?" is stale the instant it returns.
[...]

PR #748, opened 2026-08-28
fix(tui,runs): anchor the paused-spec read and confine the replan write on the run's tree

pbean:
## Problem

`tui/app.py::_paused_spec` resolved `Path(task.spec_file)` against the TUI process cwd. Under worktree isolation `StoryTask._serialized_worktree_path` persists `spec_file` *relative* to the mounted worktree and `from_dict` reads it back raw, so that bare `Path(...)` named the **main checkout's** copy — which carries the same layout.

Three review surfaces displayed the wrong file, and `_do_replan` *wrote* to it: both `reset_spec_status` and `strip_auto_run_result` succeeded, because the main checkout's copy genuinely is under `project`, so containment accepted it and `reset` returned `True`. The operator saw "plan reset to draft", the run resumed, the worktree's real spec kept its terminal status — so the next dispatch did not re-plan — and an unrelated tracked file was rewritten. Pre-existing since #82, invisible to the suite because every TUI row set an absolute `spec_file`.

## Change

`runs._task_spec_path` / `_task_spec_root` were written for this exact defect at the `bmad-loop resolve` call site. Promote both to public and route the TUI's read, the replan's `confine_root`, and `resolve.build_context` through them, so the anchor and the containment root are one claim about which tree owns the spec.

Two follow-up fixes from review (`850f65d6`):

- **`task_spec_root` could return a root that cannot confine the anchored path.** `task_spec_path` passes an absolute `spec_file` through verbatim, but the root was unconditionally the worktree. `_serialized_worktree_path` keeps a path verbatim exactly when `relative_to(worktree_path)` raises, so an absolute value beside a set `worktree_path` *is* the out-of-mount shape. `_atomic_write_spec` gates on the same lexical `is_relative_to` and silently took the plain no-follow arm, losing the confined arm's `O_NOFOLLOW` walk (#593); `_restore_rearmed_spec`, which calls the confined writer directly, raised `UnconfinedWriteError` instead — turning a recoverable re-arm abort into a lost undo. It now yields the project for that shape, compared with the same lexical test the writer gates on so root and gate agree by construction (deliberately not canonicalized — that would diverge from the gate).
- **`_paused_spec_root`'s two arms made two different claims** (`self.project`, a canonicalized constructor value, vs. the delegate's `Path(state.project)`). Now one claim. This arm is unreachable from the write path today; the fix is about not leaving a second claim for a future caller, and the docstring says so rather than overclaiming.

A spec missing *or* undecodable at the anchored path now reads as an explicit fault rather than an empty body — the `UnicodeDecodeError` arm matters because all three review surfaces call this from the Textual event loop, where an escaping raise takes the dashboard down.

### Later review rounds (`69e5d5c3`..`f9de3cb1`)

Four further passes each found the *same* defect at the next surface — a persisted `spec_file` resolved against the reader's cwd — which is the shape of the problem, not a queue of unrelated bugs:

- **The fields *beside* the spec were left on the main checkout.** The re-anchor had been adopted for `spec_file` alone, so the escalation modal read its spec text from the run's tree and its sentinel indicator from the project — one modal contradicting itself, showing a pre-planning sentinel wedge as an ordinary escalation. `task_spec_root` answers "which tree can *confine* a write", and its out-of-mount arm falls back to the project; the stories folder is a different question, so `task_stories_root` was split out to mirror `stories_engine._stories_folder`'s rule.
- **The restart arm discarded the mount but kept what was measured inside it.** `worktree_path`/`branch` were cleared while `baseline_commit`, `baseline_untracked`, `spec_file` and `dispatched_spec_file` survived, and the arm saves *before* the replacement is mounted — so a git spawn fault persisted them beside an empty `worktree_path`. A later resume then took the in-place `elif task.baseline_commit:` leg against the **main checkout**: `untracked_files(repo) - baseline_untracked` named every untracked file in the operator's own checkout as attempt debris, and recovery could restore a dead attempt's snapshot over their own copy of the spec. Neither operand failed loud — linked worktrees share the object database, so a deleted unit branch's baseline still resolves and a reset onto it still succeeds.
- **Sweep never inherited any of it.** `SweepEngine` replaces `_loop` wholesale, and `Engine._loop` is the only caller of `_finish_inflight`, which carries the re-anchor — while both engines share the discard helper. Fixed in `_recover_inflight_bundle`, above the `isolated` gate for the reason the engine puts it there: the gate is live policy, the relative spelling is persisted state.
- **Pause notifications printed the raw field** on the surface the operator reads first, sprint mode included, so `_operator_spec_path` moved from `StoriesEngine` to `Engine`. The dev-session prompt deliberately keeps the raw value: that session's cwd *is* the mount.
- **The unreadable-spec refusal was asymmetric in the wrong direction.** The previous pass refused both escalation verbs; `Resolve` writes nothing, is what repairs a bad anchor, and gating it left `close` as the modal's only action — while the `R` binding reached the same agent regardless, making the refusal advisory. `Re-arm` stays refused. The unreadable notice also *prefixed* the shared body render, which answers `""` for the failure sentence, so the modal printed the warning and "(no blocking condition recorded)" together — the second denying the first. It now replaces the body.

### The guard

Nothing made the anchor rule checkable, which is exactly why each round found only the next unanchored reader. `tests/test_portability_guard.py` now flags a raw `Path(x.spec_file)` / `Path(x.dispatched_spec_file)` **by call shape** (so an alias is caught too), allowlisting `runs.py`, `engine.py`, `verify.py` and `recovery_flow.py` as the tree-local consumers. Adding a file to that list is a claim that its cwd *is* the run's tree.

Since the guard asserts an absence, two companion rows grade the detector itself: it fires on the exact line the defect shipped as, and stays silent on the sanctioned spellings.

### Scope note

This intentionally changes `rearm_escalation` at four call sites; the promotion was otherwise rename-only. The direction is one-way: where the project contains the spec, a skipped-or-refused confined write becomes a taken one; where nothing contains it, the outcome is unchanged.

## Verification

7190 passed, 49 skipped · `pyright` 0 errors · `trunk check --all` clean (258 files).

Every behavioral claim was ablated with the bytecode cache purged and `PYTHONDONTWRITEBYTECODE=1`, control-first so a collection error could not masquerade as a pass. Reverting `task_spec_root` reddens exactly three rows — including the `UnconfinedWriteError` one — while both unchanged-shape rows stay green; two deliberately over-broad variants are each caught by the "stays on the worktree" row.

Two later rows close gaps that were **fully green under ablation**, so they graded nothing before:

- The stories/spec root split is now graded by a row where the two resolvers genuinely *disagree* (an absolute, out-of-mount `spec_file`), at both the `runs` primitive and `build_context`. Every other row builds the relative shape where they agree by construction — which is why collapsing `task_stories_root` back into `task_spec_root` left the whole suite green.
- The plan-checkpoint row asserts the **notify body**, not only the journal record. No row in the repo observed any `gates.notify` body, so every notification site could have been reverted to a bare `task.spec_file` with the suite green.

`_stories_paused_run` now refuses `spec_outside_worktree` without a `worktree_path` — a combination that read like an isolated row while silently building a non-isolated one.

One assertion is **inert on POSIX**: `str()` and `.as_posix()` are byte-identical for an absolute POSIX path, so the `spec_file` posix check only grades on Windows CI. That is written into the test docstring rather than left implicit.

## Review

Three parallel review layers ran. 7 findings patched, 8 deferred to the ledger, 9 dismissed with reasons recorded. Two reviewer claims were refuted by running them: `trunk` does not reflow the CHANGELOG bullet, and the new failure string is not parsed as spec content.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected spec and story path handling for isolated runs and worktrees.
  - Improved notifications, journals, dashboards, and context details.
  - Prevented re-arm, approval, and replanning when specs or policies are unreadable.
  - Improved recovery for missing worktrees and stale mount state.
  - Preserved spec ownership and baseline information during restarts and isolation changes.
  - Re-drive decisions now reflect the active isolation policy, with clearer correction guidance.
  - Added warnings when isolation settings change during resolution.

- **Documentation**
  - Updated guides with anchored paths, unavailable actions, and correction workflows.


pbean:
@codex review

pbean:
@coderabbitai review

pbean:
@codex review

pbean:
@coderabbitai full review

pbean:
@codex review

pbean:
@coderabbitai review

pbean:
@codex review

pbean:
@coderabbitai review

pbean:
@codex review

pbean:
@codex review

pbean:
@coderabbitai review

pbean:
@coderabbitai full review

pbean:
@codex review

pbean:
@coderabbitai full review

pbean:
@codex review

pbean:
@coderabbitai full review

pbean:
@codex review

pbean:
@coderabbitai full review

pbean:
@coderabbitai full review

pbean:
@codex review

pbean:
@codex review

pbean:
@coderabbitai full review

pbean:
Both round-7 findings validated against the code. Both are real, both are **pre-existing and outside this PR's diff**, and both are now recorded in the deferred-work ledger rather than patched here.

**1. `runs.py` — resume hold derived from a journal re-read.** The mechanism is accurate: `journal_entries_or_none` returns `None` on a `UnicodeDecodeError`, `_echo_rearm_events` then returns `False`, and the gesture resumes. But `rearm-upstream-write-unreachable` does not introduce it — `_echo_rearm_events` has **zero diff lines** across this PR's three commits, and the degrade is present verbatim at `27588b41` with its rationale already stated: *"the hold degrades with the echo... an unproven hold is a guess, and this is what the gesture did before either existed."* `rearm-spec-write-unreachable` has carried the identical exposure since it began holding the resume; the new kind only inherits it.

The suggested remedy — return the hold verdict from the repair operation instead of re-deriving it — is the right shape, and it is a design change to a released seam consumed by both the CLI and the TUI. That belongs in its own change, not a seventh round on this one.

**2. `workspace.py` — per-story branch reclaim.** `workspace.py` has **zero commits** in this PR's range (`27588b41..HEAD`); the reclaim described arrived in `0cc71643`, and the finding is explicitly framed as a re-raise ("fresh evidence beyond the earlier orphan-reclaim comment"). Recorded as your claim, flagged not-independently-verified, to be confirmed before acting.

Note this review targeted `5cfca6c6` and missed `7b7e8c33`, which adds a test pinning the non-repo degrade of the new git read (a `GitError` escaping there would abort a re-arm that had already deleted the sentinel).

For the record on the two P1s from round 6, both fixed here: the sentinel gate is narrowed by **proof**, not reachability alone — gating on reachability would have fired on 100% of isolated sentinel re-arms and, since the kind holds the resume, turned every one into a two-command gesture. Ten ablations were run, each confirmed to redden only its intended row; two of them found real defects rather than confirming the work (a redundant in-place short-circuit that *shadowed* the graded one, and the missing non-repo test). CI is green on all 10 jobs including both Windows legs.

pbean:
@codex review

pbean:
@coderabbitai full review

pbean:
Round-8 finding validated: **real, and outside this PR's diff.**

`engine.py` has **zero commits** in `27588b41..HEAD`, and `spec_reaches_the_redrive` is untouched — the diff adds four new symbols *after* it, so the apparent hits are the hunk-header context line and mentions inside new docstrings. `_finish_inflight` and `_dispatched_spec_for_attempt` are both engine-side.

I confirmed the structural half: `_release_orphaned_mount` has exactly four call sites (engine.py:1676/1707/1729/1767) and **all four handle `worktree -> none`**. There is no counterpart for the opposite transition, which is your point. The dispatch-rejection consequence I have recorded as your claim rather than verified end-to-end.

Filed in the deferred-work ledger, flagged as the **fourth finding on one seam**: spec anchoring and mount-state release across an isolation flip. The other three are the engine legs fixed in `0cc71643`, your `workspace.py` per-story-branch reclaim, and CodeRabbit's `sweep.py` `_recover_inflight_bundle` twin.

That pattern is why I am not patching it here. Rounds 7 and 8 produced **only** findings in this family and **only** outside the diff, at P2, after round 6's two P1s — severity fell and stayed fallen while the findings drifted off the change. This PR's two P1s are fixed, tested and green; the seam wants one deliberate program covering both flip directions, all four call sites, the sweep twin and the branch reclaim. A standing gap from round 5 belongs in that program too: of those four call sites, only the DEV_VERIFY leg is pinned by a test.

CI is green on `e057c162` — all 10 jobs, both Windows legs.

pbean:
@codex review

Issue #496, opened 2026-08-08
Enabled plugins are trusted by NAME, so a session can swap the module behind one


pbean:
Split out of #461 point 4 / PR #495, where the config-integrity pin closes the config half of this but not the implementation half. Found by an automated reviewer on that PR and confirmed against the code.

## The hole

`trust.require_enabled` gates in-process Python on the plugin **name** only:

```python
def require_enabled(policy, name) -> None:
    if not is_enabled(policy, name):
        raise PluginUntrusted(...)
```

and `plugins/loader.py` discovery is an overlay whose **highest** precedence source is inside the agent-writable workspace:

```
builtin (bmad_loop.data/plugins/*)         lowest
entry_point                                (inert today)
project (<project>/.bmad-loop/plugins/*)   highest — same-name override
```

`registry._construct` then resolves `Path(manifest.scripts_dir) / manifest.python.module` and calls `spec.loader.exec_module(module)`. For a project-origin manifest, `scripts_dir` is `<project>/.bmad-loop/plugins/<name>/`.

So given a project whose policy already has `[plugins] enabled = ["tea"]`, a driven session can write `.bmad-loop/plugins/tea/plugin.toml` + its own module, and the next registry build imports and executes that module **in the orchestrator process**. The name never changed, so the allowlist is satisfied. The simpler variant needs no builtin to shadow: if `enabled` names a plugin that does not exist yet, the session just creates the directory.

Unlike the ungated-declarative-shell-hooks item (also out of scope on #495), this one is inside the trust tier that is *supposed* to be gated — `plugins/trust.py`'s own docstring says a `[python]` plugin is "never imported or executed unless its name is in `[plugins] enabled`", and that is true of the name while saying nothing about the bytes.

## Why PR #495 does not close it

#495's `runsetup.config_digest` hashes `sorted(policy.plugins.enabled)`. That catches a session **adding** a name to the allowlist. It cannot catch a session changing what an already-allowlisted name resolves to, because no policy field moves. The gap is recorded in the `config_digest` docstring and the CHANGELOG entry so the pin is not read as stronger than it is.

## Shape of a fix (undecided — needs a maintainer call)

Two directions, and this is a plugin-trust-model decision rather than a hashing one, which is why it is not folded into #495:

[...]

PR #803, opened 2026-09-17
chore(plugins): remove the unused trusted field from LoadedPlugin

gmoncor:
What: Remove the trusted: bool = True field from LoadedPlugin and the trusted=False kwarg passed at its one write site in registry.py, since nothing reads it anywhere in the plugins package.

Why: A repo-wide search of .trusted accesses in src/bmad_loop/plugins/ finds only the one write site and zero reads; real trust decisions already use lp.instance is not None and lp.manifest.python is None. The field is dead and could mislead a future reader into believing it gates something.

How:
- Delete the trusted field from the LoadedPlugin dataclass in model.py.
- Delete the trusted=False kwarg at its construction site in registry.py's _resolve.

Testing: Run the existing plugins pytest suite to confirm no test references .trusted; if one does, update it in this same PR. (Three assertions in tests/test_plugin_trust.py and tests/test_plugin_workflows.py did reference .trusted and were updated to assert on the real trust signals instead: lp.instance and lp.manifest.python.)

Changelog: Removed - Remove the unused trusted field from LoadedPlugin and the trusted=False kwarg passed at its one write site in registry.py, since nothing reads it anywhere in the plugins package.

Related to #496 (plugins/trust.py::require_enabled name-gate and plugins/loader.py overlay precedence) — same subsystem, different defect; not affected by this dead-code removal.


gmoncor:
Closing: this is correct dead-code cleanup (`trusted` was never read anywhere), but on reflection it's not worth a maintainer's review time on its own — no behavior depends on it, so there's no bug or confusion it actually prevents today. Happy to fold it into a future PR that touches `LoadedPlugin`/`registry.py` for a substantive reason.

Issue #810, opened 2026-09-17
Remove deprecated [engine] compatibility block in policy.py


gmoncor:
The `[engine]` block in `policy.toml` is a compatibility alias, folded into `[plugins.unity]` at load time by `_fold_deprecated_engine` (`src/bmad_loop/policy.py`). It emits a `DeprecationWarning` at runtime saying it will be removed, but until now named no concrete target version.

This issue is the tracking/ownership point for actually deleting `_fold_deprecated_engine` and the `[engine]` compatibility path, targeted for version 0.12.0.

## Acceptance criteria
- `_fold_deprecated_engine` and the `[engine]`-block handling are removed from `src/bmad_loop/policy.py`.
- Any remaining docs/tests referencing the legacy `[engine]` block are updated to the `[plugins.unity]` form.
- Released as part of, or after, version 0.12.0.

Issue #157, opened 2026-07-16
session-end journaled 2h19 after session_timeout_min fires when the session is wedged inside a tool call


ferrandmathis-afk:
**Version:** bmad-loop 0.8.1 (uv tool, git install) · macOS 15 (Darwin 24.6.0) · git worktree isolation, `adapter.name = "claude"`, `limits.session_timeout_min = 90`
**Component:** claude adapter session teardown (`engine.py:2061` passes `timeout_s = session_timeout_min * 60` into `SessionSpec`)

## Summary

`session_timeout_min = 90` bounded the *model* on schedule but not the *session*: `session-end` was journaled **3h49 after** `session-start`, i.e. 2h19 after the deadline. The orchestrator sat on a wedged session for over two hours with no event emitted.

Timeline for a single dev story (all times local, one run, worktree isolation):

| time | source | event |
|---|---|---|
| 23:25:01 | journal | `session-start` |
| 00:55:01 | — | `session_timeout_min = 90` elapses |
| 00:55:56 | transcript | last assistant message — last sign of life |
| ~01:39 | log mtime | terminal log stops growing |
| **03:14:25** | journal | `session-end` `status=timeout` ← **2h19 late** |
| 03:15:36 | journal | `dev-decision` retry |
| 03:22:51 | journal | `run-crash` (filed separately) |

The last assistant message lands within a minute of the 90-minute deadline, so the timeout itself looks like it fires on schedule. What takes 2h19 is the stretch between firing and recording `session-end`.

## Why the session was wedged

It was stuck **inside a tool call**, not between turns. The captured pane ends on:

```
Running in the background (↓ to manage) · Waddling… (59m 11s · ↓70.4k tokens)
```

The story was standing up throwaway Postgres containers (`docker run -d … pgvector/pgvector:pg17`) to verify a DB migration, and one of those calls never returned.

This matters for the stall machinery: because the session never ends a turn, it never emits a result-less `Stop`, so `dev_stall_grace_s` / `dev_stall_nudges` (#151's territory) never engage. `session_timeout_min` is the *only* bound on this shape of hang — and it didn't translate into a timely teardown.

## What I expected

Once `timeout_s` elapses, the adapter kills the session and `session-end` is journaled promptly — seconds, not hours — whether or not the pane is blocked inside a tool call.

## Proposed fix

[...]

Issue #149, opened 2026-07-15
Claude adapter: Stop hook + idle detection re-entrant loop prevents session from ending


tronghieu:
## Description

When a Claude Code session running under bmad-loop finishes its work and enters the "Leavening" phase (stop hooks execution), a re-entrant loop can occur that prevents the session from ever properly ending.

## Steps to reproduce

1. Configure a Claude Code project with Stop hooks that take more than ~5 seconds to complete (e.g., a hook that runs `gh` CLI commands or a test suite)
2. Run `bmad-loop run` with a story
3. Let the dev/review session complete its work and attempt to stop

## What happens

1. Session finishes work → enters "Leavening" phase → stop hooks begin executing
2. Claude Code shows `running stop hooks… 0/N` (the counter never progresses past 0)
3. After ~5 seconds of hooks not completing, bmad-loop idle detection fires and writes "You appear idle in bmad-loop automation mode" into the TTY
4. Claude sees this message as user input → tries to respond/explain → this triggers the Stop hooks AGAIN
5. The cycle repeats: stop hooks run → idle detection → Claude responds → stop hooks run again → ...

## Observed behavior

- The stop hook counter stays at `0/N` indefinitely
- `bmad_loop_hook.py Stop` completes successfully each iteration (Stop events are written to the run directory)
- But the session never transitions to `session-end` in the journal
- The Claude process remains alive with an open API connection, stuck in this loop
- Tokens continue to be consumed as Claude re-processes the idle message each cycle

## Expected behavior

Stop hooks should either:
- Complete (or timeout) before idle detection can fire, OR
- Idle detection should not inject messages into the TTY while stop hooks are in progress, OR
- The session should eventually end after a finite number of idle detection cycles

## Environment

- bmad-loop 0.8.0
- Claude Code adapter
- macOS

## Impact

Sessions that are functionally complete (all work done, verification passing, commits made) cannot terminate, consuming tokens and requiring manual intervention to kill the process.

pbean:
Thanks for the detailed report — validated, and the fix is up in #151. A few findings from the investigation that refine the mechanism:

[...]

PR #811, opened 2026-09-17
docs(policy): pin retirement version for deprecated [engine] block

gmoncor:
What: Pin a concrete target version (and tracking issue) for retiring the `[engine]` compatibility alias, in both the code comment and the runtime `DeprecationWarning` message.

Why: The alias says it is a "one-release" compatibility shim but names no version or issue, so nothing owns its removal — it can (and likely will) outlive its intended lifetime, matching a known anti-pattern already called out in this project's own review conventions.

How:
- Replace "a future release" / "one-release compatibility alias" with a concrete version number (0.12.0) in the code comment and the `warnings.warn` message.
- Reference tracking issue #810 for the removal, following the `#149`/`#156`/`#157` convention already used elsewhere in this same file.

Testing: `uv run pytest -q tests/test_policy.py` — 352 passed, 2 skipped. Confirms the deprecation-warning path still matches the updated message text.

Changelog: `Changed: pinned a target version for the deprecated [engine] settings block.`

Related to #810


gmoncor:
Note for reviewers: this PR's DeprecationWarning (_fold_deprecated_engine) shares the same visibility gap tracked in #835 (discovered while addressing review feedback on #814) — Python suppresses DeprecationWarning by default outside __main__, so this notice likely doesn't reach a user in normal CLI use either. Not blocking this PR on it since it's cross-cutting and pre-existing, but flagging for awareness.

PR #814, opened 2026-09-17
fix(policy): warn on legacy 'retry' on_status_contradiction mode

gmoncor:
**What:** Pin a retirement target for the legacy `retry` value of `review.on_status_contradiction` and emit a load-time warning when it is selected.

**Why:** The option is already labeled "legacy" in its own description since #334 introduced `escalate` as the default, but has no version/issue owning its removal, and nothing warns an operator who is still using it. Related to #334 (closed), which established `escalate` as the default and demoted `retry`.

**How:**
- Add a target version/issue reference to the core.toml description and the ReviewPolicy docstring.
- Warn via `warnings.warn(...)` when `on_status_contradiction == "retry"` loads, mirroring the `_fold_deprecated_engine` pattern; the `escalate` default stays silent.
- Add a `pytest.warns` test for the `retry` case and a negative test confirming `escalate` triggers no warning.

**Testing:** `uv run pytest -q tests/test_policy.py`.

**Changelog:** `Changed: bmad-loop now warns when the legacy review.on_status_contradiction = "retry" mode is configured.`


gmoncor:
CodeRabbit's remaining finding (DeprecationWarning not surfacing in normal CLI use) is a real, cross-cutting issue — not specific to this PR's actual content, which is otherwise correct and covered by tests. Tracked separately in #835 rather than fixed here, since a good fix means a CLI-wide decision about warning categories/filters shared with #811. Leaving this in draft until that's resolved (or a decision is made that this PR doesn't need to wait on it).

